PR #7, opened 2023-02-10
Sex is required

captDaylight:
Sorry, not sorry, for crazy PR title. But this is my task ref https://www.notion.so/photons/Sex-is-not-required-on-frontend-when-it-is-required-for-the-API-15f46c740b254b93badcace18a18c33c?pvs=4

Found other fields that were not being validated on patient create form including first name, last name, phone, and email.

### DIFF
--- a/packages/elements/package.json
+++ b/packages/elements/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@photonhealth/elements",
-  "version": "0.0.81",
+  "version": "0.0.82",
   "description": "",
   "main": "dist/index.js",
   "types": "./dist/index.d.ts",
--- a/packages/elements/src/photon-patient-dialog/index.tsx
+++ b/packages/elements/src/photon-patient-dialog/index.tsx
@@ -59,7 +59,7 @@ customElement(
       createPrescription = false
     ) => {
       setLoading(true);
-      let keys: string[] = ['firstName', 'lastName', 'phone', 'sex', 'email'];
+      let keys: string[] = ['firstName', 'lastName', 'sex', 'email'];
 
       if (
         store['address_street1']?.value !== undefined ||
--- a/packages/elements/src/photon-patient-dialog/index.tsx
+++ b/packages/elements/src/photon-patient-dialog/index.tsx
@@ -59,7 +59,7 @@ customElement(
       createPrescription = false
     ) => {
       setLoading(true);
-      let keys: string[] = ['firstName', 'lastName', 'sex', 'email'];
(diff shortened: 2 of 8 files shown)


captDaylight:
Are we enforcing addresses on new patients? right now it’s using the logic from update patient which only makes you fill in more address information if one of the address fields is touched… but I imagine we’ll want to capture that information on patient creation

packages/elements/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@photonhealth/elements",
-  "version": "0.0.81",
+  "version": "0.0.82",
```
samkotlove, line 3: So is this already published?

packages/elements/src/photon-patient-dialog/index.tsx
```diff
@@ -47,7 +47,7 @@ customElement(
       const event = new CustomEvent('photon-patient-closed', {
         composed: true,
         bubbles: true,
-        detail: {},
+        detail: {}
```
samkotlove, line 50: do we have a linting file in here?

packages/elements/src/photon-patient-form/index.tsx
```diff
@@ -58,30 +59,34 @@ customElement(
...
-      validator: message(size(string(), 1, Infinity), 'Please enter a last name...'),
+      validator: message(size(string(), 1, Infinity), 'Please enter a last name...')
+    });
+    actions.registerValidator({
+      key: 'sex',
+      validator: message(enums(sexes.map((s) => s.name)), 'Please enter Sex at Birth...')
     });
     actions.registerValidator({
       key: 'phone',
-      validator: message(size(string(), 10, 12), 'Please enter a phone number...'),
```
samkotlove, line 73: Isn't 10 a minimum?

packages/elements/src/photon-patient-dialog/index.tsx
captDaylight: we do, it's global to the repo now, was gonna run  it on a separate PR so it doesn't blow up the commits

packages/elements/src/photon-patient-form/index.tsx
captDaylight: it is, but the input converts 2223334444 (10) into +12223334444 (12)

packages/elements/package.json
captDaylight: nope, I'll publish it when it lands on boson

samkotlove (approved):

PR #8, opened 2023-02-10
Update Neutron

captDaylight:
- field "sex at birth", "first name", and 'last name" are required


.github/workflows/add-comment.yml
```diff
@@ -0,0 +1,20 @@
+on:
+  pull_request:
+    types: [opened, reopened]
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v5
+        with:
```
samkotlove, line 10: we should add a filter so this only runs on pull requests to boson

samkotlove (approved):

PR #15, opened 2023-02-16
nix query string

captDaylight:
when I create a new patient it get kicked back to something like this:
```
https://app.boson.health/patients?reload=pat_01GSE0VR7SG3QKNNV82A75AJZ3-01GSE0VRBKWQ4ACDG2RFWJ4P6G 
```
which when I refresh breaks the list populating 1 out of 3 ish times

but this doesn't break
```
https://app.boson.health/patients
```
Task Ref https://www.notion.so/photons/Adding-new-patient-loads-an-empty-patient-list-on-success-80019255ab964755bdeadf6faeeb80d1?pvs=4

### DIFF
--- a/apps/app/src/views/routes/PatientForm.tsx
+++ b/apps/app/src/views/routes/PatientForm.tsx
@@ -22,7 +22,7 @@ export const PatientForm = () => {
         if (e?.detail?.createPrescription) {
           navigate(`/prescriptions/new?patientId=${id}`);
         } else {
-          navigate(`/patients?reload=${id}-${ulid()}`);
+          navigate(`/patients`);
         }
       });
       ref.current.addEventListener('photon-patient-closed', () => {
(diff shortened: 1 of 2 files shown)


apps/app/src/views/routes/PatientForm.tsx
```diff
@@ -22,7 +22,7 @@ export const PatientForm = () => {
         if (e?.detail?.createPrescription) {
           navigate(`/prescriptions/new?patientId=${id}`);
         } else {
-          navigate(`/patients?reload=${id}-${ulid()}`);
+          navigate(`/patients`);
```
samkotlove, line 25: Do we know why we had those query params? 
@jknapp25 @cawfeecoder

cawfeecoder: I used to remember - I think it had to do with how routing was setup to make sure that we'd reload the list that you get kicked back to on save/updated (since one is a webcomponent and the other is a React element)

captDaylight: From what I can tell it's not filtering the list in any way, it's the entire list either way

captDaylight: When I update a patient I'm seeing the changes reflected (though I can't update a name... should I be able to?)

https://www.loom.com/share/b400b1d631cc4aea882311b27b2f9516

jknapp25 (approved):

PR #25, opened 2023-02-20
Check for Org id match to authorize

captDaylight:
- creates an `isAuthorized` value in store that checks if sdk org matches user org
- adds a `PhotonAuthorized` wrapper that displays children based on `isAuthorized` value

cc @samkotlove @cawfeecoder

### DIFF
--- a/packages/elements/src/photon-authorized/index.tsx
+++ b/packages/elements/src/photon-authorized/index.tsx
@@ -41,7 +41,12 @@ export const PhotonAuthorized = (props: { children: JSXElement }) => {
       <Show when={client && authenticated() && !authorized() && !isLoading()}>
         <sl-alert variant="warning" open>
           <sl-icon slot="icon" name="exclamation-triangle"></sl-icon>
-          <strong>Something went wrong, please contact support at Photon Health</strong>
+          <strong>
+            Something went wrong, please contact support at Photon Health{' '}
+            <a style="text-decoration:underline" href="mailto:support@photon.health">
+              support@photon.health
+            </a>
+          </strong>
           <br />
         </sl-alert>
       </Show>
--- a/packages/elements/src/photon-authorized/index.tsx
+++ b/packages/elements/src/photon-authorized/index.tsx
@@ -41,7 +41,7 @@ export const PhotonAuthorized = (props: { children: JSXElement }) => {
       <Show when={client && authenticated() && !authorized() && !isLoading()}>
         <sl-alert variant="warning" open>
           <sl-icon slot="icon" name="exclamation-triangle"></sl-icon>
-          <strong>You are not authorized to prescribe</strong>
+          <strong>Something went wrong, please contact support at Photon Health</strong>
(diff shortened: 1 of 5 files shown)


packages/elements/src/photon-authorized/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  createEffect(() => {
+    setIsLoading(client?.authentication.state.isLoading || false);
+  });
+
+  return (
+    <>
+      <Show when={client && !authorized() && !isLoading()}>
+        <sl-alert variant="warning" open>
+          <sl-icon slot="icon" name="exclamation-triangle"></sl-icon>
+          <strong>You are not authorized to prescribe</strong>
```
captDaylight, line 30: @samkotlove  update this to say something like please contact support at Photon?

packages/elements/src/photon-multirx-form/index.tsx
```diff
@@ -331,14 +328,7 @@ customElement(
                 <sl-spinner style="font-size: 3rem;"></sl-spinner>
               </div>
             </Show>
-            <Show when={client && !authenticated() && hasError()}>
```
samkotlove, line 334: This message was incorrect before (not being authenticated != not being authorized to prescribe), but i think we should show some error in this case

packages/elements/src/photon-authorized/index.tsx
samkotlove: Yeah I would keep it broad. "something went wrong" or w/e

packages/elements/src/photon-authorized/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+  createEffect(() => {
+    setAuthenticated(client?.authentication.state.isAuthenticated || false);
+  });
+  createEffect(() => {
+    setAuthorized(client?.authentication.state.isAuthorized || false);
+  });
+
+  return (
+    <>
+      <Show when={client && !authenticated()}>
```
captDaylight, line 34: @samkotlove this should handle the not authenticated message you mentioned below

Screen Shot 2023-02-20 at 12 33 38 PM

packages/elements/src/photon-authorized/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <sl-alert variant="warning" open>
+          <sl-icon slot="icon" name="exclamation-triangle"></sl-icon>
+          <strong>You are not signed in</strong>
+          <br />
+        </sl-alert>
+      </Show>
+      <Show when={client && authenticated() && !authorized() && !isLoading()}>
+        <sl-alert variant="warning" open>
+          <sl-icon slot="icon" name="exclamation-triangle"></sl-icon>
+          <strong>Something went wrong, please contact support at Photon Health</strong>
```
samkotlove, line 44: could we include support@photon.health

samkotlove (approved):

PR #42, opened 2023-02-24
Show pharmacy selection only for Modern Pediatrics

jknapp25:
We'll want to wait til we have the go-ahead from MP to get this on neutron and photon


apps/app/src/views/routes/NewOrder/components/SelectPharmacyCard/index.tsx
```diff
@@ -12,14 +12,16 @@ import {
...
-import { types } from '@photonhealth/react';
+import { types, usePhoton } from '@photonhealth/react';
 
 import { LocationSearch } from '../../../../components/LocationSearch';
 import { LocalPickup } from './components/LocalPickup';
 import { MailOrder } from './components/MailOrder';
 import { Address } from '../../../../../models/general';
 import { SendToPatient } from './components/SendToPatient';
 
+const pharmacySelectionEnabledOrgs = [process.env.REACT_APP_MODERN_PEDIATRICS_ORG_ID];
```
samkotlove, line 23: Like this. Think we can do something similar with the `onlyCurexa` field? Not sure if that'll make it more confusing or not.

Maybe the inverse?

samkotlove (approved):
Looks good! Left one small ask but doesnt necessarily have to make it in

PR #43, opened 2023-02-27
Centralize org-specific fulfillment settings in config

jknapp25:
This is a pre-cursor to storing fulfillment settings on the org and the user.

Tested order creation as:

- default user
- peachy user
- weekend user
- modern pediatrics user
- modern pediatrics as Tim


apps/app/src/views/routes/NewOrder/components/SelectPharmacyCard/index.tsx
```diff
@@ -100,13 +104,20 @@ export const SelectPharmacyCard: React.FC<SelectPharmacyCardProps> = ({
     {
       name: 'Send to Patient',
       fulfillmentType: undefined,
-      isDisabled: onlyCurexa,
+      enabled:
+        typeof fulfillmentConfig[user.org_id] !== 'undefined'
+          ? fulfillmentConfig[user.org_id].sendToPatient &&
+            fulfillmentConfig[user.org_id].sendToPatientUsers?.includes(user.id)
```
jknapp25, line 110: @cawfeecoder it looks like auth0 isn't returning user id is that correct?

jknapp25: updated to retrieve auth0 id from token

apps/app/.env.photon
```diff
@@ -11,5 +11,9 @@ REACT_APP_DATADOG_RUM_CLIENT_TOKEN=pub175e353f3868efed64bbe14328f22536
 
 REACT_APP_CUREXA_ORG_ID=phr_01GCA54GVKA06C905DETQ9SY98
 
+REACT_APP_MODERN_PEDIATRICS_ORG_ID=org_2utnNgp5NGCy9wOb
+REACT_APP_TIM_USER_ID=google-oauth2|105026997775584560678
+REACT_APP_SARA_USER_ID=google-oauth2|109153884033500835980
```
jknapp25, line 16: sara's only present on photon

apps/app/src/configs/fulfillment.ts
```diff
@@ -0,0 +1,42 @@
...
+    mailOrderProviders: [
+      ...(process.env.REACT_APP_CUREXA_ORG_ID ? [process.env.REACT_APP_CUREXA_ORG_ID] : [])
+    ],
+    sendToPatient: false,
+    sendToPatientUsers: [] // its weird to have this here, but its temporary and keeps ts from complaining
+  },
+  // Weekend Health
+  [process.env.REACT_APP_WEEKEND_ORG_ID as string]: {
+    sendOrder: false,
+    pickUp: true,
```
samkotlove, line 15: One thing we lost was `disableOrderCreation`, could we disable pickUp here as well for weekend health? As they only want to create orders programatically, so maybe the component can just show an error if there are no valid options

samkotlove: Wait nvm, just noticed `sendOrder`, ignore me

samkotlove (commented):
1 small comment
edit: whoops nvm!

samkotlove (approved):

PR #48, opened 2023-03-01
Neutron: permission checks + centralize fulfillment settings

captDaylight:


jknapp25:
Thanks for creating this @captDaylight! 👏
We'll want to bump all lambdas on neutron (accept for add_preferred_pharmacy) before merging the fulfillment config. I was waiting on the env-specific config to do that, but we don't technically need that yet. So if @samkotlove gives it the 🟢, so do I.

jknapp25:
now that I think about it, there's a bit of a queue for fulfillment:

1. entela finishes send_slack updates 
2. neutron tf get's updated
3. merge this

jknapp25 (approved):

PR #49, opened 2023-03-01
Add env-specific fulfillment configs

jknapp25:


apps/app/src/views/routes/NewOrder/components/OrderForm.tsx
```diff
@@ -131,9 +131,9 @@ export const OrderForm = ({
   };
 
   const orderCreationEnabled =
-    typeof fulfillmentConfig[user.org_id]?.sendOrder !== 'undefined'
-      ? fulfillmentConfig[user.org_id]?.sendOrder
-      : fulfillmentConfig.default.sendOrder;
+    typeof fulfillmentConfig[process.env.REACT_APP_ENV_NAME][user.org_id]?.sendOrder !== 'undefined'
```
jknapp25, line 134: I'm not a TS expert, naturally these are complaining about potentially being undefined, but checking beforehand doesn't work either

image

cawfeecoder: if you "know" better than TypeScript that it's always defined, postfix the variable with a ! (e.g. process.env.REACT_APP_ENV_NAME!). If you think it may be undefined at any point - you should wrap it in an if statement so that TypeScript does what is called "type narrowing"

cawfeecoder: is REACT_APP_ENV_NAME ever undefined.

Also, the correct if statement would be:
```
if ('REACT_APP_ENV_NAME' in process.env) {
  ...
}
```

jknapp25: it's not, but in my experience react will always consider env variables to potentially be undefined

cawfeecoder: You could add a "!" if to end the of each use (like the above e.g.) if you know they're always defined to avoid the if statement.

apps/app/src/configs/fulfillment.ts
```diff
@@ -1,40 +1,115 @@
...
+      mailOrder: boolean;
+      mailOrderProviders: string[];
+      sendToPatient: boolean;
+      sendToPatientUsers: string[];
+    };
+  };
+};
+
+export const fulfillmentConfig: FulfillmentConfig = {
+  boson: {
```
samkotlove, line 20: Could we maybe do different files? `fulfillment.boson.ts` or s/e

jknapp25: sure!

apps/app/src/configs/fulfillment.boson.ts
```diff
@@ -0,0 +1,23 @@
+export type FulfillmentSettings = {
```
samkotlove, line 1: Could we put this elsewhere? a little weird that fulfillment.neutron.ts has to import from fulfillment.boson.ts

samkotlove (approved):

apps/app/src/configs/fulfillment.boson.ts
jknapp25: yep

samkotlove (approved):

PR #53, opened 2023-03-02
Show send to patient

jknapp25:
We had previously hidden send to patient on neutron and photon due to an accidental merge. This didn't get reverted in our recent deploy. 

Note: It's also hidden on photon. So we'll want to catch this in the large deploy PR.


apps/app/src/views/routes/NewOrder/components/SelectPharmacyCard/index.tsx
```diff
@@ -135,12 +133,6 @@ export const SelectPharmacyCard: React.FC<SelectPharmacyCardProps> = ({
         />
       )
     }
-    // {
```
captDaylight, line 138: this is still supposed to be commented out?

jknapp25: nope, removed since send to patient is second in the list

captDaylight (approved):

PR #61, opened 2023-03-03
Allow scripts to be edited before Save

captDaylight:
- adds edit button to draft prescriptions
- scrolls to prescriptions form if its clicked
- shows a dialog if there's unsaved info in the form

### 🍿 Loom
https://www.loom.com/share/b8597d6d3efb42a4ab0d6705fa5cf878

---
~~we probably need to work on this copy~~

UPDATED copy
image

task ref https://www.notion.so/photons/Allow-scripts-to-be-edited-before-Save-c1f13bebfe2846bcb0bef7b79a07fea8?pvs=4


jknapp25 (approved):
Awesome! 

Non-blocking thoughts:

- "Overwrite" is okay, but a little confusing to me. How about something like: "Remove in-progress prescription", "This will remove your in-progress prescription and replace it with the prescription to edit."
- Let's consider adding a [ghost button](https://chakra-ui.com/docs/components/button#button-variants) soon, we could re-use it for cancel and delete as well

captDaylight:
@jknapp25 would love your thoughts on the copy in this thread https://photonhealth.slack.com/archives/C033T9N949M/p1678140760862149

love ghost button idea

PR #62, opened 2023-03-06
Add Sentry

captDaylight:
In addition to Sentry's standard context, errors include additional context:

- Elements' Release Version
- photon-client context
  - audience
  - clientId
  - developmentMode
  - domain
  - organization ID
  - redirectURI
  - uri
- user info
  - Permissions
  - the user object
  - is in org, is authenticated
 
Then, according to what I'm seeing an additional error from Calvin at Weekend (different than the other one[ solved] it looks like something popped up 8 days ago. 

image

This is about the time that I added the Authorization check to `checkSessions` in elements store. I've combed through the commits from about 2 weeks ago until now that touched the elements folder and haven't found any potential problems outside of the `checkSessions` method. So, I wrapped it in a try catch that handles any global issue with `checkSession` with `Sentry.captureException(e)`

Other:
- fixed unhandled  error `TOO_SHORT`  and `NOT_A_NUMBER` for parsing phone numbers on the patient update and new patient form
- Instead of both buttons having spinners for update patient and new prescription, it's only the button you click on
Screen Shot 2023-03-06 at 3 56 44 PM


package.json
```diff
@@ -68,8 +68,9 @@
     "@emotion/styled": "^11.10.5",
     "@nanostores/persistent": "^0.7.0",
     "@nanostores/react": "^0.4.1",
-    "@sentry/react": "^7.28.1",
-    "@sentry/tracing": "^7.28.1",
+    "@sentry/browser": "^7.41.0",
+    "@sentry/react": "^7.41.0",
+    "@sentry/tracing": "^7.41.0",
```
photon-eng, line 73: What is our build size looking like for client?

packages/elements/sentry.ts
```diff
@@ -0,0 +1,13 @@
+import * as Sentry from '@sentry/browser';
+import { BrowserTracing } from '@sentry/tracing';
+import pjson from './package.json';
+
+Sentry.init({
+  dsn: 'https://1d34034cd858470ba77ff10172425af2@o1356305.ingest.sentry.io/4504792337219584',
+  environment: process.env.NODE_ENV,
```
photon-eng, line 7: This shouldnt be an env var right? our clients might modify this (or not set it at all?)

captDaylight: Was thinking this too, but there's ultimately no way around exposing the DSN when we publish to NPM even if we store this as env variable in the repo. That being said, as it's written clients shouldn't be able to modify it at runtime, the only way I can think would be they manually change the file in their node_modules.

TBH, the more I think about it and search the internet... I'm not sure how standard it is to include sentry in a public, published npm package. There could be some privacy and security issues.

captDaylight:
Update from Sentry regarding Sentry in public package:
- it is kosher ✅
- we will `want to do some configuration with the client so that you are not capturing errors outside of "your code".` ... looking into what this means more concretely
- The DSN being exposed is unavoidable, worst case someone sends extra events to our sentry org
- privacy issues should mostly be mitigated if "we set up client correctly", we can scrub user ip on by calling `setUser` or scrubbing it server side

samkotlove (approved):

PR #64, opened 2023-03-07
Make sure "at" exists

captDaylight:
If `r.at(-1)` isn't a function, we should check it:

```
a?.at(-1)?.isIntersecting && props.hasMore
```

image

task ref https://www.notion.so/photons/Request-Error-r-at-1-for-Weekend-6ccd91e3c55d481c844f0b0182aaa743?pvs=4


jknapp25 (approved):
Any additional places this occurs? I thought there was a few in this repo.

captDaylight:
@jknapp25 true, there were only two other spots where `.at(-1)` didn't have optional chaining, it technically shouldn't matter since they are scoped to a condition that already does the check, but better safe than sorry!

PR #65, opened 2023-03-07
fix button spinner, phone input errors, check session errors

captDaylight:
I'm taking out unrelated updates from the sentry PR https://github.com/Photon-Health/client/pull/62

- fix button spinners in the forms, should only spin on the button you clicked on
image
- phone input was throwing unhandled errors when the numbers were too short
- wrap checksession in try catch

### DIFF
--- a/packages/elements/src/store.ts
+++ b/packages/elements/src/store.ts
@@ -288,7 +288,10 @@ export class PhotonClientStore {
         permissions: permissions || []
       });
     } catch (err) {
-      console.log('Error checking photon session: ', err);
+      this.setStore('authentication', {
+        ...this.store.authentication,
+        isLoading: false
+      });
     }
   }
   private async login(args = {}) {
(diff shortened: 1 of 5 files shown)


packages/elements/src/store.ts
```diff
@@ -261,31 +261,35 @@ export class PhotonClientStore {
...
+      this.setStore('authentication', {
+        ...this.store.authentication,
+        user: user,
+        isLoading: false,
+        isInOrg: authenticated && hasOrgs && this.sdk.organization === user.org_id,
+        permissions: permissions || []
+      });
     } catch (err) {
-      permissions = [];
+      console.log('Error checking photon session: ', err);
```
jknapp25, line 291: `console.error`?

jknapp25 (approved):

packages/elements/src/store.ts
captDaylight: ya good call, this is where I had a sentry error capture... replaced this with:
```
      this.setStore('authentication', {
        ...this.store.authentication,
        isLoading: false
      });
```

PR #69, opened 2023-03-08
add rado to moped sendToPatientUsers

jknapp25:
For the sake of creating a branded demo for MoPed.


apps/app/src/configs/fulfillment.neutron.ts
```diff
@@ -49,7 +49,8 @@ export const fulfillmentSettings: FulfillmentSettings = {
     sendToPatient: false, // disabled for org
     sendToPatientUsers: [
       // enabled for these users
-      ...(process.env.REACT_APP_TIM_USER_ID ? [process.env.REACT_APP_TIM_USER_ID] : [])
+      ...(process.env.REACT_APP_TIM_USER_ID ? [process.env.REACT_APP_TIM_USER_ID] : []),
+      ...(process.env.REACT_APP_RADO_USER_ID ? [process.env.REACT_APP_RADO_USER_ID] : [])
```
samkotlove, line 53: nit: want to just put these in the array? the env var feels like overkill

samkotlove (approved):

apps/app/src/configs/fulfillment.neutron.ts
jknapp25: ah good call, kind of remnant stuff

PR #73, opened 2023-03-09
Photon Client in Components 

captDaylight:
Sorry, this is massive.

Broad strokes:
- Client move from elements into components
- stores move from elements into components (PhotonClientStore, PhotonPharmacy, PhotonPatient, etc)
- Elements now imports the above and wraps them 

There were some tricky (non) reactivity issues with Solid trying to `createStore` from the components app. For example, I could trigger the SDK to fetch a patient but `createEffect` wouldn't be able to detect a change upon successful response. So the caveat here is that for now I have to drill a `createStore` in `/elements` into stores coming from `/components`. For example:

``` js
// components/src/store.ts

export class PhotonClientStore {
  ...
  public constructor(sdk: PhotonClient, cs?: typeof createStore) {
    this.sdk = sdk;

    // TODO when we are no longer maintaining components inside of elements, we can remove this
    // this fixes an issue where the reactivity is lost when using the store in elements
    const _createStore = cs || createStore;
    const [store, setStore] = _createStore<{
      authentication: {
        isAuthenticated: boolean;
        isInOrg: boolean;
        isLoading: boolean;
        permissions: Permission[];
        error?: string;
        user: any;
      };
      ...
```

Now that I have a grasp on where I think the issue is happening I can create a reproducible example that I can float to the Solid community. But for the sake of getting a move on the pharmacy search component, this IMO is a working example that we can build on.
 
Tested
- [x] Pharmacy Search, enter zip code and drop down should have relevant address
- [x] Prescribe workflow, can create prescriptions, appear in app.boson.health
- [x] Patient dialog, can update patient
- [x] Patient dialog, can create a new patient
- [x] App, run through the above again


packages/components/src/stores/form.ts
```diff
@@ -0,0 +1,102 @@
...
+      };
+    }
+    initalValue = tmp;
+  }
+
+  // TODO when we are no longer maintaining components inside of elements, we can remove this
+  // this fixes an issue where the reactivity is lost when using the store in elements
+  const _createStore = cs || createStore;
+  const [store, setStore] = _createStore<Record<string, StoreValue | undefined>>(
+    initalValue ? initalValue : {}
```
jknapp25, line 22: typo "inital", can use double pipe here too

captDaylight: eagle eye!

packages/components/src/App.tsx
```diff
@@ -1,34 +1,17 @@
...
-    </article>
+    <Client
+      id="7N9QZujlNJHL8EIPqXpu1wq8OuXqoxKb"
+      org="org_KzSVZBQixLRkqj5d"
+      domain="auth.boson.health"
+      audience="https://api.boson.health"
+      uri="https://api.boson.health/graphql"
+    >
+      <PharmacySearch />
+    </Client>
```
samkotlove, line 14: What's going on in this file? the before is also confusing to be fair

packages/components/src/systems/PharmacySearch/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+export default function PharmacySearch(props: { uP?: any }) {
+  const client = props.uP ? props.uP() : usePhoton();
+  const [trigger, setTrigger] = createSignal(false);
+  const { store, actions } = PatientStore;
+
+  createEffect(async () => {
+    if (trigger()) {
+      setTrigger(false);
+      const data = await client?.clinical.patient.getPatient({
+        id: 'pat_01GQGDNW09VRK2CQ0Q41H57RVK'
```
samkotlove, line 14: ?

packages/components/src/systems/PharmacySearch/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+      });
+
+      console.log('fetch patient from sdk', data);
+
+      await actions.getSelectedPatient(client!.getSDK(), 'pat_01GQGDNW09VRK2CQ0Q41H57RVK');
```
samkotlove, line 19: ?

packages/components/src/systems/PharmacySearch/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const { store, actions } = PatientStore;
+
+  createEffect(async () => {
+    if (trigger()) {
+      setTrigger(false);
+      const data = await client?.clinical.patient.getPatient({
+        id: 'pat_01GQGDNW09VRK2CQ0Q41H57RVK'
+      });
+
+      console.log('fetch patient from sdk', data);
```
samkotlove, line 17: Need this?

packages/elements/src/photon-audit/index.tsx
```diff
@@ -1,45 +1,11 @@
...
-        </article>
-      </div>
-    );
-  }
-);
+customElement('photon-audit', {}, () => {
+  return (
+    <div>
+      <PharmacySearch uP={usePhoton} />
+    </div>
```
samkotlove, line 9: this is just a test file?

packages/components/src/App.tsx
captDaylight: This file is kinda like the equivalent of `/elements/index.html`, it's a non exported local app that I'm using to display some of the components.

packages/elements/src/photon-audit/index.tsx
captDaylight: Yup! just testing reactivity, this will be all be cleaned up in the next PR.

samkotlove (approved):

PR #79, opened 2023-03-14
Prescriptions Status Filter

captDaylight:
On the prescriptions page there is now a filter that allows you to select between none, status active, status depleted, status expired. Also it adds a query string for the filter, works or refresh. Looks good on mobile (I think).

📽
https://www.loom.com/share/bd468ad6651d448894a7a41774280a33

task Ref https://www.notion.so/photons/Status-of-the-prescription-Filter-868fbe55da3e438ba4926399be1aab7d?pvs=4

Shout out to @photonsFardad for quick turn around providing the API to filter by prescription state

### DIFF
--- a/apps/app/src/views/routes/Prescriptions.tsx
+++ b/apps/app/src/views/routes/Prescriptions.tsx
@@ -2,6 +2,7 @@ import { Link as RouterLink, useSearchParams } from 'react-router-dom';
 
 import {
   Badge,
+  Select,
   HStack,
   IconButton,
   Skeleton,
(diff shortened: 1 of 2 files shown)


apps/app/src/views/routes/Prescriptions.tsx
```diff
@@ -193,26 +213,46 @@ export const Prescriptions = () => {
...
+
+  useEffect(() => {
+    navigate({
+      search: status ? `?status=${status}` : ''
+    });
+    async function refetchData() {
+      setRows([]);
+      setFilterChangeLoading(true);
+      const { data } = await refetch(getPrescriptionsData);
+      setFilterChangeLoading(false);
```
samkotlove, line 241: Shouldnt `loading` on line 223 get set to true once we refetch?

captDaylight: I thought so at first but it's not... but probably for the best. If loading were set to `true` on refetch then the whole list would switch to skeleton rows rather than the spinner for infinite scroll as it's written.

samkotlove (approved):

PR #81, opened 2023-03-14
Edit Patient DOB working 

captDaylight:
- fixes Edit patient DOB, was showing today's date
- BONUS, fixes typing issue with phone input... first digit wasn't getting registered 🤷‍♀️

### DIFF
--- a/packages/elements/src/photon-datepicker/index.tsx
+++ b/packages/elements/src/photon-datepicker/index.tsx
@@ -23,7 +23,8 @@ customElement(
     required: false,
     invalid: false,
     helpText: undefined,
-    disabled: false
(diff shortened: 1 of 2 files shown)


packages/elements/src/photon-datepicker/index.tsx
```diff
@@ -32,11 +33,14 @@ customElement(
     invalid: boolean;
     helpText?: string;
     disabled: boolean;
+    value?: string;
```
captDaylight, line 36: we weren't capturing the value for dob input!!

jknapp25 (approved):

samkotlove (approved):

PR #85, opened 2023-03-14
Feat update photon small bugs

captDaylight:
- ability to filter by prescription status in App (for peachy)
- ability to save patient’s address from the new order form
- fix edit Patient DOB
- fix phone input not capturing first keystroke
- fix visual issue with pending prescriptions
- app was displaying the wrong DOB on the patient page, it was a day behind because of a UTC error
- Fixed some lingering pre-populated forms when switching between editing and creating patients in app


jknapp25 (approved):
send it!

PR #103, opened 2023-03-21
Revert send to patient switch flip

captDaylight:


samkotlove:
We should try to keep PR titles descriptive for git history, updated this one

PR #126, opened 2023-03-23
Prescribe Workflow mobile Layout issues

captDaylight:
cc @mrado1 
- reduce padding around the form 
- reduce card padding
- checkboxes align to the top not center
- "advanced search" should be under treatment on mobile when stacked
- remove gap when on mobile, it adds weird spacing when stacked in column

before:
Screen Shot 2023-03-23 at 10 59 59 AM

after:
image

### DIFF
--- a/packages/elements/package.json
+++ b/packages/elements/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@photonhealth/elements",
-  "version": "0.0.94",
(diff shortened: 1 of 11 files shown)


jknapp25:
a bit close?
image

jknapp25 (approved):

PR #135, opened 2023-03-24
Login with unaffiliated account should show error message

### LINKED ISSUE #125: App: display errors with login for wrong accounts
Recreate:
- logout of google accounts
- logout of photon
- try logging into photon
- after attempting to login it kicks you back to login without an error message
image

captDaylight:
https://www.loom.com/share/a5e364e1bf534331ad5a2b19e65814df

Screen Shot 2023-03-24 at 5 30 53 PM

closes #125

### DIFF
--- a/apps/app/src/hooks/useQuery.ts
+++ b/apps/app/src/hooks/useQuery.ts
@@ -1,0 +1,9 @@
+import { useMemo } from 'react';
+import { useLocation } from 'react-router-dom';
+
+function useQuery() {
+  const { search } = useLocation();
+  return useMemo(() => new URLSearchParams(search), [search]);
+}
+
+export default useQuery;
--- a/apps/app/src/views/routes/Login.tsx
+++ b/apps/app/src/views/routes/Login.tsx
@@ -25,7 +25,7 @@ export const Login = () => {
   const location = useLocation() as any;
 
   // Handle invite with redirect, even if logged in
-  const [searchParams, setSearchParams] = useSearchParams();
+  const [searchParams] = useSearchParams();
   const invite = searchParams.get('invitation');
   const org = searchParams.get('organization');
 
--- a/apps/app/src/views/routes/Login.tsx
+++ b/apps/app/src/views/routes/Login.tsx
@@ -25,9 +25,10 @@ export const Login = () => {
   const location = useLocation() as any;
 
   // Handle invite with redirect, even if logged in
-  const [params] = useSearchParams();
-  const invite = params.get('invitation');
-  const org = params.get('organization');
+  const [searchParams, setSearchParams] = useSearchParams();
+  const invite = searchParams.get('invitation');
+  const org = searchParams.get('organization');
+
   if (invite && org) {
     login({
       organizationId: org,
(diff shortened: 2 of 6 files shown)


apps/app/src/views/routes/Login.tsx
```diff
@@ -51,7 +53,13 @@ export const Login = () => {
             </Alert>
           )}
           <Stack spacing={{ base: '2', md: '3' }} textAlign="center">
-            <Heading size={breakpoint}>Log in to your account</Heading>
+            <Heading size={breakpoint}>Lorg in to your account</Heading>
```
jknapp25, line 56: Lorg

apps/app/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,9 @@
+import { useMemo } from 'react';
+import { useLocation } from 'react-router-dom';
+
+function useQuery() {
```
jknapp25, line 4: query makes me think graphql, maybe useQueryParams?

jknapp25 (approved):
Approving with small comments

apps/app/src/views/routes/SelectOrg.tsx
```diff
@@ -24,18 +25,20 @@ export const SelectOrg = () => {
...
-      setOrganization(organizations[0].id);
-      login({
-        appState: {
-          returnTo: from
-        }
-      });
+  useEffect(() => {
+    if (!loading) {
+      if (organizations?.length === 0) {
+        logout({ returnTo: `${window.location.origin}?auth0-orgs=0` });
```
samkotlove, line 31: can we make the param just orgs?

apps/app/src/hooks/useQuery.ts
captDaylight: ya good point

apps/app/src/views/routes/Main.tsx
```diff
@@ -61,7 +64,9 @@ export const Main = () => {
   }
 
   if (!isAuthenticated && !isLoading) {
-    return <Navigate to="/login" state={{ from: location }} replace />;
+    // forwarding on the query param to the login page
+    const to = `/login${query.get('auth0-orgs') === '0' ? '?auth0-orgs=0' : ''}`;
```
samkotlove, line 68: can we just forward all query params? the case i think would be (and good to test):
- i intend to go to like `/orders/new?patientId=1234`
- i click login
- I select my personal account by mistake
- i see the error message, i click login again
- I select my work email
- i have an org and get direct to the creat eorder screen for that patient

captDaylight: I updated this to forward all query params, but we're achieving the "return to" after login in the next line by setting state:
```
return <Navigate to={to} state={{ from: location }} replace />;
```
The issue I'm running into trying to pass along location state after a failed login is that when the user is logged out after signing in to the wrong account, the `returnTo` url has to be set in the auth0 admin. In an ideal scenario, I'd just redirect the user to the route they wanted to go after logout... which would kick them back to login and track the location state.

captDaylight: solved this by passing another query param `pathname` along back to the redirect url

PR #136, opened 2023-03-24
fix(pharmacy selector): showing preferred pickup when only mail order is enabled

jknapp25:
The pharmacy selector was showing a preferred pickup pharmacy for a peachy patient when they only have mail order enabled.

image

After changes:

https://user-images.githubusercontent.com/3934326/227646750-be8d4976-e453-4be0-96ea-d75205c0efce.mov


captDaylight (approved):
looks good!

PR #142, opened 2023-03-27
Elements: allow for editing preconfigured template-ids

captDaylight:
If the prescribe workflow has `template-ids` the prescribe form is hidden. This will hide the ability to edit the prescription if the prescriber clicks the edit button.

Once the form is opened it can't be closed again.


jknapp25 (approved):
Nice!

PR #149, opened 2023-03-28
feat(patient): status chat alert prompt

jknapp25:
- add patient status page chat prompt
- update chakra (double-checked with app)

image


apps/patient/src/utils/text.json
```diff
@@ -37,8 +37,12 @@
...
     "subheading": "After your prescriptions are sent to a pharmacy, we'll send you updates on their status.",
     "pickup": "Pick Up:",
+    "chat": {
+      "prompt": "At the pharmacy? Let us know if you're having any issues.",
+      "cta": "Text us now"
+    },
     "SENT": {
-      "title": "Sent to pharmacy",
+      "title": "Sent to Pharmacy",
       "description": "Your prescription fills have been sent to the pharmacy."
```
samkotlove, line 46: Unrelated, this copy seems a little strange

apps/patient/src/views/Status.tsx
```diff
@@ -101,6 +101,8 @@ export const Status = () => {
 
   const { fulfillment, pharmacy, organization } = order
 
+  const photonPhone = '5138663212'
```
samkotlove, line 104: Is this the google voice?

samkotlove (approved):

apps/patient/src/utils/text.json
jknapp25: Just matching casing on status titles

PR #150, opened 2023-03-28
Add courier option (Capsule) to pharmacy selection

jknapp25:
- added capsule
- added alto logo and id's for near-future use
- courier-specific copy on status view
- limited to MoPed and Austin TX

Example:

https://user-images.githubusercontent.com/3934326/230501497-d7a0f8f7-7cdf-401a-bc2b-5c93ae41b24d.mov


apps/patient/src/views/Status.tsx
```diff
@@ -35,13 +35,18 @@ export const Status = () => {
   const [searchParams] = useSearchParams()
   const orderId = searchParams.get('orderId')
   const token = searchParams.get('token')
+  const courier = searchParams.get('courier')
```
samkotlove, line 38: Do we need this if we have the check for the pharmacy id?

samkotlove (approved):

apps/patient/src/views/Status.tsx
jknapp25: There's no pharmacy details on the order if we immediately transition to the status view...

image

PR #151, opened 2023-03-29
fix(deps): housekeeping

jknapp25:
- non-dev dependency updates
- de-dupe

Verified app and patient are working as usual


.vscode/settings.json
```diff
@@ -19,5 +19,6 @@
     "source.fixAll.eslint": true,
     "source.fixAll.stylelint": true
   },
-  "editor.defaultFormatter": "esbenp.prettier-vscode"
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "typescript.tsdk": "node_modules/typescript/lib"
```
jknapp25, line 23: designates workspace version as the ts to use instead of global, solution for [this issue](https://github.com/chakra-ui/chakra-ui/issues/3714)

captDaylight (approved):
love it!

PR #158, opened 2023-03-29
Elements: Check for consent 

captDaylight:
You cannot skip consent on localhost as it is never a verifiable first party app, so this will call `getAccessTokenWithConsent` if `getAccessToken` fails.


packages/elements/src/photon-multirx-form-wrapper/index.tsx
```diff
@@ -42,7 +42,13 @@ customElement(
     const { actions: patientActions } = PatientStore;
 
     onMount(async () => {
-      const token = await client!.getSDK().authentication.getAccessToken();
+      let token;
```
samkotlove, line 45: Can we reuse the function where this check already exists?

PR #174, opened 2023-03-30
feat: cancel order and prescription

jknapp25:
- cancel order
- cancel prescription
- align order details and fulfillment table columns
- CopyText comp


apps/app/src/views/routes/Prescription.tsx
```diff
@@ -70,21 +72,49 @@ export const Prescription = () => {
...
-          <Button
-            aria-label="New Order"
-            as={RouterLink}
-            to={`/orders/new?patientId=${patient.id}&prescriptionId=${id}`}
-          >
-            Create Order
-          </Button>
-        )}
+        <HStack>
+          {loading ? (
```
samkotlove, line 81: These buttons (included create order), should only show if prescription is ACTIVE

jknapp25: added disabling to both

apps/app/src/views/routes/Order.tsx
```diff
@@ -85,11 +121,73 @@ export const Order = () => {
...
+        {loading ? (
+          <Skeleton width="112px" height="32px" borderRadius="md" />
+        ) : (
+          <Button
+            size="sm"
+            aria-label="Cancel Order"
+            isDisabled={order.fulfillment?.type !== types.FulfillmentType.MailOrder}
+            onClick={async () => {
+              const decision = await confirmWrapper('Cancel this order?', {
+                description: 'You will not be able to recover this order.',
```
samkotlove, line 168: Feel like we can modify the copy a little, maybe "You will not be able to undo this" or something?

jknapp25: updated

samkotlove (approved):

Issue #168, opened 2023-03-30
Misalignment address form


captDaylight:
<img width="568" alt="Screen Shot 2023-03-30 at 10 04 14 AM" src="https://user-images.githubusercontent.com/700617/228861775-22876b70-36da-45b5-8b8a-34af59d8696b.png">

PR #180, opened 2023-03-31
Neutron: Patient Datepicker

captDaylight:
- patient datepicker can't be in future, initializes empty when creating patient

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29681,7 +29681,6 @@
       "version": "0.0.98",
       "license": "ISC",
       "dependencies": {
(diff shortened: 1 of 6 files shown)


jknapp25:
Feature works, but margin seems to do a bit of strange stuff when errors show?

image

jknapp25 (approved):
Approving with comment, probably to be fixed in a later boson commit

captDaylight:
Good find @jknapp25 thx
https://github.com/Photon-Health/client/issues/181
https://github.com/Photon-Health/client/issues/168

PR #193, opened 2023-04-04
Prevent Prescription Quantity set to 1 

### LINKED ISSUE #172: Wrong Count of 1 When creating prescriptions
Providers are putting in a count of 1 for prescriptions (which is wrong). Checked the API request and we are getting 1 from them, just wanna be sure that our frontend isnt somehow making this 1

captDaylight:
If you select a medication rather than a treatment it should not automatically pass validation.

Screen Shot 2023-04-04 at 10 57 59 AM

closes #172

### DIFF
--- a/packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
+++ b/packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
@@ -139,7 +139,7 @@ export const AddPrescriptionCard = (props: {
             <photon-number-input
               class="flex-grow flex-1 w-2/5 sm:w-auto"
               label="Quantity"
-              value={props.store['dispenseQuantity']?.value ?? 0}
+              value={props.store['dispenseQuantity']?.value ?? null}
               required="true"
               min={1}
               invalid={props.store['dispenseQuantity']?.error ?? false}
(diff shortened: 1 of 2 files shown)


packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -212,7 +212,7 @@ export const AddPrescriptionCard = (props: {
           <photon-number-input
             class="flex-grow flex-shrink flex-1"
             label="Refills"
-            value={props.store['refillsInput']?.value ?? 0}
+            value={props.store['refillsInput']?.value ?? null}
```
samkotlove, line 215: This one I think we can leave as 0

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -322,19 +322,19 @@ export const AddPrescriptionCard = (props: {
                   });
                   props.actions.updateFormValue({
                     key: 'refillsInput',
-                    value: 0
+                    value: null
```
samkotlove, line 325: same as above

samkotlove (approved):
1 small comment but looks good!

PR #194, opened 2023-04-04
Neutron: New Prescription Default values

captDaylight:
- Selecting a medication (rather than a treatment) no longer automatically passes validation


jknapp25 (approved):
nice catch!

PR #196, opened 2023-04-04
Miscellaneous send to patient updates

jknapp25:
- swap Radish green logo with red one they shared
- truncate hours to fix showing two lines on small mobile with day of week suffix
- still show hours when no ratings available


apps/patient/src/configs/settings.ts
```diff
@@ -35,8 +35,8 @@ export const settings = {
     },
     // Radish Health
     [process.env.REACT_APP_RADISH_HEALTH_ORG_ID as string]: {
-      logo: 'radish_health_logo.svg',
-      accentColor: '#84a98c'
+      logo: 'Red-Radish-Logo.svg',
```
captDaylight, line 38: can we make - into _ and lower case?
Screen Shot 2023-04-04 at 3 09 46 PM

captDaylight (approved):

PR #206, opened 2023-04-10
Display Blocking Order Address Errors

### LINKED ISSUE #205: Address Error When Submitting Order
If the address is invalid, you can’t send an order (which is correct) but the error doesn’t show up until you click edit

image
image

captDaylight:
- Elements: fixes validation on the update / create so that incorrect zip codes don't get introduced
- zip codes use the format of either 12345 or 12345-1234
- App: the address form will open up and display the error if theres something blocking sending the order

Can use the below patient to test
Screen Shot 2023-04-10 at 2 14 43 PM

closes #205

### DIFF
--- a/packages/elements/src/photon-patient-form/index.tsx
+++ b/packages/elements/src/photon-patient-form/index.tsx
@@ -6,7 +6,7 @@ import { createFormStore } from '../stores/form';
 import { PatientStore } from '../stores/patient';
 import { PharmacyStore } from '../stores/pharmacy';
 import tailwind from '../tailwind.css?inline';
-import { email, empty, message, numericString, notFutureDate } from '../validators';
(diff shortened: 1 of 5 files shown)


packages/elements/src/photon-patient-form/index.tsx
```diff
@@ -88,10 +88,7 @@ customElement(
     });
     actions.registerValidator({
       key: 'address_zip',
-      validator: message(
-        union([size(numericString(), 0, 10), empty()]),
-        'Please enter a valid zip code...'
-      )
+      validator: message(union([zipString(), empty()]), 'Please enter a valid zip code...')
```
samkotlove, line 91: 🔥

samkotlove (approved):

PR #209, opened 2023-04-10
Photon: Fix Order Validation 

captDaylight:
- Elements: fixes validation on the update / create so that incorrect zip codes don't get introduced
- zip codes use the format of either 12345 or 12345-1234
- App: the address form will open up and display the error if theres something blocking sending the order


jknapp25 (approved):
lgtm!

Issue #178, opened 2023-03-31
Check for Duplicate dependencies 


captDaylight:

captDaylight:
Locking `package/components/package.json` broke elements. Need to clean whatever is happening there too

PR #215, opened 2023-04-11
SolidJS Pharmacy Search

captDaylight:
- added "particle" components, all tested with storybook 
- implementation of PharmacySearch "system" component

Todos
- [ ] revisit state management
- [ ] modal location selector
- [ ] handle Google map errors, ability to pass in your own Google API key
- [ ] wrap as an element (last time I had issues with reactivity)
- [ ] incorporate patients' preferred pharmacy list
- [ ] Playwright interactive testing, need to make a dummy gmail account for authentication


packages/components/package.json
```diff
@@ -32,12 +31,14 @@
...
   "dependencies": {
-    "@tailwindcss/typography": "^0.5.9",
     "@photonhealth/sdk": "*",
+    "@tailwindcss/forms": "^0.5.3",
+    "@tailwindcss/typography": "^0.5.9",
     "clsx": "^1.2.1",
     "graphql": "^16.6.0",
     "graphql-tag": "^2.12.6",
     "jwt-decode": "^3.1.2",
+    "solid-heroicons": "^3.1.1",
```
samkotlove, line 41: Should we lock these versions?

packages/components/src/particles/Button/index.tsx
```diff
@@ -1,52 +1,36 @@
...
-    size === 'xl' && 'text-lg py-4 px-6',
-    size === 'lg' && 'text-base py-3 px-4',
-    size === 'md' && 'text-sm py-2 px-3',
-    size === 'sm' && 'text-xs py-1.5 px-2.5',
-    size === 'xs' && 'text-xs py-1 px-2',
-    disabled
-      ? 'opacity-50 cursor-not-allowed'
-      : 'hover:bg-primary500 focus-visible:outline-primary600'
-  );
+  const buttonClasses = clsx('font-semibold shadow-sm', {
```
samkotlove, line 15: What's going on here?

samkotlove: Nvm think i follow

packages/components/src/particles/Button/index.tsx
```diff
@@ -1,52 +1,36 @@
...
+    'rounded-md': ['xl', 'lg', 'md'].includes(merged.size),
+    rounded: !['xl', 'lg', 'md'].includes(merged.size),
+    'text-white bg-indigo-600 hover:bg-indigo-500 focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 hover:bg-indigo-500':
+      merged.variant === 'primary',
+    'rounded bg-white text-gray-900 shadow-sm ring-1 ring-inset ring-gray-300 hover:bg-gray-50':
```
samkotlove, line 20: How would we handle dynamic theming? (very future problem, but just thinking if it's worth preparing for, like centralizing colors or something)

packages/components/src/particles/ComboBox/ComboBox.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+    });
+
+    return (
+      <ComboBox {...args}>
+        <ComboBox.Input onInput={(e) => setQuery((e.target as HTMLInputElement).value)} />
+        <ComboBox.Options>
+          <For each={filteredPeople()}>
+            {(person) => (
+              <ComboBox.Option key={person.id} value={person.name}>
+                {person.name}
```
samkotlove, line 37: Do we need value and child value?

packages/components/package.json
captDaylight: yup, locked em up

packages/components/src/particles/Button/index.tsx
captDaylight: I set up the tailwind config, with a package[ tw-colors](https://www.npmjs.com/package/tw-colors). An earlier version shows how you can theme the buttons with this, and turns out to be pretty trivial change with find and replace text once we think through the variable setup. I paused on rolling this out until we get this first version not to slow things down and because there are a few tailwind theme libraries that are still low support, so I want to see how that changes over the next months before we go all in on one

https://github.com/Photon-Health/client/blob/438eec043904a7361ed136cc578a8c78f0612d38/packages/components/tailwind.config.cjs#L13

packages/components/src/particles/Button/index.tsx
```diff
@@ -1,52 +1,36 @@
...
+    'text-sm px-2.5 py-1.5': merged.size === 'md',
+    'text-sm px-2 py-1': merged.size === 'sm',
+    'text-xs px-2 py-1': merged.size === 'xs',
+    'opacity-50 cursor-not-allowed': merged.disabled
+  });
 
   return (
-    <button class={buttonClasses} type="button" disabled={disabled} onClick={handleClick}>
-      {children}
+    <button class={buttonClasses} type="button" {...merged}>
```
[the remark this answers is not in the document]
captDaylight: Good catch, best practice not to do that. solved with split props, looks a little cumbersome in Solid but this is how it do:

```
  const merged = mergeProps({ variant: 'primary', size: 'lg' }, props);
  const [otherProps, buttonProps] = splitProps(merged, ['variant', 'size', 'children']);
```

packages/components/src/particles/ComboBox/ComboBox.stories.tsx
captDaylight: I think this interface is going to change, specifically what values look like, but the child value might be different for example how it's happening inside of `PharmacySearch`:

``` jsx
<ComboBox.Option
  key={pharmacy.id}
  value={`${pharmacy.name}, ${formattedAddress}`}
>
  <div>{pharmacy.name}</div>
  <div class="text-xs">{formattedAddress}</div>
</ComboBox.Option>
```

packages/components/package.json
captDaylight: actually had to roll this back, ended up breaking elements, going to punt this with to https://github.com/Photon-Health/client/issues/178

apps/app/src/configs/fulfillment.neutron.ts
```diff
@@ -63,5 +59,25 @@ export const fulfillmentSettings: FulfillmentSettings = {
...
+  // MisterRx
+  [process.env.REACT_APP_MISTERRX_ORG_ID as string]: {
+    sendOrder: true,
+    pickUp: false,
+    mailOrder: true,
+    mailOrderProviders: [
+      ...(process.env.REACT_APP_CUREXA_ORG_ID ? [process.env.REACT_APP_CUREXA_ORG_ID] : [])
+    ],
+    sendToPatient: false,
+    sendToPatientUsers: []
```
samkotlove, line 81: might need to rebase or merge boson on? I think these changes are unrelated to yours

samkotlove (approved):
🚢

PR #240, opened 2023-04-21
Align Link List with Patient Page

captDaylight:
Screen Shot 2023-04-21 at 2 30 48 PM


samkotlove (approved):

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -175,7 +175,13 @@ export const AddPrescriptionCard = (props: {
                   dosageCalculatorRef.open = true;
                 }}
               >
-                <sl-icon name="calculator"></sl-icon>
+                <button
```
captDaylight, line 178: @samkotlove I'm sneaking this one in too, for some reason passing a DOM element rather than text is causing an issue with the hit detection on the icon.

PR #246, opened 2023-04-24
Swap Capsule with Alto

jknapp25:
image


apps/patient/.env.boson
```diff
@@ -11,7 +11,7 @@ REACT_APP_MODERN_RITUAL_ORG_ID=test3
 REACT_APP_RADISH_HEALTH_ORG_ID=test4
 REACT_APP_RIVER_HEALTH_ORG_ID=test5
 
-REACT_APP_CAPSULE_PHARMACY_ID=phr_01G9CM8XB71GYHQVFMZBXSVTBF
-REACT_APP_ALTO_PHARMACY_ID=phr_01G9CM93X1NFP1C9H9K50DPKHX
+REACT_APP_CAPSULE_PHARMACY_ID=courier1
+REACT_APP_ALTO_PHARMACY_ID=courier2
```
samkotlove, line 15: ?

jknapp25: re-added since they don't go anywhere on boson

apps/patient/src/components/CourierOptions.tsx
```diff
@@ -23,6 +26,7 @@ interface Props {
 }
 
 export const CourierOptions = ({
+  alto,
```
captDaylight, line 29: can we add the default props for alto and capsule here as opposed to below? bit more legible and streamlined

``` tsx
export const CourierOptions = ({
  alto = false,
  capsule = false, 
 ...
}
```

captDaylight (approved):

samkotlove (approved):

PR #247, opened 2023-04-24
Update neutron

jknapp25:
- Ability to set Address in Prescribe Workflow
- Swap Capsule with Alto

### DIFF
--- a/apps/patient/src/components/StatusStepper/index.tsx
+++ b/apps/patient/src/components/StatusStepper/index.tsx
@@ -1,6 +1,7 @@
 import { Box, Center, Container, Stack } from '@chakra-ui/react'
 import { Step } from './components/Step'
 import t from '../../utils/text.json'
+import { formatAddress } from '../../utils/general'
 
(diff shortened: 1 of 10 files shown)


apps/patient/src/components/StatusStepper/index.tsx
```diff
@@ -21,17 +23,25 @@ export const StatusStepper = ({ status, isCourier }: Props) => {
...
-                description={t.status[fulfillmentType].states[state].description}
-                isActive={currentStep === id}
-                isCompleted={currentStep > id}
-                isLastStep={states.length === id + 1}
-              />
-            ))}
+            {states.map((state, id) => {
+              const title = t.status[fulfillmentType].states[state].title
+              const description =
+                state === 'DELIVERING'
```
samkotlove, line 29: Where does delivering come from?

jknapp25: frontend

jknapp25: I'm just matching the 4 pickup states to 4 courier states

samkotlove (approved):

PR #261, opened 2023-05-02
Display Template Name

captDaylight:
Choosing to display the template name before the treatment name, my thinking is to me to lead with the disambiguation. 

Screen Shot 2023-05-02 at 11 56 31 AM

Caveat:
When selected, we are passing around the treatment data, so when we display it in the selector or in Pending Prescriptions section it doesn't have access to the template data at the moment.
Screen Shot 2023-05-02 at 11 58 34 AM
Screen Shot 2023-05-02 at 11 59 02 AM

### DIFF
--- a/packages/elements/src/photon-treatment-select/index.tsx
+++ b/packages/elements/src/photon-treatment-select/index.tsx
@@ -57,16 +57,11 @@ customElement(
 
     const getData = (filter: string): (Treatment | PrescriptionTemplate)[] => {
       if (store.catalogs.data.length > 0) {
-        const data = props.offCatalogOption
-          ? [
-              props.offCatalogOption as Treatment,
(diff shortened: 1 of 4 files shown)


packages/elements/src/photon-treatment-select/index.tsx
```diff
@@ -123,36 +110,31 @@ customElement(
           hasMore={false}
           selectedData={props.selected ?? (props.offCatalogOption as Treatment)}
           displayAccessor={(t: Treatment | PrescriptionTemplate, groupAccess: boolean) => {
-            return 'name' in t ? (
+            return 'name' in t && !('treatment' in t) ? (
```
samkotlove, line 113: do we not need this part?
```
!('treatment' in t)
```

PR #262, opened 2023-05-02
Add order canceled event to webhook form

jknapp25:
Screen Shot 2023-05-02 at 9 26 31 AM


apps/app/src/views/routes/Settings/components/WebhooksForm.tsx
```diff
@@ -77,6 +77,7 @@ export const WebhooksForm = ({ isOpen, close }: WebhooksFormProps) => {
     'photon:order:fulfillment',
     'photon:order:completed',
     'photon:order:error',
+    'photon:order:canceled',
```
samkotlove, line 80: There's another spot in the code with the list of webhooks, do we need it there too?

samkotlove (approved):

PR #271, opened 2023-05-03
Handle error if medication already added to catalog

captDaylight:
If you re add a medication to the catalog that already exists it throws an error. This handles it silently.

### DIFF
--- a/packages/elements/src/photon-med-search-dialog/index.tsx
+++ b/packages/elements/src/photon-med-search-dialog/index.tsx
@@ -53,13 +53,17 @@ customElement(
             setLoading(true);
             if (addToCatalog()) {
               const addCatalogMutation = client!.getSDK().clinical.catalog.addToCatalog({});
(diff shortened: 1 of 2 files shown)


packages/elements/src/photon-med-search-dialog/index.tsx
```diff
@@ -53,13 +53,17 @@ customElement(
...
+              try {
+                await addCatalogMutation({
+                  variables: {
+                    catalogId: catalogId(),
+                    treatmentId: medication()?.id
+                  },
+                  awaitRefetchQueries: false
+                });
+              } catch {
+                // do nothing, this is expected if medication exists in catalog
```
samkotlove, line 65: should we log here?

captDaylight: Works for me!

samkotlove (approved):

PR #285, opened 2023-05-05
Disable order cancel button in canceled state

### LINKED ISSUE #267: Canceled button is still showing for orders that are in canceled state
Screen Shot 2023-05-02 at 5 16 52 PM
(From Neutron)
The canceled button shouldn't show (or should be disabled) for orders that have already been canceled

jknapp25:
closes #267


captDaylight (approved):
HAWT

PR #291, opened 2023-05-09
Flatten fill list on Orders Page

captDaylight:
before
image
after
Screen Shot 2023-05-09 at 10 04 30 AM

### DIFF
--- a/apps/app/src/views/routes/Orders.tsx
+++ b/apps/app/src/views/routes/Orders.tsx
@@ -95,11 +95,11 @@ const renderRow = (order: any) => {
   const extId = order.externalId || <Text as="i">None</Text>;
 
   const fills = order.fills
-    .reduce((prev: any[], curr: any) => {
+    .reduce((prev: any[], cur: any) => {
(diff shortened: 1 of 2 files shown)


apps/app/src/views/routes/Orders.tsx
```diff
@@ -94,10 +94,19 @@ const renderRow = (order: any) => {
...
+      const exists = prev.some((fill) => fill.prescription.id === cur.prescription.id);
+      if (!exists) {
+        prev.push(cur);
+      }
+      return prev;
+    }, [])
+    .reduce((prev: string, cur: any) => {
+      const fill = cur.treatment.name;
+      return prev ? `${prev}, ${fill}` : fill;
+    }, '');
```
samkotlove, line 109: Could we not just do like
```
// Create a set of names
const medicationNames = new Set();
// Add names to set
order.fills.map(item => medicationNames.add(item.treatment.name))
// Turn into array
const names = Array.from(medicationNames)
```
Or something? Seems like the reduce's create a lot of unnecessary looping

PR #323, opened 2023-05-22
Bug/number input

captDaylight:
https://stackoverflow.blog/2022/12/26/why-the-number-input-is-the-worst-input/

and we have the numerical keyboard to boot

image

### DIFF
--- a/packages/components/src/systems/DoseCalculator/index.tsx
+++ b/packages/components/src/systems/DoseCalculator/index.tsx
@@ -35,10 +35,9 @@ export interface DoseCalculatorProps {
 type SignalSetter = (v: number) => void;
 
 const handleInput: (setter: SignalSetter) => (e: Event) => void = (setter) => (e) => {
-  const target = e.target as HTMLInputElement;
-  const val = parseFloat(target.value);
-  if (!isNaN(val)) {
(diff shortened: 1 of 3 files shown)


packages/components/src/systems/DoseCalculator/index.tsx
```diff
@@ -32,6 +32,15 @@ export interface DoseCalculatorProps {
   }) => void;
 }
 
+type SignalSetter = (v: number) => void;
+
+const handleInput: (setter: SignalSetter) => (e: Event) => void = (setter) => (e) => {
+  const target = e.currentTarget as HTMLInputElement;
```
samkotlove, line 38: Can we leave a comment as to why we do it this way?

packages/components/src/systems/DoseCalculator/index.tsx
```diff
@@ -32,6 +32,15 @@ export interface DoseCalculatorProps {
   }) => void;
 }
 
+type SignalSetter = (v: number) => void;
+
+const handleInput: (setter: SignalSetter) => (e: Event) => void = (setter) => (e) => {
```
samkotlove, line 37: Should this be handled within the input component itself?

PR #340, opened 2023-05-24
Fix error reading "patient"

captDaylight:
https://photon-health.sentry.io/issues/4171455131/?project=6641717&referrer=slack

bigger question is why do some prescriptions not have patients


jknapp25:
`TypeError: Cannot read properties of null (reading 'patient')` sounds more like prescription isn't present sometimes to me

jknapp25 (approved):

PR #351, opened 2023-05-26
Remove asap ready-by option

jknapp25:
For test details [see here](https://www.notion.so/photons/Remove-ASAP-option-67a0fd490acd43eb8a8f224563709995)

image

### DIFF
--- a/apps/patient/src/views/ReadyBy.tsx
+++ b/apps/patient/src/views/ReadyBy.tsx
@@ -12,7 +12,6 @@ import t from '../utils/text.json';
 import { OrderContext } from './Main';
 
 const checkDisabled = (option: string): boolean => {
-  if (option === 'tomorrow') return false;
   const currentTime = dayjs();
   const afterHoursOption = t.readyBy.options[5];
   const afterHoursStarts = '6:00 pm';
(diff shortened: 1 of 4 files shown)


apps/patient/src/views/ReadyBy.tsx
```diff
@@ -56,7 +56,7 @@ export const ReadyBy = () => {
 
           <VStack spacing={3} w="full">
             {t.readyBy.options.map((option, i) => {
-              const isDisabled = i !== (0 | 6) && currentTimeIsAfterOption(option);
+              const isDisabled = i !== 6 && currentTimeIsAfterOption(option);
```
captDaylight, line 59: better way to do this than using a hard coded index?

jknapp25: fixed

apps/patient/src/views/ReadyBy.tsx
```diff
@@ -11,7 +11,8 @@ import { PoweredBy } from '../components/PoweredBy';
 import t from '../utils/text.json';
 import { OrderContext } from './Main';
 
-const currentTimeIsAfterOption = (option: string): boolean => {
+const checkDisabled = (option: string): boolean => {
+  if (option === 'tomorrow') return false;
```
captDaylight, line 15: will this catch if the text is "Tomorrow"? is it lowercased somewhere?

jknapp25: ah interesting, works without upper case, removing

PR #374, opened 2023-06-05
Show after hours option at all times

jknapp25:
We were disabling the after hours option after 6pm but this doesn't really make sense in case it's 7pm and you do want it after hours. I'm not sure what the original idea behind this one was, so removing this limitation.

### DIFF
--- a/apps/patient/src/views/ReadyBy.tsx
+++ b/apps/patient/src/views/ReadyBy.tsx
@@ -13,8 +13,8 @@ import { OrderContext } from './Main';
 
 const checkDisabled = (option: string): boolean => {
   const currentTime = dayjs();
-  const timetoCheckDayjs = dayjs(option, 'h:mm a');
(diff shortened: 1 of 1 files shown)


apps/patient/src/views/ReadyBy.tsx
```diff
@@ -13,10 +13,7 @@ import { OrderContext } from './Main';
 
 const checkDisabled = (option: string): boolean => {
   const currentTime = dayjs();
-  const afterHoursOption = t.readyBy.options[5];
-  const afterHoursStarts = '6:00 pm';
-  const timetoCheck = option === afterHoursOption ? afterHoursStarts : option;
-  const timetoCheckDayjs = dayjs(timetoCheck, 'h:mm a');
+  const timetoCheckDayjs = dayjs(option, 'h:mm a');
```
captDaylight, line 16: nit, can this be `timeToCheckDayJs`

captDaylight (approved):

PR #379, opened 2023-06-07
Enable mail order statuses for River Health

jknapp25:
- Frontend prep for showing mail order status page for River Health
- Additional cleanup of fulfillment types

### DIFF
--- a/apps/patient/src/utils/general.ts
+++ b/apps/patient/src/utils/general.ts
@@ -84,15 +84,62 @@ export const getHours = (
   };
 };
 
+const COMMON_COURIER_PHARMACY_IDS = [
+  // Capsule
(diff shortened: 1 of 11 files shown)


apps/patient/src/utils/general.ts
```diff
@@ -84,33 +84,78 @@ export const getHours = (
   };
 };
 
+const COMMON_COURIER_PHARMACY_IDS = [
```
captDaylight, line 87: 😬

captDaylight (approved):

apps/patient/src/utils/general.ts
jknapp25: lol

PR #380, opened 2023-06-07
Use GraphQL directly for Location Search

captDaylight:
This is a PoC for the Pharmacy Search to see how it feels to use GraphQL directly in our components rather than rely on complicated calls buried in stores or SDK. IMO I like how it feels.

Do we need Apollo? Can we a more minimal client (and not dedicated to React) like URQL or this [other client](https://github.com/Photon-Health/client/blob/boson/apps/patient/src/configs/graphqlClient.ts) in patient (though I do like the caching that URQL and Apollo prevides)?  @jknapp25 you added https://github.com/jasonkuhrt/graphql-request, how do you like it?


jknapp25:
`graphql-request` seemed lightweight, not married too it though, just chose it for speed

packages/components/src/systems/PharmacySearch/index.tsx
```diff
@@ -16,31 +32,50 @@ export interface PharmacyProps {
...
+  const [location, setLocation] = createSignal<Location | null>(null);
+  const [pharmacies, setPharmacies] = createSignal<Pharmacy[] | null>(null);
+  const [fetching, setFetching] = createSignal(false);
+  const [openLocationSearch, setOpenLocationSearch] = createSignal(false);
+
+  async function fetchPharmacies() {
+    const { data } = await client!.sdk.apollo.query({
+      query: GetPharmaciesQuery,
+      variables: {
+        location: { latitude: location()?.latitude, longitude: location()?.longitude, radius: 20 }
```
enteladuka, line 47: Might want to double check that if you don't pass in a radius it defaults to something in the backend

enteladuka (approved):

PR #397, opened 2023-06-26
Pharmacy Select Component

captDaylight:
[Demo 📺](https://www.loom.com/share/e4ee66b43c154481b5bba84d581ca12f?sid=2f904476-1304-4fc5-a431-95ca6106fafe)

Screen Shot 2023-06-30 at 12 03 06 PM

#### Test
Local
- [x] Patient with Address set
- [x] patient without Address Set
- [x] Patient with preferred Pharmacy

Send to Patient
- [x] check that order made with undefined fulfillment type
- [x] Receive text

MailOrder
- [x]  if `pharmacy-id` is set, it should not display the pharmacy select

App
- [x] can create prescriptions

## New Components
[See them live here](https://www.chromatic.com/library?appId=640675659e06690283898857)

### `<Card />`
Creates a simple bordered card. 

Screen Shot 2023-06-30 at 8 00 14 AM

Multiple children passed results in card with sections.

Screen Shot 2023-06-30 at 8 01 09 AM

### `<RadioGroup />`
Uses radio inputs under the hood for accessibility. `RadioGroup.Option` wraps each element in the group.

``` tsx
        <RadioGroup
          label="Pharmacies"
          setSelected={(s) => console.log('selected: ', s)}
          initSelected={pharmacies[1].id}
        >
          {pharmacies.map(({ id, name, address }) => (
            <RadioGroup.Option value={id}>
              <div>
                <div class="mr-4">{name}</div>
                <div class="text-sm text-slate-500">{address}</div>
              </div>
            </RadioGroup.Option>
          ))}
        </RadioGroup>
```

Screen Shot 2023-06-30 at 7 58 48 AM

### `<Tabs />`

Screen Shot 2023-06-30 at 8 02 29 AM

### `<Text />`
Pass in `loading` and  `sampleLoadingText` along with `size` and you get Skeleton text while you wait for values to load.

Frame 19 (1)

### `<Badge />`

Screen Shot 2023-06-30 at 8 46 28 AM
Screen Shot 2023-06-30 at 8 49 31 AM

### Dropdown direction
Now if you're near the bottom of the viewport the dropdown goes up.

Screen Shot 2023-06-30 at 10 26 03 AM


jknapp25:
nit: the pharmacy select input looks a bit inconsistent in terms of size/padding compared to above days supply, etc

jknapp25:
At first glance, this UI is still a bit confusing too me. It's not immediately clear that the preferred tag is associated with the value in the input. Might have to do with input text being a bit too gray, feels like it's placeholder text?

image

jknapp25:
Do you think wrapped text would be clearer?

image

jknapp25:
What do up/down carets do?

Screen Shot 2023-06-30 at 9 27 14 AM

packages/components/src/utils/getLocation.ts
```diff
@@ -0,0 +1,27 @@
...
+  addressOrLocation: string | Coordinates,
+  geocoder: google.maps.Geocoder
+): Promise<Location> => {
+  const data = await geocoder!.geocode({
+    ...(typeof addressOrLocation === 'string'
+      ? { address: addressOrLocation }
+      : { location: { lat: addressOrLocation.latitude, lng: addressOrLocation.longitude } })
+  });
+
+  const latitude = data.results[0].geometry.location.lat();
```
jknapp25, line 20: If geocode fails will these error?

captDaylight:
> nit: the pharmacy select input looks a bit inconsistent in terms of size/padding compared to above days supply, etc

@jknapp25 good eye, the "days supply" is the existing elements, written in Shoelace. The new select input is using [TailwindUI](https://tailwindui.com/) as a guide, so there will be some inconsistencies until we're able to swap out all of the components.

captDaylight:
> Ah, drop-up/down. Any way to keep these from overlapping text?

@jknapp25 that's a good point, I'm going to lop off the "Brooklyn, NY" since it's repeated below.

Screen Shot 2023-06-30 at 2 02 53 PM

packages/components/src/utils/getLocation.ts
captDaylight: @jknapp25 I'll need to think about this one, good catch. IMO we can release since this is how it's handled now, but we should track this with a task to show a error handler somehow

jknapp25:
@captDaylight will long street names wrap?

packages/components/src/utils/getLocation.ts
jknapp25: maybe optional chaining is a quick safety check?

jknapp25 (approved):
Great work @captDaylight !

samkotlove (approved):

PR #405, opened 2023-07-05
Clear form when selecting different treatments

captDaylight:
https://www.notion.so/photons/Clear-form-when-selecting-different-treatments-042a4537039f4e41aac74d53f43798e5?pvs=4

### DIFF
--- a/packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
+++ b/packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
@@ -316,34 +316,7 @@ export const AddPrescriptionCard = (props: {
                     'notes'
                   ]);
                   setOffCatalog(undefined);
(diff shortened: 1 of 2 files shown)


jknapp25 (approved):
Nice @captDaylight! Does the form have initial form values? Could we just revert to those?

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -314,34 +316,7 @@ export const AddPrescriptionCard = (props: {
                     'notes'
                   ]);
                   setOffCatalog(undefined);
-                  props.actions.updateFormValue({
```
captDaylight, line 317: @jknapp25 good callout on the init form values, I couldn't find it but I did find some duplicate code that could use the new function :)

PR #413, opened 2023-07-06
Better visibility on dedicated view buttons

jknapp25:
Re-used standard page view layout for order, prescription, patient

image


captDaylight (approved):
Looks amazing

PR #437, opened 2023-07-14
Add ability for patient to set preferred pharmacy

jknapp25:
In order to auto-route more orders we're letting patients set their preferred pharmacy in Navigate.

https://github.com/Photon-Health/client/assets/3934326/ce6e3fef-3cc0-43c1-b395-e970a42f3fbf


apps/patient/src/components/PharmacyCard.tsx
```diff
@@ -119,13 +126,16 @@ export const PharmacyCard = memo(function PharmacyCard({
...
-                as={pharmacy.info === 'Previous' ? FiRotateCcw : FiStar}
-              />
-              <TagLabel> {pharmacy.info}</TagLabel>
+          {preferred ? (
+            <Tag size="sm" colorScheme="yellow">
+              <TagLeftIcon boxSize="12px" as={FiStar} />
+              <TagLabel> Preferred</TagLabel>
+            </Tag>
+          ) : null}
+          {previous ? (
```
captDaylight, line 135: can we make this `previous && !preferred` had a chat with @mrado1 about this for the pharmacy search:
https://github.com/Photon-Health/client/blob/71b86c28b522a428d79df0863f5e2d8953868dff/packages/components/src/systems/PharmacySearch/index.tsx#L237

the gist is it's a little long and confusing when both badges are there

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -305,6 +300,62 @@ export const Pharmacy = () => {
     }
   };
 
+  const setPreferredPharmacy = async (patientId: string, pharmacyId: string) => {
```
captDaylight, line 303: wdyt about putting this mutation in an API folder? I read [this](https://www.apollographql.com/blog/apollo-client/architecture/client-side-architecture-basics/#should-we-include-graphql-mutations-in-our-presentation-components) and he suggests queries in component and mutations separate. I kinda like the logic

captDaylight (approved):

apps/patient/src/components/PharmacyCard.tsx
jknapp25: updated

apps/patient/src/views/Pharmacy.tsx
jknapp25: Yeah i like that! was looking for ways to cleanup, getting a bit crazy. Today is also crazy, but gonna try to get this out. I'll followup with some refactoring for this.

PR #439, opened 2023-07-14
Neutron: Template names, Patient Preferred

captDaylight:
- [Add and Edit Template Names In App #438](https://github.com/Photon-Health/client/pull/438)
- [Add ability for patient to set preferred pharmacy #437](https://github.com/Photon-Health/client/pull/437)


jknapp25:
@captDaylight this is waiting on set preferred backend in services

PR #468, opened 2023-07-21
Status page improvements

jknapp25:
Closes: [Update Chakra for improved stepper design](https://www.notion.so/photons/Update-Chakra-for-improved-stepper-design-225b8e0525a047d0a070e3cc45340b5e?pvs=4)

We've been receiving some confusion, especially from peachy, around the status view. It appears that people are primarily confused by the stepper currently showing the active step with a green dot in the middle. I think the confusion is mostly coming from peachy because they don't also do pharmacy selection so they are missing a bit of context not having already walked through the flow a bit.

- Better StatusStepper coloring
- Add tracking id link
- Update Chakra

image


apps/patient/src/views/Status.tsx
```diff
@@ -143,19 +143,25 @@ export const Status = () => {
...
+              <Text display="inline" color="gray.600">
                 {t.status.MAIL_ORDER.trackingNumber}
               </Text>
-              <Text display="inline" ms={2}>
+              <Link
+                href={`https://google.com/search?q=${fulfillment.trackingNumber}`}
+                display="inline"
+                ms={2}
+                color="brandLink"
+                fontWeight="medium"
```
captDaylight, line 154: can you add a ` target="_blank"`?

captDaylight (approved):

apps/patient/src/views/Status.tsx
jknapp25: good call!

PR #469, opened 2023-07-21
Update Rx Hierarchy

captDaylight:
https://www.notion.so/photons/Rx-Page-Info-Hierarchy-e155b8dba73046d390a26dc1e977d9ef?pvs=4

- [x] Could we move instructions and notes into the existing “prescription details” section? They are a part of the prescription details, not separate pieces
- [x] can we call “notes” -> “pharmacy notes”
- [x] could we move id and external id somewhere else? They are not a part of the prescription details so should be moved to a different part

(bonus New)
- [x] changed cancel button to red so it feels more dangerous ⚠️
- [x] get rid of table in favor of grid items so that they collapse to vertical stacked mobile, right now I have to scroll off the right side of the page to see the text
  - 
- [x] Lightened and shrunk the description/name column

# Rx Page 
### Desktop
Frame 21
### Mobile
Screen Shot 2023-07-21 at 3 12 59 PM

# Order Page (Bonus)
### Desktop
Frame 20 (1)

### Mobile
Screen Shot 2023-07-21 at 2 47 41 PM

### DIFF
--- a/apps/app/src/views/components/CopyText.tsx
+++ b/apps/app/src/views/components/CopyText.tsx
@@ -10,7 +10,6 @@ const CopyText = ({ text }: { text: string }) => {
         whiteSpace={{ base: 'nowrap', sm: 'normal' }}
         overflow={{ base: 'hidden', sm: 'visible' }}
         textOverflow={{ base: 'ellipsis', sm: 'clip' }}
-        maxWidth={{ base: '100px', sm: 'none' }}
       >
         {text}
       </Text>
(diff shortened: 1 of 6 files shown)


jknapp25:
Bit of extra space here, looks like hidden divider?
image

jknapp25:
Do you think we might want to leave the descriptors large on desktop? The small on mobile looks perfect, but feels a little misaligned to me on desktop when they're in the same row.

image

jknapp25:
nit: found some extra bottom padding on mobile

image

jknapp25:
These are great updates @captDaylight! Do you think we should do the patient page while we're at it? Looks a bit inconsistent now.

jknapp25:
pharmacy id could probably span more on mobile now

Screen Shot 2023-07-21 at 1 22 26 PM

captDaylight:
> Bit of extra space here, looks like hidden divider? image

fixed 👍

jknapp25 (approved):

PR #484, opened 2023-07-27
Session replay masking

jknapp25:
Mask any sensitive information from session replays

Learn about masking here: https://docs.datadoghq.com/real_user_monitoring/session_replay/privacy_options/

PR #483, opened 2023-07-27
Ability to cancel order

captDaylight:
- [1. ability to cancel order ](https://www.notion.so/photons/Add-ability-to-cancel-an-order-014f9cef17134624aa5124d2df84157a?pvs=4)
Screen Shot 2023-07-27 at 2 13 37 PM

- [2. destructive actions to the bottom of page](https://www.notion.so/photons/Move-destructive-actions-to-bottom-of-Pages-3187cfaa37b048ec9ea94fdb7dc1b51d?pvs=4)
Screen Shot 2023-07-27 at 2 04 52 PM

PR #488, opened 2023-07-31
Photon: Cancel Order

captDaylight:
- [Ability to cancel order #483](https://github.com/Photon-Health/client/pull/483)
- [Session replay masking #484](https://github.com/Photon-Health/client/pull/484)
- [Set up Support Page #485](https://github.com/Photon-Health/client/pull/485)
- [Select default mail order if there is only one #486](https://github.com/Photon-Health/client/pull/486)


samkotlove (changes requested):
One sec, need to make sure photon lambda is good to go
https://github.com/Photon-Health/terraform/pull/1231

PR #492, opened 2023-07-31
Photon: Ability to cancel order, create order disabling, replay masking, etc

jknapp25:
Ability to cancel order (#483)
Disable create order when rx depleted (https://github.com/Photon-Health/client/pull/489)
Add Carbon Health branding (https://github.com/Photon-Health/client/pull/490)
Session replay masking (#484)
Set up Settings page (#485)
Select default mail order if there is only one (#486)


captDaylight:
@samkotlove good to merge to photon? https://github.com/Photon-Health/terraform/pull/1231

samkotlove (changes requested):
needs infras deploy to prod

samkotlove (approved):
All good

PR #493, opened 2023-08-02
Duplicate renew

captDaylight:
https://www.notion.so/photons/Renew-Prescription-65236d222d9d44c4be3f56b8e0683ea6?pvs=4
Screen Shot 2023-08-02 at 3 45 15 PM

(bonus)
https://www.notion.so/photons/Prescriptions-in-depleted-show-canceled-message-befa7a9fb43345f4b03fe04060830ea2?pvs=4

### DIFF
--- a/packages/sdk/src/types.ts
+++ b/packages/sdk/src/types.ts
@@ -583,7 +583,8 @@ export type PrescriptionInput = {
 export enum PrescriptionState {
   Active = 'ACTIVE',
   Depleted = 'DEPLETED',
-  Expired = 'EXPIRED'
+  Expired = 'EXPIRED',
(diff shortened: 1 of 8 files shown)


packages/sdk/src/types.ts
```diff
@@ -583,7 +583,8 @@ export type PrescriptionInput = {
 export enum PrescriptionState {
   Active = 'ACTIVE',
   Depleted = 'DEPLETED',
-  Expired = 'EXPIRED'
+  Expired = 'EXPIRED',
+  Canceled = 'CANCELED'
```
jknapp25, line 587: think we should extract types into their own package soon?

PR #497, opened 2023-08-03
Pass in Patient data to Prescribe Workflow

captDaylight:
Instead of relying on syncing our customer's data ahead of time, we can make patients on the fly by passing in their information like this:

``` html
<photon-prescribe-workflow
  patient='{"externalId":"123abc","firstName":"paul","lastName":"thomas","dateOfBirth":"2000-01-01","sex":"MALE","phone":"+17135609948","email":"paul@aoeu.com"}'
  address='{"city":"brooklyn","postalCode":"11221","state":"NY","street1":"372 Jefferson Ave"}'
></photon-prescribe-workflow>
```

### DIFF
--- a/packages/elements/src/photon-multirx-form/components/PatientCard.tsx
+++ b/packages/elements/src/photon-multirx-form/components/PatientCard.tsx
@@ -1,9 +1,10 @@
 import { message } from '../../validators';
 import { string, any, record } from 'superstruct';
-import { createSignal, onMount, Show, createEffect } from 'solid-js';
+import { createSignal, onMount, Show, createEffect, createMemo } from 'solid-js';
 import { PatientStore } from '../../stores/patient';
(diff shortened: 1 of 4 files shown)


packages/elements/src/photon-multirx-form/components/PatientCard.tsx
```diff
@@ -50,15 +53,71 @@ export const PatientCard = (props: {
     }
   };
 
+  const fetchOrCreatePatient = async () => {
+    const { data } = await props.client!.getSDK().clinical.patient.getPatients({
+      name: `${props?.patient?.firstName} ${props?.patient?.lastName}`
+    });
```
samkotlove, line 59: I thnk we should be able to do a call to the `patient` query and pass in external id

PR #501, opened 2023-08-09
Duplicate prescription button disabled if lacking permissions

captDaylight:
https://www.notion.so/photons/Duplicate-prescription-button-should-be-disabled-if-i-can-t-prescribe-ba7ffb415189428c9f98602fa43028fd?pvs=4


apps/app/src/hooks/usePermissions.ts
```diff
@@ -0,0 +1,37 @@
+import { usePhoton } from '@photonhealth/react';
+import jwtDecode from 'jwt-decode';
+import { Permission } from 'packages/sdk/dist/types';
+import { useEffect, useState } from 'react';
+
+function checkHasPermission(subset: Permission[], superset: Permission[]) {
```
jknapp25, line 6: is this the same idea?

`return subset.every(permission => superset.includes(permission));`

captDaylight: love it

apps/app/src/views/routes/Prescription.tsx
```diff
@@ -68,6 +69,7 @@ export const CANCEL_PRESCRIPTION = gql`
 export const Prescription = () => {
   const params = useParams();
   const id = params.prescriptionId;
+  const hasPermission = usePermissions(['write:prescription']);
```
jknapp25, line 72: Does write:prescription also require auth check?

jknapp25 (approved):
Cool!

apps/app/src/views/routes/Prescription.tsx
captDaylight: not sure I follow, these permissions are coming available after authentication

PR #503, opened 2023-08-10
Add patient rerouting

jknapp25:
- Add ability for patient to re-route order
- Api in components was getting unwieldy, pulled it out to get some organization
- Fetch 5 pharmacies on load and show more (increased from 3)
- Fixed pharmacy incorrectly showing "Closed" when it's just not returning google details
- Skip details fetch if business_status is not `OPERATIONAL`
- Immediately show pharmacy on /status using the selected pharmacy
- Add Blueberry pediatrics settings on photon and hide rerouting for their patients


apps/patient/src/utils/general.ts
```diff
@@ -188,3 +189,97 @@ export const countFillsAndRemoveDuplicates = (fills: types.Fill[]): FillWithCoun
...
+  const name = p.name;
+  const address = p.address ? formatAddress(p.address) : '';
+  const placeRequest = {
+    query: name + ' ' + address,
+    fields: ['place_id']
+  };
+  let placeId: string;
+  let placeStatus: string;
+  try {
+    const { response, status }: any = await query('findPlaceFromQuery', placeRequest);
```
jknapp25, line 222: move to api/internal

apps/patient/src/utils/general.ts
```diff
@@ -188,3 +189,97 @@ export const countFillsAndRemoveDuplicates = (fills: types.Fill[]): FillWithCoun
...
+      'opening_hours',
+      'utc_offset_minutes', // this gives us isOpen()
+      'rating',
+      'business_status'
+    ]
+  };
+  let details: any;
+  let detailsStatus: string;
+  try {
+    const { response, status }: any = await query('getDetails', detailsRequest);
```
jknapp25, line 249: move to api/internal

apps/patient/src/api/internal.ts
```diff
@@ -0,0 +1,162 @@
...
+  searchParams: {
+    latitude: number;
+    longitude: number;
+    radius: number;
+  },
+  limit: number,
+  offset: number,
+  token: string
+) => {
+  graphQLClient.setHeader('x-photon-auth', token);
```
captDaylight, line 28: I'm seeing some repetition here, maybe there are some reusable parts? in here... maybe this auth token could be set once and used throughout via a provider? 

Error handling could be standardized, done like this in some places:

```
      const { message } = error.response.errors[0];
      throw new Error(message);
```
or 
``` 
throw new Error(error.response.errors[0].message);
```

apps/patient/src/views/Status.tsx
```diff
@@ -45,54 +39,53 @@ export const Status = () => {
...
+    setSubmitting(true);
 
-      graphQLClient.setHeader('x-photon-auth', token);
-      const results: any = await graphQLClient.request(MARK_ORDER_AS_PICKED_UP, {
-        markOrderAsPickedUpId: orderId
-      });
+    try {
+      const result: boolean = await markOrderAsPickedUp(orderId, token);
 
       setTimeout(() => {
```
captDaylight, line 58: I know this not new, but I never noticed this timeout... what's it for?

apps/patient/src/utils/general.ts
```diff
@@ -188,3 +189,97 @@ export const countFillsAndRemoveDuplicates = (fills: types.Fill[]): FillWithCoun
...
+    fields: ['place_id']
+  };
+  let placeId: string;
+  let placeStatus: string;
+  try {
+    const { response, status }: any = await query('findPlaceFromQuery', placeRequest);
+    placeId = response[0]?.place_id;
+    placeStatus = status;
+  } catch (error) {
+    console.error(JSON.stringify(error, undefined, 2));
```
captDaylight, line 226: sup with double console?

apps/patient/src/views/Status.tsx
```diff
@@ -101,14 +94,21 @@ export const Status = () => {
...
-  if (error) {
-    return (
-      <Alert status="error">
-        <AlertIcon />
-        {error}
-      </Alert>
-    );
-  }
+  const initializePharmacy = async (p: EnrichedPharmacy) => {
+    await addRatingsAndHours(p);
```
captDaylight, line 98: shouldn't there be a returned value? or is this mutating the object? this would be a side effect we should stay away from

wouldn't the function parameter be  `Pharmacy` and the return valued `EnrichedPharmacy`? Looks like we're just renaming the type

apps/patient/src/utils/general.ts
```diff
@@ -188,3 +189,97 @@ export const countFillsAndRemoveDuplicates = (fills: types.Fill[]): FillWithCoun
...
+  const detailsRequest = {
+    placeId,
+    fields: [
+      'opening_hours',
+      'utc_offset_minutes', // this gives us isOpen()
+      'rating',
+      'business_status'
+    ]
+  };
+  let details: any;
```
captDaylight, line 246: @marisbest2 is gonna come after you

jknapp25: 🤦‍♂️

jknapp25: should have left this in draft lol

apps/patient/src/utils/general.ts
jknapp25: ah good catch

apps/patient/src/views/Status.tsx
jknapp25: smoothing out the flow, instant resolution is jumpy

apps/patient/src/views/Status.tsx
jknapp25: yeah good call, i'll fix that

PR #512, opened 2023-08-18
Neutron: Patient rerouting, remove instructions on rx review

jknapp25:
Add patient rerouting (#503)
Remove patient instructions on rx review (#511)


apps/patient/src/api/internal.ts
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Mutations
+ */
+
+export const markOrderAsPickedUp = async (orderId: string) => {
+  try {
+    const response: { markOrderAsPickedUp: boolean } = await graphQLClient.request(
+      MARK_ORDER_AS_PICKED_UP,
+      { markOrderAsPickedUpId: orderId }
```
captDaylight, line 72: `markOrderAsPickedUpId` is a lil confusing, maybe just `orderId`?

jknapp25: yeah I agree, not sure why we originally did that, updating will require updating the services backend so I'll leave for next time i revisit that mutation

apps/patient/src/views/Main.tsx
```diff
@@ -1,86 +1,70 @@
...
+const OrderContext = createContext(null);
+export const useOrderContext = () => useContext(OrderContext);
 
 export const Main = () => {
   const [searchParams] = useSearchParams();
   const orderId = searchParams.get('orderId');
   const token = searchParams.get('token');
 
+  if (token) {
+    setAuthHeader(token);
```
captDaylight, line 23: This is a side effect, should be wrapping this in an effect ... or would be ideally be handled via context / provider

apps/patient/src/views/Main.tsx
```diff
@@ -1,86 +1,70 @@
...
-  const [error, setError] = useState<string | undefined>(undefined);
 
   const navigate = useNavigate();
   const location = useLocation();
 
   if (location.pathname !== '/canceled') {
     if (!orderId || !token) {
       console.error('Missing orderId or token in search params');
-      navigate('/no-match');
+      navigate('/no-match', { replace: true });
```
captDaylight, line 34: side effect, sholud be  in a useEffect

jknapp25: fixed

apps/patient/src/views/Main.tsx
jknapp25: fixed

captDaylight (approved):

PR #521, opened 2023-08-23
Scroll to top on status

jknapp25:
- scroll to top on status view, helps when a low pharmacy in the list was selected (always thought this was a weird datadog thing)
- remove unnecessary reload for pharmacy details


captDaylight (approved):
hawt

PR #526, opened 2023-08-24
Show different warning on Order Creation if "Send to Patient"

captDaylight:
https://www.notion.so/photons/Warning-for-sending-order-doesn-t-make-sense-in-ROUTING-c6302e242160405f9a1b9b4779f07e59?pvs=4

### DIFF
--- a/apps/app/src/views/routes/NewOrder/components/OrderForm.tsx
+++ b/apps/app/src/views/routes/NewOrder/components/OrderForm.tsx
@@ -218,7 +218,10 @@ export const OrderForm = ({
 
         if (
           await confirmWrapper('Send Order?', {
-            description: 'This order will be immediately sent to the pharmacy of choice.',
+            // determine if order is "Send to Patient" by checking if pharmacyId is set
+            description: values.pharmacyId
+              ? 'This order will be sent immediately to the pharmacy of choice.'
+              : "This order will await the patient's choice of pharmacy.",
             cancelText: 'Keep Editing',
             confirmText: 'Yes, Send Order',
             darkMode: colorMode !== 'light'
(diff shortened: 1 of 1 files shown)


samkotlove (approved):

apps/app/src/views/routes/NewOrder/components/OrderForm.tsx
```diff
@@ -218,7 +218,10 @@ export const OrderForm = ({
 
         if (
           await confirmWrapper('Send Order?', {
-            description: 'This order will be immediately sent to the pharmacy of choice.',
+            // determine if order is "Send to Patient" by checking if pharmacyId is set
+            description: values.pharmacyId
+              ? 'This order will be sent immediately to the pharmacy of choice.'
+              : "This order will await the patient's choice of pharmacy.",
```
marisbest2, line 224: "This order will be sent when the patient chooses a pharmacy"

Maybe? 🤷

apps/app/src/views/routes/NewOrder/components/OrderForm.tsx
```diff
@@ -218,7 +218,10 @@ export const OrderForm = ({
 
         if (
           await confirmWrapper('Send Order?', {
-            description: 'This order will be immediately sent to the pharmacy of choice.',
+            // determine if order is "Send to Patient" by checking if pharmacyId is set
+            description: values.pharmacyId
```
marisbest2, line 222: Should we determine the message based `pharmacyId` vs however we know that its a "send to patient"? 

🤷

marisbest2 (approved):
made an alternate sugestion, but looks good to me

apps/app/src/views/routes/NewOrder/components/OrderForm.tsx
captDaylight: yeah that's better, adding!

apps/app/src/views/routes/NewOrder/components/OrderForm.tsx
captDaylight: That's currently how we determine if an order is being sent to the patient, the `createOrder` mutation is implicitly missing `pharmacyId`

PR #537, opened 2023-08-31
Optionally get a reason for canceling from provider

captDaylight:
ref https://www.notion.so/photons/Provider-able-to-add-reason-for-cancelation-85020b64ddbb4d8ca919ae12da88446c?pvs=4

- [x] Do we need to update the  reasons?

Screen Shot 2023-09-01 at 11 21 57 AM

### DIFF
--- a/apps/app/src/mutations/index.ts
+++ b/apps/app/src/mutations/index.ts
@@ -1,8 +1,8 @@
 import gql from 'graphql-tag';
 
 export const CANCEL_ORDER = gql`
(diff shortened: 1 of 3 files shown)


jknapp25:
Do you think it would help to provide descriptions alongside these reasons or are they pretty self-explanatory for providers?

captDaylight:
I think these are pretty self explanatory, but if we're not getting response we can revisit,good idea tho @jknapp25

PR #570, opened 2023-09-21
Add weight to prescription

captDaylight:
url be like `http://localhost:3000/prescriptions/new?weight=100.5`

Will result in this:

Screen Shot 2023-09-21 at 3 52 57 PM

Each new Rx draft will have reset to this note

https://www.notion.so/photons/Add-Weight-param-to-site-b2ae75c6e56e46fb94ba95bdb663542e?pvs=4


samkotlove (approved):
LGTM

PR #571, opened 2023-09-21
Stp demo follow-up

jknapp25:
- provider name -> patient name
- handle phone number via url param

### DIFF
--- a/apps/patient/src/views/Pharmacy.tsx
+++ b/apps/patient/src/views/Pharmacy.tsx
@@ -57,6 +57,7 @@ export const Pharmacy = () => {
   const token = searchParams.get('token');
   const isReroute = searchParams.get('reroute');
   const isDemo = searchParams.get('demo');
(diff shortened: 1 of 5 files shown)


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -299,13 +300,13 @@ export const Pharmacy = () => {
...
           triggerDemoNotification(
-            '5416029101',
+            phone,
             'photon:order:placed',
             selectedPharmacy.name,
             formatAddress(selectedPharmacy.address)
           );
 
-          navigate(`/status?demo=true`);
+          navigate(`/status?demo=true&phone=${phone}`);
```
samkotlove, line 309: why do we need phone number here?

jknapp25: to trigger other sms, was thinking okay for now later i'll move to context

captDaylight (approved):

PR #576, opened 2023-09-25
Patient card refactor

captDaylight:
Screen Shot 2023-09-25 at 12 16 55 PM

https://www.notion.so/photons/Refactor-Patient-Card-In-Elements-5e60487183eb429ba2df4db5f20f654a?pvs=4

based on these wireframes, see all states below

https://www.notion.so/photons/Wireframe-updated-patient-card-aee505ca974549e181e4bc82ac225d93?pvs=4

### DIFF
--- a/packages/elements/src/photon-multirx-form/components/PatientCard.tsx
+++ b/packages/elements/src/photon-multirx-form/components/PatientCard.tsx
@@ -1,6 +1,6 @@
 import { message } from '../../validators';
 import { string, any, record } from 'superstruct';
-import { createSignal, onMount, Show, createEffect } from 'solid-js';
+import { createSignal, onMount, Show, createEffect, createMemo } from 'solid-js';
 import { PatientInfo } from '@photonhealth/components';
 import { PatientStore } from '../../stores/patient';
 import { PhotonClientStore } from '../../store';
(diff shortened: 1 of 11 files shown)


packages/elements/src/photon-multirx-form/components/PatientCard.tsx
```diff
@@ -64,89 +65,44 @@ export const PatientCard = (props: {
...
+      </Show>
+      <Show when={patientId() && !props.hideAddress}>
+        <photon-card>
+          <PatientInfo
+            patientId={patientId()}
+            weight={props?.weight}
+            editPatient={props?.enableOrder ? () => setDialogOpen(true) : undefined}
+          />
           <photon-patient-dialog
+            hide-create-prescription={true}
```
jknapp25, line 97: just checking this should always be true

jknapp25 (approved):

packages/elements/src/photon-multirx-form/components/PatientCard.tsx
captDaylight: Just in this situation where we're setting the flag we'll be falling back to not showing the "create prescription" button (because we're already creating an Rx). 

In other scenarios, like creating a patient, we'll want the option to create a Rx immediately after saving.

PR #579, opened 2023-09-27
Settings page v2

marisbest2:
Settings page v2


apps/app/src/views/routes/SettingsV2/apollo.tsx
```diff
@@ -0,0 +1,79 @@
+import {
```
captDaylight, line 1: We have something similar in the SDK, would love to improve that tbh so we have one place we can point to

captDaylight:
not to futz with your process, but I see you're creating a `/SettingsV2` folder, which makes it hard to see what's being added or modified. why not just work directly in the folder?

.graphqlrc.yml
```diff
@@ -0,0 +1,9 @@
+projects:
+  # 'app-core':
+  #   schema: ['https://api.boson.health/graphql']
+  #   documents: ['apps/app/**/*.{graphql,js,ts,jsx,tsx}']
+
+  'app-v2':
+    schema: ['http://clinical-api.tau.health:8080/graphql']
+    documents: ['apps/app/**/SettingsV2/**/*.{ts,tsx}']
+    extensions: { languageService: { cacheSchemaFileForLookup: true } }
```
marisbest2, line 9: Gotta figure out a sane approach to this so that we _can_ point locally, or _can_ point remotely.

apps/app/src/views/routes/SettingsV2/apollo.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { setContext } from '@apollo/client/link/context/index.js';
+import { PhotonClient } from '@photonhealth/sdk';
+import { createContext, useContext } from 'react';
+
+let _apollo: ApolloClient<NormalizedCacheObject> | undefined;
+let _photonClient: PhotonClient | undefined;
+
+const clinicalApiUrl = process.env.REACT_APP_CLINICAL_GRAPHQL_URI as string;
+
+const apollo = (photonClient: PhotonClient) => {
```
marisbest2, line 17: Note: I specifically want two separate ApolloClients here (at least for now) so that we don't need to deal with conflicting type definitions

captDaylight: I think I'm rehashing things we talked about last Friday, but gonna recap some thoughts:

An alternate solution would be to have the SDK have two `HttpLink`s, one for the existing URI and the other for the new one. Then, you could use `ApolloLink.split` to decide which link to use based on some context like:

`useQuery(QUERY, { variables, context: { clientName: 'newClinical' } }) // bad naming`

But, in this solution you're saying that we would have conflicting type definitions because the Apollo client would have collisions on the type names because both endpoints are sharing the same cache. That makes sense ✅

So at least for now, we'll agree this is a necessary evil until we're fully migrated over. That means we'll have two instances of the Apollo client and recognize the hit we're taking (how big? 🤷‍♂️). Probably not something we'd want on the Patient App (tho not even using apollo there)... but fine for now on the clinical app (tho bearing in mind some orgs' clinicians are using their phones).

apps/app/src/App.tsx
```diff
@@ -53,38 +57,51 @@ export const App = () => {
...
+                <Route path=":prescriptionId" element={<Prescription />} />
+              </Route>
+              <Route path="/orders">
+                <Route path="/orders" element={<Orders />} />
+                <Route path="new" element={<NewOrder />} />
+                <Route path=":orderId" element={<Order />} />
+              </Route>
+              <Route path="/support" element={<Support />} />
+              <Route path="/playground" element={<Playground />} />
+              {useSettingsV2 ? (
```
captDaylight, line 83: 🏆

apps/app/src/views/routes/SettingsV2/components/invites/InviteList.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { graphql } from 'apps/app/src/gql';
+import { Outlet } from 'react-router-dom';
+import { useClinicalApiClient } from '../../apollo';
+import { InviteItem } from './InviteItem';
+
+const invitesQuery = graphql(/* GraphQL */ `
+  query InvitesQuery {
+    invites {
+      ...InviteFragment
+      id
```
captDaylight, line 28: haven't run this, but does id need to be here if also in `InviteFragment`?

captDaylight:
Looks like the backend change is merged in, this is good to play around with it locally?  

Some things are incomplete like `onSubmit` in `Organization.tsx`, is the thinking to get this in and incrementally move forward as this is only for Boson?

.graphqlrc.yml
marisbest2: Yeah. The issue is that this config is used by the VSCode syntax highlighter...

samkotlove:
Screenshot 2023-11-15 at 7 00 58 PM

apps/app/src/views/routes/Settings/components/Profile.tsx
```diff
@@ -10,45 +10,75 @@ import {
...
-    {
-      title: 'Organization ID',
-      value: organization?.id
+  const address = useMemo(() => {
+    const addressData = data?.me.address;
+    if (!addressData) {
+      return undefined;
     }
-  ];
+    return `${addressData.street1} ${addressData.street2 ? `${addressData.street2} ` : ''}${
```
jknapp25, line 66: Do you mind re-using `formatAddress`?

jknapp25:
image

marisbest2:
> image

@jknapp25 is this on mobile? I'm not worried about that for now.

jknapp25:
image

jknapp25:
@marisbest2 yeah mobile, just highlighting what i notice, not sure what your focus is rn

jknapp25:
How about red for destructive action?
image

jknapp25:
image

jknapp25:
not a part of your additions, but I've always felt like these headers were a bit superfluous, wdyt?

image

jknapp25:
Maybe hide pagination before loaded?
image

apps/app/src/views/routes/Settings/index.tsx
```diff
@@ -1,117 +1,184 @@
...
-        break;
-      case '/settings/catalog':
-        setTabIndex(2);
-        break;
-      default:
-        navigate('/settings/user');
-        break;
+    const idx = tabs.findIndex((path) => path === pathname);
+    if (
+      idx >= 0 &&
```
jknapp25, line 117: Mind being a little more verbose? I'm having trouble following.

.graphqlrc.yml
```diff
@@ -0,0 +1,13 @@
+projects:
+  # 'app-core':
```
captDaylight, line 2: still need these commented out things?

apps/app/src/gql/fragment-masking.ts
```diff
@@ -0,0 +1,66 @@
+import { ResultOf, DocumentTypeDecoration, TypedDocumentNode } from '@graphql-typed-document-node/core';
```
captDaylight, line 1: wdyt about not committing this folder?

https://github.com/dotansimha/graphql-code-generator/discussions/4253

apps/app/src/views/routes/Settings/components/invites/InviteItem.tsx
```diff
@@ -0,0 +1,29 @@
+import { Badge, Td, Text, Tr } from '@chakra-ui/react';
+import { FragmentType, graphql, useFragment } from 'apps/app/src/gql';
+
+export const inviteFragment = graphql(/* GraphQL */ `
```
captDaylight, line 4: I like that you're doing it this way to get the invite prop type 🙏, good example for the rest of the codebase

apps/app/src/views/routes/Settings/components/organization/Organization.tsx
```diff
@@ -0,0 +1,238 @@
...
+    </Button>
+  </>
+);
+
+export const Organization = () => {
+  const [isEditing, setIsEditing] = useState(false);
+  const client = useClinicalApiClient();
+  const { data, error } = useQuery(organizationQuery, { client });
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [updateOrganization, { loading: mutationLoading, error: mutationError }] = useMutation(
```
captDaylight, line 92: can you remove eslint if you remove `mutationError`?

apps/app/src/views/routes/Settings/components/organization/OrganizationEditForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { FC } from 'react';
+import * as yup from 'yup';
+import { FormikStateSelect, yupStateSchema } from '../utils/States';
+
+export const organizationFormSchema = yup.object({
+  id: yup.string().required(),
+  name: yup.string().required().min(1, 'Missing organization name'),
+  fax: yup
+    .string()
+    .matches(/^\s*(?:\+?(\d{1,3}))?[-. (]*(\d{3})[-. )]*(\d{3})[-. ]*(\d{4})(?: *x(\d+))?\s*$/, {
```
captDaylight, line 12: not urgent, smells like this is reusable regex for the codebase

apps/app/src/views/routes/Settings/components/templates/TemplateForm/components/CtaButtons.tsx
```diff
@@ -0,0 +1,27 @@
...
+  <HStack alignSelf="end">
+    <Button variant="outline" onClick={handleCancel} isLoading={loading && isSubmitting}>
+      Cancel
+    </Button>
+    <Button
+      type="submit"
+      colorScheme="brand"
+      form="template-builder"
+      onClick={() => setSubmitting(true)}
+      disabled={loading || catalogs.loading}
```
captDaylight, line 20: not urgent, but jomi wants to move away from disabled states and towards it triggering toast messages that inform you what's missing

apps/app/src/views/routes/Settings/components/templates/TemplateForm/components/RefillsInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Field, FieldProps } from 'formik';
+
+interface RefillsInputProps {
+  errors: any;
+  touched: any;
+  hidden: boolean;
+  setFieldValue: any;
+}
+
+export const RefillsInput = ({ errors, touched, hidden, setFieldValue }: RefillsInputProps) => {
```
captDaylight, line 21: wish this could be used in elements too 😭

apps/app/src/views/routes/Settings/components/utils/Roles.tsx
```diff
@@ -0,0 +1,75 @@
...
+export interface RolesSelectProps {
+  onChange: (role: readonly Role[]) => void;
+  onBlur: () => void;
+  value: Role[];
+}
+export const RolesSelect = (props: RolesSelectProps) => {
+  const client = useClinicalApiClient();
+  const [loadRoleOptions] = useLazyQuery(allRolesQuery, {
+    client,
+    fetchPolicy: 'cache-first'
```
captDaylight, line 49: reason for setting `fetchPolicy` here? is it not cache-first already?

apps/app/src/views/routes/Settings/index.tsx
```diff
@@ -1,117 +1,187 @@
...
-  const organization = getOrganization();
   const [tabIndex, setTabIndex] = useState(0);
-
   const { pathname } = useLocation();
   const navigate = useNavigate();
 
+  const client = useClinicalApiClient();
+  const { data, loading } = useQuery(settingsPageQuery, {
+    client,
+    fetchPolicy: 'cache-first'
```
captDaylight, line 105: same

captDaylight (approved):
Looks great, works well

apps/app/src/gql/fragment-masking.ts
marisbest2: I went back and forth. It'd require adding a build step and causes some issues when people check the code out. But yeah we should get rid of it soon-ish

apps/app/src/views/routes/Settings/components/templates/TemplateForm/components/CtaButtons.tsx
marisbest2: We'll deal with this later

.graphqlrc.yml
marisbest2: Yeah, just shows people how to quickly swap to local dev. I couldn't figure out how to do it with a command

PR #580, opened 2023-09-27
Display and add to patient med history

captDaylight:
For:
- https://www.notion.so/photons/Add-Medical-History-panel-to-prescribe-view-e0a3dbb669734c5ba794055943825df7?pvs=4
- https://www.notion.so/photons/Add-ability-to-add-externally-reported-medical-history-for-a-patient-522000f490ba4a01b2512da4c7407262?pvs=4

✨ _video demo_
https://www.loom.com/share/b2f3c546b60b432798bfb3e6c22c8a5d

### DIFF
--- a/packages/components/src/systems/PatientMedHistory/index.tsx
+++ b/packages/components/src/systems/PatientMedHistory/index.tsx
@@ -1,0 +1,97 @@
+import { createSignal, createEffect, Show, For } from 'solid-js';
+import gql from 'graphql-tag';
+import { usePhotonClient } from '../SDKProvider';
+import { PatientMedication } from '@photonhealth/sdk/dist/types';
+import Table from '../../particles/Table';
+import Text from '../../particles/Text';
+import generateString from '../../utils/generateString';
+
+const GET_PATIENT = gql`
+  query GetPatient($id: ID!) {
+    patient(id: $id) {
+      id
+      medicationHistory {
+        active
+        prescription {
+          id
+          writtenAt
+        }
+        medication {
+          id
+          name
+        }
+      }
+    }
+  }
+`;
+
+type PatientMedHistoryProps = {
+  patientId: string;
+};
+
+const LoadingRowFallback = () => (
+  <Table.Row>
+    <Table.Cell>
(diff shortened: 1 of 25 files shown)


jknapp25 (approved):
suuuuper cool, love the org setting too, great work 👏🏼

packages/components/src/systems/PatientMedHistory/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+      query: GET_PATIENT_MED_HISTORY,
+      variables: { id: props.patientId }
+    });
+    if (data?.patient?.medicationHistory) {
+      setMedHistory(data.patient.medicationHistory);
+    }
+  };
+
+  const addMedHistory = async (medicationId: string) => {
+    await (client as PhotonClient)!.apollo.mutate({
```
marisbest2, line 80: Why does this need the type assertion?

packages/components/src/utils/formatDate.ts
```diff
@@ -1,22 +1,13 @@
...
+  const year = date.getFullYear();
 
-  // Check if month and day are within range
-  if (month < 1 || month > 12 || day < 1 || day > 31) {
-    return '';
-  }
-
-  // use template literals to format the date string
-  return `${year}-${month.toString().padStart(2, '0')}-${day.toString().padStart(2, '0')}`;
+  const formattedDate = `${month}/${day}/${year}`;
```
marisbest2, line 11: Why'd we change our default here?

marisbest2 (approved):
Two qs. but LGTM

packages/components/src/utils/formatDate.ts
captDaylight: this wasn't being used anywhere inside of the `components` so I reimplemented it.

packages/components/src/systems/PatientMedHistory/index.tsx
captDaylight: this one is a true murder mystery, IDE was acting up before, but now I just removed and it's fine :|

PR #592, opened 2023-10-09
Prevent duplicates in table view

captDaylight:
ref https://www.notion.so/photons/After-scrolling-sometimes-last-table-result-is-duplicated-eb67ed350412406aa4f1ef0ad5284eeb?pvs=4

removed duplicate infinite scroll logic


apps/app/src/views/routes/Orders.tsx
```diff
@@ -224,6 +225,7 @@ export const Orders = () => {
 
   useEffect(() => {
     if (!loading && orders) {
+      console.log(orders);
```
samkotlove, line 228:
```suggestion
```

marisbest2 (approved):

PR #604, opened 2023-10-11
Better Auth Error messages Elements

captDaylight:
don't love the nested fallback `Show` ... but Solid wants to have a single return unlike React


jknapp25 (approved):

packages/elements/src/photon-authorized/index.tsx
```diff
@@ -46,30 +51,27 @@ export const PhotonAuthorized = (p: { children: JSXElement; permissions?: Permis
     );
   });
 
+  const canAccess = createMemo(() => authenticated() && inOrg() && hasPermission());
```
samkotlove, line 54: We want the `hasPermission` to be a different error message, something like "You are not authorized to prescribe"

PR #605, opened 2023-10-12
Capture Map Issues

captDaylight:
One of our clinicians occasionally has trouble with google maps not loading. I can't recreate it so this does two things:

1. there's a chance that google maps api hasn't loaded yet, so this tries a few times to grab it, not just on component load
2. better sentry handling, I haven't seen any errors yet so that tells me it's not getting caught atm

### DIFF
--- a/apps/app/src/views/components/LocationSearch.tsx
+++ b/apps/app/src/views/components/LocationSearch.tsx
@@ -61,9 +61,36 @@ export const LocationSearch = ({ isOpen, onClose }: LocationSearchProps) => {
     });
   };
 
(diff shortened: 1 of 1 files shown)


jknapp25:
Shouldn't we have seen this on the patient side as well? What if we move it outside of the component so it's loaded when the file is initially executed by the browser?

captDaylight:
@jknapp25 we've only heard this from one provider, I've not been able to recreate. this is more to see if I can snare this bug in the wild 🐛 🥅

jknapp25 (approved):

PR #637, opened 2023-10-31
add new permissions

marisbest2:
Update permissions list


captDaylight (approved):
someday these will be not managed in two places

PR #642, opened 2023-11-07
Capsule test setup

jknapp25:
Re-enable featured courier delivery option for patients located in Austin, TX

image


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -422,10 +422,13 @@ export const Pharmacy = () => {
     }
   }, [location]);
 
-  const patientAddressInAustinTX =
+  const deliveryAddressInAustinTX =
     order?.address?.city === 'Austin' && order?.address?.state === 'TX';
```
samkotlove, line 426: Can we check the list of zipcodes we have?

jknapp25: ah, yep added

apps/patient/src/data/capsuleAustinZipcodes.ts
```diff
@@ -0,0 +1,58 @@
+export const capsuleAustinZipCodes: string[] = [
```
samkotlove, line 1: nit: can we make this a map? Then we can do .has whcih is faster

captDaylight (approved):

PR #663, opened 2023-11-14
Readd templates on prescription creation

samkotlove:


jknapp25:
One catch, should disable "Add to templates" checkbox after adding each one

jknapp25 (approved):

PR #666, opened 2023-11-16
Patient UI polishing

jknapp25:
- bottom padding on pizza tracker
- remove logo flicker

### DIFF
--- a/apps/patient/src/components/Nav.tsx
+++ b/apps/patient/src/components/Nav.tsx
@@ -1,4 +1,3 @@
-import { useState, useEffect } from 'react';
 import {
   Alert,
   Box,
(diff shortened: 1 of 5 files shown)


marisbest2 (approved):
This looks fine.

But out of curiosity, can't we just use a standard `` tag with a src and set to null (or fixed height/width div) while it's loading? Or Chakra's `<Image>` with a fallback? 

Alternatively, instead of doing this in the react context, we can make a `Logo` component that just does all of it.

jknapp25:
@marisbest2 we could! But I'm trying to prevent the header showing up blank. I'd rather it shows up with the logo on first look, feels like a nicer experience to me.

jknapp25:
I probably will create a new Logo component tho, my current is specifically svg.

PR #667, opened 2023-11-16
Hide address form on prescription only form

samkotlove:
When we released the combined order/presccription view, we added a requirement to set address. That was carried over to the single prescription workflow, where an address _should not_ be required to prescribe


jknapp25 (approved):
Nice catch. There are two settings: `sendOrder` and `enableRxAndOrder`, just want to make sure these aren't being mixed up.

PR #670, opened 2023-11-20
Fix set "picked up" not showing in ready

jknapp25:
Option to set picked up was not surfacing in READY fulfillment state. I think this was just an oversight. At first we based off of order state, then fulfillment Received, should have been received/ready.

### DIFF
--- a/packages/sdk/src/types.ts
+++ b/packages/sdk/src/types.ts
@@ -131,6 +131,18 @@ export enum FulfillmentType {
   PickUp = 'PICK_UP'
 }
 
+export enum FulfillmentState {
+  Sent = 'SENT',
+  // Pick Up
+  Received = 'RECEIVED',
+  Ready = 'READY',
+  PickedUp = 'PICKED_UP',
+  // Mail Order
+  Filling = 'FILLING',
+  Shipped = 'SHIPPED',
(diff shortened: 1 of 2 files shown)


packages/sdk/src/types.ts
```diff
@@ -416,7 +428,7 @@ export type OrderFilter = {
 export type OrderFulfillment = {
   __typename?: 'OrderFulfillment';
   carrier?: Maybe<Scalars['String']>;
-  state: Scalars['String'];
+  state: FulfillmentState;
```
samkotlove, line 431: Aren't these autogenerated?

jknapp25: I believe the types in `dist` get generated from src

marisbest2: I dont think we have type generation yet

marisbest2:
FWIW I much prefer Typescript string literals to enums. Enums are much harder to read than strings, require being imported, are javascript objects under the hood, among other reasons

packages/sdk/src/types.ts
captDaylight: correct, we don't, yet

marisbest2 (approved):

PR #689, opened 2023-11-29
remove superfluous buttons and add to sidebar when small

marisbest2:


jknapp25 (approved):
can we just remove commented out comps?

PR #694, opened 2023-11-30
proof of concept for clinical API in elements

marisbest2:
Proof of concept for including the clinical API in elements. We should use this as the basis for implementing Signature Attestation once we're ready for it


captDaylight:
@marisbest2 removing now that we have this working in elements with https://github.com/Photon-Health/client/pull/798

PR #699, opened 2023-11-30
allow sorting the user table

marisbest2:
All done on the FE for now. 

Filtering is too complicated an affordance


captDaylight:
great implementation!

One thing, when I sort by role I'd expect the last one in descending order would be the longest. Code looks correct, I don't see it why this would happen.

Screen Shot 2023-12-01 at 10 01 17 AM

captDaylight (approved):

PR #716, opened 2023-12-04
add resend and delete invite

photonsFardad:


marisbest2 (commented):
Code generally looks good. We should get Jomi to sign off.

thebadprince:
@photonsFardad @marisbest2  LGTM, but do we need a confirmation for each? (i.e. a modal that opens up `Delete Invite?` / `Resend Invite?` Yes / No). Just curious. But that can also be a separate PR if we need. 

@jknapp25 also added a new toast component on the Prescribe that we can use as UI feedback (i.e. a notification that says `Invite deleted` / `Invite resent`). I hit the resent invite button and I wasn't sure if it actually resent it or not.

marisbest2 (approved):

PR #717, opened 2023-12-04
Fix disabled create order still allowing creation

jknapp25:
Clicking the disabled create order icon button on prescriptions list will still take you to the create order page.

image

Added toast on click:

image


captDaylight:
@jknapp25 great catch, can you trigger a toast in that condition? Jomi wants to move away from silent disabled buttons

jknapp25:
> @jknapp25 great catch, can you trigger a toast in that condition? Jomi wants to move away from silent disabled buttons

Added toast

image

captDaylight (approved):

captDaylight:
maybe blue and the subtle variatn? seems intense when it's solid red lol

PR #719, opened 2023-12-05
Fix get directions intermittently failing

jknapp25:
Noticed errors on pizza tracker "get directions" in Datadog. Looks like `focus()` is not needed here. Removing has the same behavior of opening in new tab but appears more reliable.

MDN on focus(): 

> It may fail due to user settings and the window isn't guaranteed to be frontmost before this method returns.


captDaylight (approved):
Nice find

PR #730, opened 2023-12-07
Create Patient without setting Pharm

captDaylight:
In the elements patient form, when creating a patient or editing a patient without a preferred pharmacy it would default to the first in the list according to their address. Now it doesn't default to that.


jknapp25:
location seems to be hanging in loading

image

jknapp25:
wanna split it up while we're in ther'?

Jomi's latest:

image

packages/elements/src/photon-multirx-form/index.tsx
```diff
@@ -190,6 +192,17 @@ customElement(
     };
 
     const submitForm = async (enableOrder: boolean) => {
+      if (
```
jknapp25, line 195: 👍🏼

samkotlove (approved):

PR #736, opened 2023-12-08
Fix Decimal input on Mobile for "Quantity" input 

captDaylight:
This is a hack to fix a bug on mobile where the form is cleared when the user types a decimal in the dispense quantity input. But because of the hack, we have to initialize the value to register the validator that would otherwise be registered with the input.


marisbest2 (approved):
Wow thats ugly. 🤷

PR #741, opened 2023-12-09
Pharmacy selection filters

jknapp25:
Refetch pharmacies with filter criteria when filter enabled. Also some copy updates, pluralization, cleanup.

https://github.com/Photon-Health/client/assets/3934326/170b30c1-d08b-4f2f-a231-dddfa6752dad

### DIFF
--- a/apps/patient/src/api/external.ts
+++ b/apps/patient/src/api/external.ts
@@ -57,7 +57,7 @@ export const getPlace = async (
     }
     return null;
   } catch (e) {
-    console.error(e);
+    console.log(e);
     return null;
   }
 };
--- a/apps/patient/src/components/PharmacyCard.tsx
+++ b/apps/patient/src/components/PharmacyCard.tsx
@@ -93,7 +93,7 @@ interface PharmacyCardProps {
   onSetPreferred?: () => void;
   onChangePharmacy?: () => void;
   onGetDirections?: () => void;
-  isPharmacySelection?: boolean;
+  selectable?: boolean;
 }
 
 export const PharmacyCard = memo(function PharmacyCard({
@@ -108,7 +108,7 @@ export const PharmacyCard = memo(function PharmacyCard({
   onChangePharmacy,
   onSetPreferred,
   onGetDirections,
-  isPharmacySelection = false
+  selectable = false
 }: PharmacyCardProps) {
   if (!pharmacy) return null;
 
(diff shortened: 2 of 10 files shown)


photonsFardad (approved):

apps/patient/src/components/PharmacyCard.tsx
```diff
@@ -119,6 +121,7 @@ export const PharmacyCard = memo(function PharmacyCard({
       borderColor={selected && onSelect ? 'brand.600' : 'white'}
       onClick={() => onSelect && onSelect()}
       mx={isMobile ? -3 : undefined}
+      cursor={isPharmacySelection ? 'pointer' : undefined}
```
samkotlove, line 124: what's this doing?

samkotlove (dismissed):
Not quite sure about the logic. You're conditionally showing filters if there are open and 24hr pharmacies loaded. Why not just search for those pharmacies directly and always show the filters? The backend should support that

apps/patient/src/components/PharmacyCard.tsx
jknapp25: better name would have been `selectable`, updated, on status page pharmacy card can't be selected so not using pointer

apps/patient/src/components/PharmacyFilters.tsx
```diff
@@ -0,0 +1,42 @@
...
+}: PharmacyFiltersProps) => {
+  return (
+    <HStack>
+      <Text>Filter by</Text>
+      <Button
+        size="sm"
+        variant="filter"
+        isActive={enableOpenNow}
+        onClick={() => {
+          if (enable24Hr) setEnable24Hr(false);
```
jknapp25, line 23: Filters are mutually exclusive since openNow and open24Hr is kind of weird

marisbest2: I think it's fine to allow clicking both and I think it'd be a bit weird if clicking "openNow" turned off `24hr` filter

marisbest2: Either way, you can do `setEnable24Hr(false)` without having the `if` statement

marisbest2 (approved):
Next year, we should do something about caching ratings results

apps/patient/src/api/external.ts
```diff
@@ -57,7 +57,7 @@ export const getPlace = async (
     }
     return null;
   } catch (e) {
-    console.error(e);
+    console.log(e);
```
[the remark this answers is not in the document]
jknapp25: It's just an error saying 0 results found so not that helpful to see it in datadog

jknapp25:
@marisbest2 feedback from jomi is to go with mutually exclusive filters since open24hr includes open now

PR #743, opened 2023-12-12
personal templates

marisbest2:


.graphqlrc.yml
```diff
@@ -1,7 +1,10 @@
 projects:
-  # 'app-core':
-  #   schema: ['https://api.boson.health/graphql']
-  #   documents: ['apps/app/**/*.{graphql,js,ts,jsx,tsx}']
+  'app-core':
+    schema:
+      'https://egkkxiywpvbtplabf5mo6syu3a.appsync-api.us-east-1.amazonaws.com/graphql':
+        headers: { 'x-api-key': 'da2-evtma6y5jfb7tnhxhxqojjsfvi' }
```
marisbest2, line 5: This is the boson API key and shouldn't be considered secret

packages/sdk/src/lib.ts
```diff
@@ -123,6 +129,12 @@ export class PhotonClient {
         query: {
           fetchPolicy: 'cache-first',
           errorPolicy: 'all'
+        },
+        mutate: {
```
marisbest2, line 133: Note to @captDaylight to look at and test this

apps/app/src/views/routes/Settings/components/templates/TemplateActions.tsx
```diff
@@ -1,24 +1,31 @@
...
   MenuList,
-  useColorMode,
-  useBreakpointValue
+  useColorMode
 } from '@chakra-ui/react';
 import { FiEdit, FiEye, FiMoreVertical, FiTrash } from 'react-icons/fi';
 
 import { usePhoton } from '@photonhealth/react';
 
+import { PrescriptionTemplate } from 'packages/sdk/dist/types';
```
captDaylight, line 14: thoughts in general on `import type { ...yadayada...`

apps/app/src/views/routes/Settings/components/templates/TemplateForm/index.tsx
```diff
@@ -1,155 +1,236 @@
...
+import { types, usePhoton } from '@photonhealth/react';
+import { CATALOG_TREATMENTS_FIELDS } from 'apps/app/src/model/fragments';
+import { TEMPLATE_INITIAL_VALUES, TEMPLATE_SCHEMA, TemplateSchemaType } from './utils';
+
+type PrescriptionTemplate = types.PrescriptionTemplate;
+
+interface CreateRxTemplateVariables {
+  catalogId: string;
+  treatmentId: string;
+  dispenseAsWritten: boolean | undefined;
```
captDaylight, line 40: more curious, you like this way over the `?:` shorthard? your way is definitely more legible when you skim

apps/app/src/views/routes/Settings/components/templates/TemplateForm/utils.ts
```diff
@@ -25,23 +26,27 @@ export const TEMPLATE_SCHEMA = yup.object({
...
 });
 
-export const TEMPLATE_INITIAL_VALUES = {
+export type TemplateSchemaType = yup.InferType<typeof TEMPLATE_SCHEMA>;
+
+export const TEMPLATE_INITIAL_VALUES: TemplateSchemaType = {
   treatment: {
     id: '',
-    name: '',
+    // name: '',
```
captDaylight, line 39: ?

captDaylight:
clicking "create template" could use a little [feedback](https://chakra-ui.com/docs/components/button#button-loading-state), IRL I clicked it twice
Screen Shot 2023-12-13 at 12 42 28 PM

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -378,9 +426,31 @@ export const AddPrescriptionCard = (props: {
                 }
               />
             </Show>
+            <Show when={props.store.addToTemplates?.value ?? false}>
```
captDaylight, line 429: nice touch

captDaylight:
I'm guessing there's probably not a reason to edit if a template is private or org-wide, just saying since I see it's disabled.

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -81,27 +101,30 @@ export const AddPrescriptionCard = (props: {
...
-            notes: props.store.notes.value,
-            fillsAllowed: props.store.refillsInput.value + 1,
-            addToTemplates: props.store.addToTemplates?.value ?? false,
-            catalogId: props.store.catalogId.value ?? undefined
-          }
-        ]
+        value: [...(props.store.draftPrescriptions?.value || []), draft]
       });
+
+      const addToTemplate = props.store.addToTemplates?.value ?? false;
```
captDaylight, line 126: these should be `createMemo` if you're deriving from a reactive variable

packages/elements/src/photon-treatment-select/index.tsx
```diff
@@ -77,6 +80,16 @@ customElement(
       return [];
     };
 
+    const onOpen = async () => {
+      // if (store.catalogs.data.length === 0) {
```
captDaylight, line 84: remove commented condition

jknapp25:
think we could just show "3" here and >5 pages use ellipsis?

image

apps/app/src/views/components/TablePage.tsx
```diff
@@ -104,20 +106,23 @@ export const TablePage = (props: TablePageProps) => {
...
+          <Stack
+            direction={{ base: 'column', md: `row${ctaRight ? '-reverse' : ''}` }}
+            justify="space-between"
+          >
             {ctaRouteColorTextUndefined ? null : (
               <Button
                 as={ctaOnClick ? undefined : RouterLink}
                 to={ctaRoute || ''}
                 onClick={ctaOnClick}
-                leftIcon={<FiPlus fontSize="1.25rem" />}
```
jknapp25, line 120: double-checking that this is meant to affect prescription/order tables too

jknapp25:
Possible to use a little more of the horizontal space here? Lot's of text in this mobile table.

Screen Shot 2023-12-13 at 11 38 14 AM

packages/sdk/src/lib.ts
captDaylight: this worked perfectly on the create templates mutation, but not seeing it happen for deleting templates.

captDaylight (approved):
Don't see any real blockers, just a few comments and questions

jknapp25:
I think secondary "Keep Editing" is residual, I wasn't editing in this case

image

packages/sdk/src/lib.ts
marisbest2: I actually don't think it works in either place. We need to change the appsync cache

apps/app/src/views/routes/Settings/components/templates/TemplateActions.tsx
marisbest2: oh yeah, this is actually a bad import. We should export all types from the sdk at the top level. `Nx` doesnt like when you reach across module boundaries

marisbest2: I'll fix

apps/app/src/views/routes/Settings/components/templates/TemplateForm/index.tsx
marisbest2: they mean different things, and its actually valuable in a React context. 

```tsx
interface Props1 {
   a?: string
}

interface Props2 {
   a: string | undefined;
}

const Element1 = (props: Props1) => <div />;
const Element2 = (props: Props2) => <div />;

// no error because `a` is not required
const NoError = () => <Element1 />; 

// error: Property 'a' is missing in type '{}' but required in type 'Props2'.ts(2741)
const Error = () => <Element2 />; // 
```

Generally, especially for React elements that are part of a tree, I prefer writing it out because I'm less likely to miss a spot where its needed. In a library or where its a flag, its often better to use the `?:` version.

apps/app/src/views/routes/Settings/components/templates/TemplateForm/utils.ts
marisbest2: oh, yeah should fix that

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
marisbest2: I don't really know how solid works, but I'd think that I dont because its just part of the `handleAddPrescription`, not part of the render logic. I just need a handle on it before the form is cleared. 

Am I wrong?

apps/app/src/views/components/TablePage.tsx
marisbest2: 🤷 I kinda hate the `+` tbh

jknapp25: maybe double-check w/ jomi?

PR #751, opened 2023-12-12
Update masking on patient card

jknapp25:
Noticed that the Text web comp doesn't include any text elements other than span so that mask isn't getting applied to anything. This just puts it back up top where it applies to everything in the card.


captDaylight:
I'm confused, why can't it mask a span?

captDaylight (approved):

PR #763, opened 2023-12-14
bugfix templates pagination

marisbest2:


jknapp25:
on templates tab only i keep getting this

Screen Shot 2023-12-14 at 11 05 45 AM

jknapp25 (approved):
Nice, also noticed one other small thing, next is still enabled with nothing showing. Non-blocker tho.

image

PR #764, opened 2023-12-14
bump versions

marisbest2:


captDaylight (approved):
added some labels, I tihnk the "elements version change" helps me see what's included when I write up the version notes

PR #766, opened 2023-12-14
Add Amazon Pharmacy as option for Sana

jknapp25:


packages/settings/src/lib/neutron.ts
```diff
@@ -199,6 +199,7 @@ export const neutron: {
     ...defaultSettings,
     logo: 'sesame_logo.jpg',
     accentColor: '#5224C7',
+    enableCourierNavigate: false,
```
marisbest2, line 202: do we actually want to do this?

marisbest2 (approved):

packages/settings/src/lib/neutron.ts
jknapp25: yep, just to highligh costco for sesame

PR #768, opened 2023-12-14
0 paginate

marisbest2:
Good catch @jknapp25


jknapp25:
Nice! Could you add to the catalog table as well? Do you think this could be baked into the component?

jknapp25 (approved):

marisbest2:
ugh. I'll deal with that later

PR #778, opened 2023-12-21
Status page updates

jknapp25:
This includes the status page updates documented [here](https://www.figma.com/file/BkeblOMtaBc6BM5TDZ0VPJ/Patient-App?type=design&node-id=760-318&mode=design&t=yDgswGSO7WCY8iNv-0) as well as some cleanup that made that easier to do. Also some other quality improvements.

- copy updates (buncha new stuff, pluralize ui, dynamic header/subheader)
- tag copy: 24hr -> Open 24 Hours
- fix: it was defaulting to pickup states if patient selected (not going directly to status view)
- show fake rx header on boson and neutron
- hide courier/mail order pharmacy details on status page
- fix: canceled and no match ui weren't surfacing
- debounce location search


apps/patient/src/components/LocationModal.tsx
```diff
@@ -80,16 +89,14 @@ export const LocationModal = ({ isOpen, onClose }: any) => {
   };
 
   const loadOptions = (inputValue: string, callback: (options: any) => void) => {
```
captDaylight, line 91: This `loadOptions` function is just passing the variables to `debouncedSearchForLocations`, maybe get rid of it

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -340,17 +339,31 @@ export const Pharmacy = () => {
           setTimeout(() => {
             setShowFooter(false);
 
+            // Fudge it so that we can show the pharmacy card on initial load of the
+            // status view for all types. On my christmas list for 2024 is better
```
captDaylight, line 343: 🎅

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -446,20 +459,23 @@ export const Pharmacy = () => {
     }
   }, [enableOpenNow, enable24Hr]);
 
+  const flattenedFills = countFillsAndRemoveDuplicates(order.fills);
```
captDaylight, line 462: no need to do this in this PR, but maybe these could just be derived in the provider and passed with the useOrderContext. Seen this logic twice now

apps/patient/src/views/Status.tsx
```diff
@@ -180,16 +165,24 @@ export const Status = () => {
     }
   }, []);
 
-  // Only show "Text us now" prompt if pickup and RECEIVED or READY
-  const showChatAlertStates: types.FulfillmentState[] = ['RECEIVED', 'READY'];
-  const showChatAlert = showChatAlertStates.includes(fulfillment?.state);
+  const flattenedFills = countFillsAndRemoveDuplicates(fills);
```
captDaylight, line 168: :)

apps/patient/src/components/LocationModal.tsx
jknapp25: cleans up the jsx a little, but I guess it's a bit superfluous, I'll remove

apps/patient/src/views/Pharmacy.tsx
jknapp25: added to context

PR #787, opened 2023-12-29
Fix auth on browsers with ITP (i.e. safari)

samkotlove:
Was able to get auth to work on safari (and presumably other browser with ITP (Intelligent Tracking Prevention) enabled ). The key was using [Refresh Token Rotation](https://auth0.com/docs/secure/tokens/refresh-tokens/refresh-token-rotation) and setting storage to be done in localstorage instead of cookies.

I also upgraded the version of auth0 we were using but this didnt end up having an impact, but left it as it took some work to get upgraded and we were very behind

### DIFF
--- a/apps/app/src/views/routes/SelectOrg.tsx
+++ b/apps/app/src/views/routes/SelectOrg.tsx
@@ -76,6 +76,7 @@ export const SelectOrg = () => {
               onClick={() => {
                 setOrganization(id);
                 login({
+                  organizationId: id,
                   appState: {
                     returnTo: from
                   }
--- a/packages/sdk/src/auth.ts
+++ b/packages/sdk/src/auth.ts
@@ -142,8 +142,7 @@ export class AuthManager {
     const opts: GetTokenSilentlyOptions | GetTokenWithPopupOptions = {
       authorizationParams: {
         audience: audience || this.audience || undefined,
-        ...(this.organization ? { organization: this.organization } : {}),
-        scope: 'openid profile email offline_access'
+        ...(this.organization ? { organization: this.organization } : {})
       }
     };
 
--- a/packages/sdk/src/auth.ts
+++ b/packages/sdk/src/auth.ts
@@ -99,7 +99,6 @@ export class AuthManager {
       authorizationParams = Object.assign(opts, { appState });
     }
     opts.authorizationParams = authorizationParams;
-    console.log('login', opts, organizationId, this.organization);
 
     return this.authentication.loginWithRedirect(opts);
   }
--- a/packages/sdk/src/auth.ts
+++ b/packages/sdk/src/auth.ts
@@ -5,7 +5,8 @@ import {
   LogoutOptions as Auth0LogoutOptions,
   RedirectLoginOptions,
   RedirectLoginResult,
-  User
+  User,
+  AuthorizationParams
 } from '@auth0/auth0-spa-js';
 
 const CODE_RE = /[?&]code=[^&]+/;
@@ -83,20 +84,22 @@ export class AuthManager {
    * @returns
    */
   public async login({ organizationId, invitation, appState }: LoginOptions): Promise<void> {
-    let opts: RedirectLoginOptions<any> = { redirectMethod: 'assign' };
-
+    const opts: RedirectLoginOptions<any> = {};
+    let authorizationParams: AuthorizationParams = {};
     if (organizationId || this.organization) {
-      opts = Object.assign(opts, {
+      authorizationParams = Object.assign(opts, {
         organization: organizationId || this.organization
       });
     }
     if (invitation) {
-      opts = Object.assign(opts, { invitation });
+      authorizationParams = Object.assign(opts, { invitation });
     }
 
     if (appState) {
-      opts = Object.assign(opts, { appState });
+      authorizationParams = Object.assign(opts, { appState });
     }
+    opts.authorizationParams = authorizationParams;
+    console.log('login', opts, organizationId, this.organization);
 
     return this.authentication.loginWithRedirect(opts);
   }
@@ -108,8 +111,10 @@ export class AuthManager {
    */
   public async logout({ returnTo, federated = false }: LogoutOptions): Promise<void> {
     const opts: Auth0LogoutOptions = {
-      ...(returnTo ? { returnTo } : {}),
-      ...(federated ? { federated } : {})
+      logoutParams: {
+        ...(returnTo ? { returnTo } : {}),
+        ...(federated ? { federated } : {})
+      }
     };
 
     return this.authentication.logout(opts);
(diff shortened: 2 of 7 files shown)


apps/app/src/views/routes/SelectOrg.tsx
```diff
@@ -76,6 +76,7 @@ export const SelectOrg = () => {
               onClick={() => {
                 setOrganization(id);
                 login({
+                  organizationId: id,
```
samkotlove, line 79: this part is confusing to me, i'm not quite sure why this would be required. My best guess is that as a part of upgrading there was a change in behavior

packages/sdk/src/lib.ts
```diff
@@ -86,19 +86,26 @@ export class PhotonClient {
...
       this.audience = 'https://api.neutron.health';
       this.uri = 'https://api.neutron.health/graphql';
     }
+
+    const params: Auth0ClientOptions = {
+      domain: domain ? domain : developmentMode ? 'auth.neutron.health' : 'auth.photon.health',
+      clientId: clientId,
+      cacheLocation: 'localstorage',
+      useRefreshTokens: true,
+      useRefreshTokensFallback: true,
```
samkotlove, line 101: these three lines were the main thing that changed here

packages/sdk/src/auth.ts
```diff
@@ -83,20 +84,21 @@ export class AuthManager {
...
     if (invitation) {
-      opts = Object.assign(opts, { invitation });
+      authorizationParams = Object.assign(opts, { invitation });
     }
 
     if (appState) {
-      opts = Object.assign(opts, { appState });
+      authorizationParams = Object.assign(opts, { appState });
     }
+    opts.authorizationParams = authorizationParams;
```
samkotlove, line 101: purely a contract change in the upgrade

packages/sdk/src/auth.ts
```diff
@@ -108,8 +110,10 @@ export class AuthManager {
    */
   public async logout({ returnTo, federated = false }: LogoutOptions): Promise<void> {
     const opts: Auth0LogoutOptions = {
-      ...(returnTo ? { returnTo } : {}),
-      ...(federated ? { federated } : {})
+      logoutParams: {
+        ...(returnTo ? { returnTo } : {}),
+        ...(federated ? { federated } : {})
+      }
```
samkotlove, line 116: contract change with upgrade

packages/sdk/src/auth.ts
```diff
@@ -136,8 +140,11 @@ export class AuthManager {
     }
   ): Promise<string> {
     const opts: GetTokenSilentlyOptions | GetTokenWithPopupOptions = {
-      audience: audience || this.audience || undefined,
-      ...(this.organization ? { organization: this.organization } : {})
+      authorizationParams: {
+        audience: audience || this.audience || undefined,
+        ...(this.organization ? { organization: this.organization } : {}),
+        scope: 'openid profile email offline_access'
```
samkotlove, line 146: need to test if this `scope` part is actually necessary

packages/sdk/src/auth.ts
```diff
@@ -162,10 +170,12 @@ export class AuthManager {
     { audience }: { audience?: string } = {
       audience: this.audience
     }
-  ): Promise<string> {
+  ): Promise<string | undefined> {
```
samkotlove, line 173: their contract changed, not 100% that this wont have downstream consequences

packages/sdk/src/auth.ts
```diff
@@ -162,10 +170,12 @@ export class AuthManager {
...
     const opts: GetTokenWithPopupOptions = {
-      audience: audience || this.audience || undefined,
-      ...(this.organization ? { organization: this.organization } : {})
+      authorizationParams: {
+        audience: audience || this.audience || undefined,
+        ...(this.organization ? { organization: this.organization } : {})
+      }
```
samkotlove, line 178: contract change from sdk upgrade

packages/sdk/src/lib.ts
```diff
@@ -86,19 +86,26 @@ export class PhotonClient {
...
+      authorizationParams: {
+        redirect_uri: redirectURI,
+        audience: this.audience
```
samkotlove, line 104: not 100% if `audience` addition is necessary, need to confirm

samkotlove: just confirmed, is necessary

packages/sdk/src/auth.ts
samkotlove: Confirmed not necessary, removed

packages/sdk/src/auth.ts
```diff
@@ -136,8 +140,10 @@ export class AuthManager {
     }
   ): Promise<string> {
     const opts: GetTokenSilentlyOptions | GetTokenWithPopupOptions = {
-      audience: audience || this.audience || undefined,
-      ...(this.organization ? { organization: this.organization } : {})
+      authorizationParams: {
+        audience: audience || this.audience || undefined,
+        ...(this.organization ? { organization: this.organization } : {})
+      }
```
samkotlove, line 146: just a contract change

packages/sdk/src/auth.ts
```diff
@@ -150,6 +156,7 @@ export class AuthManager {
         throw e;
       }
     }
+    if (!token) throw new Error('Missing token');
```
samkotlove, line 159: typescript was yelling at me

apps/app/src/views/routes/SelectOrg.tsx
captDaylight: same thing on line 43?

captDaylight: also is this something we need in `/Users/paulchristophe/Documents/GitHub/client/packages/elements/src/photon-multirx-form/index.tsx` line 127?

captDaylight:
on chrome logging in via localhost: `https://auth.boson.health/authorize?clientId=7N9QZujlNJHL8EIPqXpu1wq8OuXqoxKb&useRefreshTokensFallback=true&authorizationParams=%5Bobject%20Object%5D&scope=openid%20profile%20email%20offline_access&response_type=code&response_mode=query&state=RlJiMUh4SkJDZF9tR1VkNW5xSFoyRnZzQU9iNUhualFEMy1XaWFDT1RmYw%3D%3D&nonce=Z0dNcUVwSDBxeG9wZk8uSXJKRy56T2hfMGFlM0FqenNjOFZkeV9TaEZDeA%3D%3D&code_challenge=9SlR-2Si9Wh-dq5SMyZvPav0L67oIsQ4RfSlAeaOQ1k&code_challenge_method=S256&auth0Client=eyJuYW1lIjoiYXV0aDAtc3BhLWpzIiwidmVyc2lvbiI6IjEuMjIuNiJ9`
Screen Shot 2024-01-02 at 9 55 40 AM

captDaylight (approved):

marisbest2:
- Does this break when we don't have access to localstorage? Do we care?
- We should test that this will work on non- boson/photon.health domains

Otherwise LGTM

marisbest2 (approved):

packages/sdk/src/auth.ts
marisbest2: Yeah, `getTokenWithPopup` could be `undefined` so token is `string | undefined` if like 154 is the last executing line. You could probably push this up to there.

packages/sdk/src/auth.ts
marisbest2: Specifically, if I'm reading the code right, this method
- Triggers a login
- Saves the result of that login to the cache
- Reads the access_token out of the cache

So if its undefined, it presumably means the token wasn't able to be read from the cache, probs indicating that the cache isnt readable

PR #796, opened 2024-01-04
signature attestation FE; still needs to be wired up to the API

marisbest2:


packages/components/src/systems/SignatureAttestation/SignatureAttestationModal.tsx
```diff
@@ -0,0 +1,252 @@
...
+        }
+        ... on NotApplicableSignatureAttestation {
+          reason
+        }
+      }
+    }
+  }
+`;
+
+export type GetCurrentUserSignatureAttestationStatusType = {
```
captDaylight, line 31: any way we can generate these the same way we're doing for app?

packages/components/src/systems/SignatureAttestation/SignatureAttestationModal.tsx
```diff
@@ -0,0 +1,252 @@
...
+    ) {
+      setStatus({ status: 'COMPLETE' });
+    } else
+      setStatus({
+        status: 'NEEDS ATTESTATION',
+        version: req.data.me.signatureAttestationStatus.version
+      });
+  };
+
+  const dispatchSignatureAttestationAgreed = () => {
```
captDaylight, line 149: you don't need to do this unless you're trying to communicate with something in the react app, otherwise the prescribe form internally is just solid so pass a callback function

packages/components/src/systems/SignatureAttestation/SignatureAttestationModal.tsx
```diff
@@ -0,0 +1,252 @@
...
+  };
+
+  createEffect(() => {
+    if (status().status === 'COMPLETE') {
+      props.onAgreeComplete?.();
+      dispatchSignatureAttestationAgreed();
+    }
+  });
+
+  createEffect(() => {
```
captDaylight, line 174: is this happening on mount? if so use `onMount`

packages/components/src/systems/SignatureAttestation/SignatureAttestationModal.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+type Status =
+  | {
+      status: 'LOADING';
+    }
+  | { status: 'NEEDS ATTESTATION'; version: string }
+  | { status: 'COMPLETE' }
+  | { status: 'ERROR'; errors: any[] };
+
+export const SignatureAttestationModal = (props: SignatureAttestationModalProps) => {
```
captDaylight, line 118: nit on name, call it `SignatureAttestationCard`

packages/components/src/systems/SignatureAttestation/index.ts
```diff
@@ -0,0 +1 @@
+export { SignatureAttestationModal } from './SignatureAttestationModal';
```
captDaylight, line 1: I should do this everywhere...

packages/components/src/systems/SignatureAttestation/SignatureAttestationModal.tsx
marisbest2: Not easily / quickly. I have a TODO item to go back and do these though.

packages/components/src/systems/SignatureAttestation/SignatureAttestationModal.tsx
captDaylight: or I guess maybe you're trying to avoid drilling the callback through the prescribe wrapper to the prescribe element, all good in that case

marisbest2: yeah exactly.

captDaylight (approved):

packages/components/src/systems/SignatureAttestation/SignatureAttestationModal.tsx
marisbest2: It _should_ happen any time the state of the modal / component changes. I dont understand the lifecycle enough. I think thats more than just onMount. Especially because I dont think you're guaranteed to have the `client` on mount?

marisbest2:
WHY IS PRETTIER FAILING!

packages/components/src/systems/SignatureAttestation/SignatureAttestationModal.tsx
marisbest2: Its a modal (takes over the whole area) not just a card

PR #798, opened 2024-01-05
Clinical API in elements

captDaylight:
- adds a public `apolloClinical` to the SDK
- removes the apollo client in the settings page
- updates all Settings queries and mutations to point to SDK apollo instance
- small changes to importing react that was bugging me


packages/sdk/src/lib.ts
```diff
@@ -95,9 +111,19 @@ export class PhotonClient {
     this.audience = audience;
     this.uri = uri;
     this.clinicalUrl = getClinicalUrl(uri);
+
+    this.clinicalApiUri = `${
+      this.clinicalUrl?.includes('photon')
```
captDaylight, line 116: this I'm looking for feedback on... on the one hand we could pass the clinical uri the way we're passing lambdas... or we could just derive it the way @marisbest2. ain't pretty but it does the job, wdyt?

apps/app/src/App.tsx
```diff
@@ -3,7 +3,7 @@ import { useColorMode } from '@chakra-ui/react';
 import { BrowserRouter, Route, Routes } from 'react-router-dom';
 
 import { AppState } from '@auth0/auth0-react';
-import { PhotonClient, PhotonProvider } from '@photonhealth/react';
+import { PhotonClient, PhotonProvider } from 'packages/react';
```
marisbest2, line 6: this seems dangerous / we lose the NX tree I think

packages/react/src/provider.tsx
```diff
@@ -108,6 +114,7 @@ export type GetAllergensReturn = {
 };
 
 export interface PhotonClientContextInterface {
+  clinicalClient: ApolloClient<object> | undefined;
```
marisbest2, line 117: this should probably be `ApolloClient<NormalizedCacheObject>`

packages/sdk/src/lib.ts
```diff
@@ -43,8 +43,19 @@ export class PhotonClient {
...
 
+  /**
+   * The GraphQL endpoint of Service's clinical API
+   * */
+  private clinicalApiUri?: string;
+
+  /**
+   * The clinical app url
+   */
   public clinicalUrl?: string;
```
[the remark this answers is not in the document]
captDaylight: 100% agree

packages/elements/src/photon-client/index.tsx
```diff
@@ -119,29 +119,38 @@ customElement(
...
+
+    createEffect(() => {
       if (!store()?.authentication.state.isLoading) {
         if (!store()?.authentication.state.isAuthenticated && props.autoLogin) {
           const args: any = { appState: {} };
           if (props.redirectPath) {
             args.appState.returnTo = props.redirectPath;
           }
-          await store()?.authentication.login(args);
+          store()?.authentication.login(args);
```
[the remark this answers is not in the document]
captDaylight: solid barks if you have a async function handling a reactive effect. The one above I just removed the logic that needs to be async, but in this createEffect case it doesn't need to await on anything

apps/app/src/App.tsx
captDaylight: fixed

packages/elements/src/photon-client/index.tsx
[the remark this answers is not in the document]
marisbest2: got it

packages/sdk/src/lib.ts
captDaylight: check again

apps/app/src/App.tsx
```diff
@@ -24,16 +24,15 @@ import { Prescription } from './views/routes/Prescription';
 import { PrescriptionForm } from './views/routes/PrescriptionForm';
 import { Prescriptions } from './views/routes/Prescriptions';
 import { Settings } from './views/routes/Settings';
-import { ClinicalApiProvider } from './views/routes/Settings/apollo';
 import { Support } from './views/routes/Support';
 import { UpdatePatientForm } from './views/routes/UpdatePatientForm';
+import { Env } from 'packages/sdk/src/utils';
```
marisbest2, line 29: this should be imported from `@photonhealth/sdk`

packages/sdk/src/lib.ts
```diff
@@ -92,26 +110,36 @@ export class PhotonClient {
       redirect_uri: redirectURI,
       cacheLocation: 'memory'
     });
-    this.audience = audience;
-    this.uri = uri;
-    this.clinicalUrl = getClinicalUrl(uri);
+    this.audience = env ? lambdasApiUrl[env] : audience;
+    this.uri = env ? `${lambdasApiUrl[env]}/graphql` : uri;
+    this.clinicalUrl = env ? clinicalAppUrl[env] : getClinicalUrl(uri);
```
marisbest2, line 115: I think you want the opposite. Especially if `env` has a default value.

marisbest2 (approved):
Approved with comments

apps/app/src/App.tsx
captDaylight: fixed

packages/sdk/src/lib.ts
captDaylight: makes sense, changing

PR #800, opened 2024-01-05
Remove pharmacy ratings

jknapp25:
Ratings are costing us $2-3k/mo, we're removing to save money given these don't seem vital to pharmacy selection.


apps/patient/src/views/Status.tsx
```diff
@@ -45,10 +44,13 @@ export const Status = () => {
 
   const [submitting, setSubmitting] = useState<boolean>(false);
   const [successfullySubmitted, setSuccessfullySubmitted] = useState<boolean>(false);
-  const [enrichedPharmacy, setEnrichedPharmacy] = useState<EnrichedPharmacy | undefined>(undefined);
 
   const { fulfillment, pharmacy, address } = order;
 
+  // const [enrichedPharmacy, setEnrichedPharmacy] = useState<EnrichedPharmacy | undefined>(
```
captDaylight, line 50: can get rid of the comments here and below?

apps/patient/src/views/Status.tsx
```diff
@@ -215,10 +206,10 @@ export const Status = () => {
               ) : null}
             </Box>
           </VStack>
-          {enrichedPharmacy ? (
+          {pharmacy ? (
             <Box width="full">
               <PharmacyCard
-                pharmacy={enrichedPharmacy}
+                pharmacy={pharmacyWithHours}
```
captDaylight, line 212: I see this was changed from `enrichedPharmacy`, but in `Pharmacy.tsx` I see you're using the type with that name too... is a pharmacy with hours = enriched?

captDaylight (approved):

apps/patient/src/views/Status.tsx
jknapp25: yeah

PR #804, opened 2024-01-08
add edit profile

photonsFardad:
image

image


apps/app/src/views/routes/Settings/components/profile/ProfileEditForm.tsx
```diff
@@ -0,0 +1,479 @@
...
+import { FormikStateSelect, yupStateSchema } from '../utils/States';
+import { FormikTouched, FormikErrors, ErrorMessage, Field, Formik } from 'formik';
+import { Role } from 'packages/sdk/dist/types';
+
+interface EditProfileActionProps {
+  userId: string;
+  onClose: () => void;
+}
+
+const EditProfileActionGetUserQuery = graphql(/* GraphQL */ `
```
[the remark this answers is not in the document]
marisbest2: Can we use `me` here instead?

apps/app/src/views/routes/Settings/components/profile/Profile.tsx
```diff
@@ -2,28 +2,42 @@ import {
...
 
 const profileQuery = graphql(/* GraphQL */ `
   query MeProfileQuery {
     me {
+      id
       name {
+        first
+        last
+        middle
+        title
```
marisbest2, line 35: Do we need all these fields at this level?

If I'm reading what you're trying to do correctly, what you want is a fragment, defined down on the form level, and then included up here. And then you'd use `useFragment` in the form. You can roughly copy from `InviteItem` or `UserItem`

apps/app/src/views/routes/Settings/components/profile/ProfileEditForm.tsx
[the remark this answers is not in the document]
marisbest2: We should actually be using a fragment here

marisbest2 (changes requested):
We should use a fragment instead of another query. 
And ideally this would work in similar UX to how we have it for Organization

apps/app/src/views/routes/Settings/components/profile/Profile.tsx
photonsFardad: done, updated to a common user fragment

apps/app/src/views/routes/Settings/components/profile/ProfileEditForm.tsx
```diff
@@ -0,0 +1,479 @@
...
+import { rolesSchema } from '../utils/Roles';
+import { useMutation, useQuery } from '@apollo/client';
+import { useClinicalApiClient } from '../../apollo';
+import { graphql } from 'apps/app/src/gql';
+import { FormikStateSelect, yupStateSchema } from '../utils/States';
+import { FormikTouched, FormikErrors, ErrorMessage, Field, Formik } from 'formik';
+import { Role } from 'packages/sdk/dist/types';
+
+interface EditProfileActionProps {
+  userId: string;
```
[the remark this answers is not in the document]
photonsFardad: removed

apps/app/src/views/routes/Settings/components/profile/ProfileEditForm.tsx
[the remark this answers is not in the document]
photonsFardad: done, updated to a common user fragment

thebadprince:
@photonsFardad Looks good to me. Can we add a confirmation toast that says `Profile updated` after they hit the `Update Profile` button? Similar to how we have it after a user creates a template:

Screenshot 2024-01-09 at 1 33 43 PM

IMO good practice for us UX-wise to always have success/error feedback for the user.

apps/app/src/views/routes/Settings/components/profile/Profile.tsx
```diff
@@ -90,10 +112,29 @@ export const Profile = () => {
...
+
+            <Box>
+              <Button size={'sm'} colorScheme={'brand'} leftIcon={<EditIcon />} onClick={onOpen}>
+                Update Profile
+              </Button>
+            </Box>
+          </Flex>
+          <Modal isOpen={isOpen} onClose={onClose} size="xl">
+            <ModalOverlay />
+            {isOpen && (
```
marisbest2, line 132: This should only show up if `data` is not null.

apps/app/src/views/routes/Settings/components/profile/ProfileEditForm.tsx
```diff
@@ -0,0 +1,431 @@
...
+import { useMutation } from '@apollo/client';
+import { usePhoton } from '@photonhealth/react';
+import { FragmentType, graphql, useFragment } from 'apps/app/src/gql';
+import { FormikStateSelect, yupStateSchema } from '../utils/States';
+import { FormikTouched, FormikErrors, ErrorMessage, Field, Formik } from 'formik';
+import { Role } from 'packages/sdk/dist/types';
+import { userFragment } from '../utils/UserFragment';
+
+interface EditProfileActionProps {
+  user?: FragmentType<typeof userFragment>;
```
marisbest2, line 29: This should be `user: FragmentType<blah>`. Its always required and non-nullable

apps/app/src/views/routes/Settings/components/profile/ProfileEditForm.tsx
```diff
@@ -0,0 +1,431 @@
...
+} from '@chakra-ui/react';
+import * as yup from 'yup';
+import { rolesSchema } from '../utils/Roles';
+import { useMutation } from '@apollo/client';
+import { usePhoton } from '@photonhealth/react';
+import { FragmentType, graphql, useFragment } from 'apps/app/src/gql';
+import { FormikStateSelect, yupStateSchema } from '../utils/States';
+import { FormikTouched, FormikErrors, ErrorMessage, Field, Formik } from 'formik';
+import { Role } from 'packages/sdk/dist/types';
+import { userFragment } from '../utils/UserFragment';
```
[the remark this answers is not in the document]
marisbest2: It also helps with caching as I _thing_ fragments are cached by name

apps/app/src/views/routes/Settings/components/users/UsersList.tsx
```diff
@@ -35,14 +35,32 @@ const usersQuery = graphql(/* GraphQL */ `
...
       name {
+        first
         full
+        last
+        middle
+        title
       }
       roles {
+        description
+        id
```
marisbest2, line 61: I dont think you need any of these fields.

apps/app/src/views/routes/Settings/components/utils/UserFragment.tsx
```diff
@@ -0,0 +1,30 @@
+import { graphql } from 'apps/app/src/gql';
+export const userFragment = graphql(/* GraphQL */ `
```
marisbest2, line 2: This should be defined and in the specific files its being used and only specify the specific fields needed. It should also be named something like `<Component>UserFragment` (e.g. `EditRolesActionUserFragment`). Fragments need to be globally uniquely named

marisbest2 (changes requested):
Needs some changes for fragment encapsulation / fragment colocation. (you can read up on Relay-style fragments, though we dont need to go full relay) 

Also, I think it'd be much better UX if its just an inline form, similar to how it works for Organization, rather than a modal.

apps/app/src/views/routes/Settings/components/profile/Profile.tsx
```diff
@@ -2,36 +2,56 @@ import {
...
+        first
+        full
+        last
+        middle
+        title
+      }
+      roles {
+        description
+        id
+        name
```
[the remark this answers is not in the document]
photonsFardad: We are they can update them

apps/app/src/views/routes/Settings/components/profile/ProfileEditForm.tsx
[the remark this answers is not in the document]
photonsFardad: done

apps/app/src/views/routes/Settings/components/profile/Profile.tsx
photonsFardad: done

apps/app/src/views/routes/Settings/components/profile/ProfileEditForm.tsx
photonsFardad: done

apps/app/src/views/routes/Settings/components/utils/UserFragment.tsx
photonsFardad: done

photonsFardad: removed

apps/app/src/views/routes/Settings/components/profile/Profile.tsx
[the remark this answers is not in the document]
photonsFardad: extra fields removed

apps/app/src/views/routes/Settings/components/users/UsersList.tsx
photonsFardad: updated, extra fields removed

apps/app/src/views/routes/Settings/components/profile/ProfileEditForm.tsx
```diff
@@ -0,0 +1,431 @@
...
+type ProviderYupType = ProfileYupType['provider'];
+type ProviderFormikTouchedType = FormikTouched<ProviderYupType>;
+type ProviderFormikErrorsType = FormikErrors<ProviderYupType>;
+
+export const EditProfileAction: React.FC<EditProfileActionProps> = ({ user, onClose }) => {
+  const { clinicalClient } = usePhoton();
+  const currentUser = useFragment(userFragment, user);
+  const [updateMyProfile, { error }] = useMutation(UpdateMyProfileMutation, {
+    client: clinicalClient,
+    refetchQueries: ['MeProfileQuery', 'EditProfileActionGetUserQuery']
```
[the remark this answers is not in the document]
photonsFardad: done

thebadprince:
Heads up that the modal title still says: `Assign roles to user`. Should be something like `Edit Profile` 

Screenshot 2024-01-09 at 6 54 00 PM


That said, agree with @marisbest2 here if we can do it:

> Also, I think it'd be much better UX if its just an inline form, similar to how it works for Organization, rather than a modal.

If so, we won't need the title of the modal.

apps/app/src/views/routes/Settings/components/profile/Profile.tsx
```diff
@@ -233,6 +235,12 @@ export const Profile = () => {
                   })
             }
           });
+          toast({
+            title: 'Role updated',
```
thebadprince, line 239: We'll need this to be `Profile updated` instead of `Role updated`

thebadprince:
Looks good to me aside from the comment above. Thanks for these changes. For some reason, mine is missing Name though:

Screenshot 2024-01-10 at 10 56 32 AM

apps/app/src/views/routes/Settings/components/profile/Profile.tsx
photonsFardad: done

marisbest2 (approved):

PR #819, opened 2024-01-14
Fix empty medical history bug and fix medical history link

samkotlove:


packages/components/src/systems/PatientMedHistory/index.tsx
```diff
@@ -64,7 +64,7 @@ export default function PatientMedHistory(props: PatientMedHistoryProps) {
   const client = usePhotonClient();
   const [medHistory, setMedHistory] = createSignal<PatientMedication[] | undefined>(undefined);
 
-  const baseURL = createMemo(() => `${client?.clinicalUrl}prescriptions/`);
+  const baseURL = createMemo(() => `${client?.clinicalUrl}/prescriptions/`);
```
samkotlove, line 67: We should be using a url builder for situations like this to avoid these edge cases

PR #830, opened 2024-01-16
Zendesk Form in Clinical App

captDaylight:
- moves order page buttons to the top right
- creates a new component for submitting Zendesk tickets
- unifies how toasts look across our chakra and elements apps

Can we remove Zendesk chat after this?

Screen Shot 2024-01-16 at 10 40 27 AM


apps/app/codegen.ts
```diff
@@ -3,7 +3,7 @@ import { CodegenConfig } from '@graphql-codegen/cli';
 const config: CodegenConfig = {
   //schema: 'http://clinical-api.tau.health:8080/graphql',
   schema: 'http://clinical-api.boson.health/graphql',
-  documents: ['apps/app/src/**/Settings/**/*.tsx'],
+  documents: ['apps/app/src/**/Settings/**/*.tsx', 'apps/app/src/views/components/TicketModal.tsx'],
```
captDaylight, line 6: @marisbest2  and I talked about this, this might get messy adding single files but fine for now. Long term, we can add some sort of identifier to the graphql operation that identifies which schema it's pointing to.

apps/app/src/views/components/StyledToast.tsx
```diff
@@ -0,0 +1,49 @@
+import { Box, HStack, IconButton, Text, Icon, VStack } from '@chakra-ui/react';
+import { FiCheckCircle, FiX, FiInfo } from 'react-icons/fi';
+
+type ToastType = 'success' | 'info';
```
captDaylight, line 4: Screen Shot 2024-01-16 at 10 37 28 AM

apps/app/src/views/routes/Order.tsx
```diff
@@ -186,10 +193,36 @@ const cancelReasons = [
   'Other'
 ];
 
+export default function uniqueFills(order: OrderType): Fill[] {
```
marisbest2, line 196: why is this a `default` export?

apps/app/src/views/routes/Order.tsx
```diff
@@ -339,354 +373,446 @@ export const Order = () => {
...
+    instructions: fill?.prescription?.instructions
+  })}
+`
+)}
+
+---- 
+Description: 
+        `
+        }
+      />
```
marisbest2, line 459: Can we put this in its own file / component?

apps/app/src/views/routes/Order.tsx
```diff
@@ -339,354 +373,446 @@ export const Order = () => {
...
+              borderColor="red.500"
+              textColor="red.500"
+              colorScheme="red"
+              isLoading={updating}
+              loadingText="Canceling..."
+              isDisabled={
+                loading ||
+                order.state === types.OrderState.Canceled ||
+                order.state === types.OrderState.Completed ||
+                order?.fulfillment?.state === 'SHIPPED'
```
marisbest2, line 476: Shouldnt it be hidden if these are the case?

apps/app/src/views/routes/Order.tsx
```diff
@@ -339,354 +373,446 @@ export const Order = () => {
...
-                />
+                <Stack flexShrink={0}>
+                  <OrderStatusBadge
+                    fulfillmentState={order.fulfillment?.state}
+                    orderState={order.state}
+                  />
+                </Stack>
               )}
+
+              <CopyText size="xs" text={order?.id} />
```
marisbest2, line 543: should this should if `order?.id` is null?

marisbest2 (approved):
Left some comments. Hard to tell whats actually changed...

apps/app/src/views/routes/Order.tsx
captDaylight: You mean specifically the template literal? I can move it out

apps/app/src/views/routes/Order.tsx
captDaylight: This is so that in some states the button is mysteriously gone, leaving the user confused. a better version of this also explains why it's disabled.

apps/app/src/views/routes/Order.tsx
captDaylight: The component returns null
```
if (!text) return null;
```

PR #847, opened 2024-01-17
small refactor for login logic

marisbest2:
So there WAS a bug here. These two code paths need to be mutually exclusive

### DIFF
--- a/packages/elements/src/photon-client/index.tsx
+++ b/packages/elements/src/photon-client/index.tsx
@@ -162,24 +162,19 @@ customElement(
     });
 
     createEffect(() => {
-      if (!store()?.authentication.state.isLoading && props.externalUserId != null) {
-        if (
-          (store().authentication.state.user as User).sub?.split('|').reverse()[0] !==
-          props.externalUserId
-        ) {
-          store()?.authentication.logout();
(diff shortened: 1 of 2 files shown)


jknapp25 (approved):

packages/elements/src/photon-client/index.tsx
```diff
@@ -162,24 +162,19 @@ customElement(
...
-        }
-      }
-    });
-
-    createEffect(() => {
-      if (!store()?.authentication.state.isLoading) {
-        if (!store()?.authentication.state.isAuthenticated && props.autoLogin) {
+        } else if (
+          props.externalUserId != null &&
+          (store().authentication.state.user as User | undefined)?.sub?.split('|').reverse()[0] !==
```
samkotlove, line 174: I still think we should be doing a substring check instead of doing the whole split thing

samkotlove (commented):
Need some comments, this is hard to track what the intent is (true previously as well)

packages/elements/src/photon-client/index.tsx
marisbest2: that still doesnt work because ids arent guaranteed to not be substrings of each other

captDaylight (approved):

PR #848, opened 2024-01-18
Mrochlin/spinner not logged in

marisbest2:
Fix Max's issue where he sees the "You are not logged in" during the Oauth dance. Instead show a spinner


packages/elements/src/photon-authorized/index.tsx
```diff
@@ -69,7 +70,19 @@ export const PhotonAuthorized = (p: { children: JSXElement; permissions?: Permis
...
+              fallback={<AlertMessage message="You are not signed in" />}
+            >
+              {/* 
+                If using auto login, we expect that this will only momentarily be visible
+                Either it'll kick out to an auth page, or it'll do the background OAuth dance
+                In either case, a spinner is the correct design
+              */}
+              <div class="w-full flex justify-center">
+                <Spinner color="green" />
+              </div>
```
samkotlove, line 84: What about when autoLogin is true but login failed? is that even possible?

samkotlove (approved):

PR #855, opened 2024-01-18
Tracking Combine and Duplicate orders in the prescribe workflow

captDaylight:
I didn't have any luck triggering datadog actions from within the elements, so next best thing I'm creating events that trigger the action in react.

[Example of an event triggered when the duplicate dialog is opened.](https://app.datadoghq.com/rum/sessions?query=%40type%3Aaction%20%40action.target.name%3Aduplicate-order-dialog-open&cols=&event=AgAAAY0eOKxEMRd05wAAAAAAAAAYAAAAAEFZMGVPTWtEQUFCR3RiczV0TXRSR3dBUQAAACQAAAAAMDE4ZDFlM2UtNmQxYy00ZTgyLWJmMTgtMzNhODIwMjczMmYy&viz=stream&from_ts=1705523904499&to_ts=1705610304499&live=true)

- right now all of the events that I'm triggering have empty custom attributes. it could be useful in the future, do you think I should be passing anything in with this? we're just looking to tally up the events right?
- I switched `RecentOrders` into it's own file, I'm going to chip away at this so that we don't have any named components with the name `index`


packages/components/src/systems/RecentOrders/RecentOrders.tsx
```diff
@@ -0,0 +1,201 @@
+import { createContext, JSXElement, useContext, createEffect } from 'solid-js';
```
captDaylight, line 1: ignore this, just copying it over from `index`

marisbest2:
My comments:
- Use named exports (in addition to a default if you really want)
- We shouldn't rely on having a raw HTML element to dispatch the event to then forward up. Instead, create an empty hidden `div` in the top level `<photon-client>`, attach that `div`'s `ref` to the `store` and then add a `dispatchEvent` (or better, `datadogEvent` to the `store`. This avoids any issues if we want to trigger a datadog event without a raw HTML element

packages/components/src/utils/dispatchDatadogAction.ts
```diff
@@ -0,0 +1,17 @@
+export const dispatchDatadogAction = (
+  action: string,
+  data: {
+    [key: string]: unknown;
+  },
+  ref?: HTMLElement
```
marisbest2, line 6: This should probably be `{ dispatchEvent(CustomEvent): void }` and not be optional.

captDaylight:
@jknapp25 timestamp makes sense on the action, does that not come for free with the action? the other info would seem to be better associated with the session (org) and view (patient id on the prescription/new), wdyt?

jknapp25:
@captDaylight I think it does, but i can't find it.

Oh sorry I meant prescriber (userId) not patient.

apps/app/src/views/routes/Main.tsx
```diff
@@ -39,6 +40,16 @@ export const Main = () => {
       localStorage.removeItem('previouslyAuthed');
       setPreviouslyAuthed(false);
     }
+    if (isAuthenticated && !isLoading) {
+      // global context to the datadog RUM session
+      datadogRum.addRumGlobalContext('org', {
+        orgId: user.org_id
+      });
+      datadogRum.setUser({
```
jknapp25, line 48: didn't know about these, awesome!

jknapp25 (approved):

PR #857, opened 2024-01-18
Mrochlin/merge bot

marisbest2:
This here is the merge bot. The problem we have today is that our `boson` commit log is rather... unwieldly. This is because we don't use squash commits. We dont use squash commits because we roughly follow `gitflow` which says that our production (i.e. photon/neutron) branches should have a history characterized by `merge` commits. This is good because its pretty obvious how to read the commit log for Neutron/Photon and how to roll them back.

BUT. Github is a pain and won't let us set merge strategies on a per-branch basis. 

So the plan is:
- Merges to `boson` should always be SQUASH merges. This will be enforced by requiring that `boson` have a linear git history. Yay!
- Merges to `photon` and `neutron` should always be MERGE merges. In order to enforce this, we'll make it so that the only way to merge into photon/neutron is using this here "merge-bot".

How does the merge-bot work? I'm glad you asked.
When you add the label `ready`, the merge bot will tell github to set this PR to "auto merge when ready". GH's auto-merge feature (which is relatively new) will wait for required reviews and required checks. Then, once those are done, it'll merge using the strategy set (which is the `merge` strategy!)


captDaylight:
@marisbest2 mind adding a few words about what you're trying to achieve

.github/workflows/merge-bot.yml
```diff
@@ -0,0 +1,36 @@
...
+      - edited
+  pull_request_review:
+    types:
+      - dismissed
+      - submitted
+
+jobs:
+  merge:
+    name: Merge
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'ready') }}
```
samkotlove, line 21: Will this react if a pull request is good to go (approved, test passing etc.) and _then_ I add ready?

.github/workflows/merge-bot.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
+      - uses: tibdex/github-app-token@v1
+        id: generate-token
+        with:
+          app_id: ${{ secrets.CPR_APP_ID }}
+          private_key: ${{ secrets.CPR_APP_SECRET }}
```
samkotlove, line 31: believe this was used in another job but what's CPR?

.github/workflows/merge-bot.yml
marisbest2: yeah

marisbest2: All the bot does is tell github "merge when Github thinks its ready"

captDaylight (approved):

samkotlove (approved):

PR #861, opened 2024-01-19
Ref in SDK

captDaylight:
This change to not require a ref param for `dispatchDatadogAction` was going to go in with https://github.com/Photon-Health/client/pull/855 but ran into an issue that I want to highlight.

Adding the ref to the `components` Provider is pretty straightforward. (though maybe a reason to do this in elements).

The problem comes up when I try to use the function, specifically in `RecentOrdersDuplicateDialog`. When I use it in a createEffect no problem. However, when I use it as an `onClick` the `usePhotonClient` triggers `usePhotonClient must be used within a SDKProvider`.

So first question is, am I making a mistake?

If not, is the extra overhead of all the `dispatchDatadogAction` only being called in a `createEffect` which often means an signal to trigger to signal it...


apps/app/src/views/routes/Main.tsx
```diff
@@ -67,6 +69,24 @@ export const Main = () => {
     }
   }, [isLoading, error]);
 
+  useEffect(() => {
```
captDaylight, line 72: move this this event handler up

packages/components/src/systems/SDKProvider/index.tsx
```diff
@@ -1,17 +1,36 @@
...
+    <SDKContext.Provider value={store}>
+      <div ref={ref}>{props.children}</div>
+    </SDKContext.Provider>
+  );
 }
 
 export function usePhotonClient() {
-  return useContext(SDKContext);
+  const context = useContext(SDKContext);
+  if (!context) {
```
captDaylight, line 32: check that the context is available

packages/components/src/systems/RecentOrders/RecentOrdersDuplicateDialog.tsx
```diff
@@ -8,24 +8,27 @@ import { dispatchDatadogAction } from '../../utils/dispatchDatadogAction';
...
   createEffect(() => {
     if (state.isDuplicateDialogOpen) {
-      dispatchDatadogAction('prescribe-duplicate-dialog-open', {}, ref);
+      dispatchDatadogAction('prescribe-duplicate-dialog-open', {});
+    }
+  });
+
+  createEffect(() => {
+    if (exit()) {
+      dispatchDatadogAction('prescribe-duplicate-dialog-exit', {});
```
captDaylight, line 22: This attempt to trigger the `prescribe-duplicate-dialog-exit` works.

packages/components/src/systems/RecentOrders/RecentOrdersDuplicateDialog.tsx
```diff
@@ -8,24 +8,27 @@ import { dispatchDatadogAction } from '../../utils/dispatchDatadogAction';
...
   });
 
   return (
     <Dialog
       open={state.isDuplicateDialogOpen}
-      onClose={() => {
-        dispatchDatadogAction('prescribe-duplicate-dialog-exit', {}, ref);
-        actions.setIsDuplicateDialogOpen(false);
-      }}
+      onClose={() => actions.setIsDuplicateDialogOpen(false)}
```
captDaylight, line 29: ...but if I try to trigger `prescribe-duplicate-dialog-exit` here it will fail and say it doesn't have access to the context.

packages/components/src/systems/RecentOrders/RecentOrdersDuplicateDialog.tsx
```diff
@@ -8,24 +8,27 @@ import { dispatchDatadogAction } from '../../utils/dispatchDatadogAction';
...
 
 export default function RecentOrdersDuplicateDialog() {
-  let ref: Ref<any> | undefined;
   const [state, actions] = useRecentOrders();
+  const [exit, setExit] = createSignal(false);
 
   createEffect(() => {
     if (state.isDuplicateDialogOpen) {
-      dispatchDatadogAction('prescribe-duplicate-dialog-open', {}, ref);
+      dispatchDatadogAction('prescribe-duplicate-dialog-open', {});
```
captDaylight, line 16: This works as expected.

packages/components/src/systems/SDKProvider/index.tsx
```diff
@@ -1,17 +1,36 @@
...
+    ref
+  });
+
+  createEffect(() => {
+    setStore({ client: props.client, ref });
+  });
+
+  return (
+    <SDKContext.Provider value={store}>
+      <div ref={ref}>{props.children}</div>
```
marisbest2, line 25: Oh I DEFINITELY wouldn't do this. Do
```suggestion
      <div ref={ref} hidden />{props.children}
```

marisbest2: Otherwise youre changing the tree and it can get real messy

PR #863, opened 2024-01-21
remove user

photonsFardad:


apps/app/apps/app/src/gql/fragment-masking.ts
marisbest2, line 1: Where did this come from?

apps/app/apps/app/src/gql/gql.ts
marisbest2, line 1: Looks like you somehow got some files added

apps/app/src/views/routes/Settings/components/users/UserItemActions.tsx
```diff
@@ -48,8 +55,21 @@ interface UserItemActionsProps {
   user: FragmentType<typeof userFragment>;
 }
 
+const RemoveUserFromOrganizationMutation = graphql(/* GraphQL */ `
+  mutation RemoveUserFromOrganization($userId: ID!) {
```
marisbest2, line 59:
```suggestion
  mutation UserItemActionRemoveUserFromOrganization($userId: ID!) {
```

apps/app/src/views/routes/Settings/components/users/UserItemActions.tsx
```diff
@@ -63,6 +83,54 @@ export const UserItemActions: React.FC<UserItemActionsProps> = ({ user }) => {
           <MenuItem icon={<FiEdit fontSize="1.2em" />} onClick={onOpen}>
             Edit Roles
           </MenuItem>
+          <MenuItem
+            icon={<FiTrash fontSize="1.3em" color="red" />}
+            disabled={deleteLoading}
+            onClick={async () => {
+              const decision = await confirmWrapper('Remove user?', {
+                description: <Text mb={2}>This will remove user from organization.</Text>,
```
marisbest2, line 91: Is there a reason we're overloading the styling here?

apps/app/src/views/routes/Settings/components/users/UserItemActions.tsx
```diff
@@ -63,6 +83,54 @@ export const UserItemActions: React.FC<UserItemActionsProps> = ({ user }) => {
           <MenuItem icon={<FiEdit fontSize="1.2em" />} onClick={onOpen}>
             Edit Roles
           </MenuItem>
+          <MenuItem
+            icon={<FiTrash fontSize="1.3em" color="red" />}
+            disabled={deleteLoading}
+            onClick={async () => {
```
marisbest2, line 89: Can we extract this to a method higher up and use `useCallback`. We should just default get in the habit of this. 

See https://semaphoreci.com/blog/react-usecallback-hook#:~:text=Benefits%20of%20using%20the%20useCallback%20hook&text=Restricting%20rendering%20of%20child%20components,as%20props%20to%20child%20components.

apps/app/src/views/routes/Settings/components/users/UserItemActions.tsx
```diff
@@ -63,6 +83,54 @@ export const UserItemActions: React.FC<UserItemActionsProps> = ({ user }) => {
...
+              const decision = await confirmWrapper('Remove user?', {
+                description: <Text mb={2}>This will remove user from organization.</Text>,
+                cancelText: 'Cancel',
+                confirmText: 'Delete',
+                colorScheme: 'red'
+              });
+              if (decision) {
+                removeUser({
```
marisbest2, line 97: I think you want to `await` this

apps/app/src/views/routes/Settings/components/users/UserItemActions.tsx
```diff
@@ -63,6 +83,54 @@ export const UserItemActions: React.FC<UserItemActionsProps> = ({ user }) => {
...
+                  variables: {
+                    userId: userData.id
+                  }
+                });
+                toast({
+                  position: 'top',
+                  // TODO: Override default solid theme with this outline variant
+                  render: ({ onClose }) => (
```
marisbest2, line 105: I _think_ that Paul (@captDaylight) made a new toast component that has consistent styling

apps/app/src/views/routes/Settings/components/users/UserItemActions.tsx
marisbest2, line 91: Also, text should say `This will remove the user from the organization`, or better: `{{ user.name }} will be removed from {{ organization.name }}`.

marisbest2 (changes requested):
Left comments. Need to fix prettier. 

Playing with it, it looks like the toast fires before the delete is actually processed. I think ideally the `UserList` should go into a loading state when the delete processes, or the modal shouldn't close until its been refetched. Probably adding the `await` and maybe also maybe add [`awaitRefetchQueries: true`](https://www.apollographql.com/docs/react/data/mutations/#awaitrefetchqueries) to the mutation options should do it.

Also, I think that the button text should be red in the menu, but defer to @thebadprince on that

apps/app/src/views/routes/Settings/components/users/UserItemActions.tsx
photonsFardad: this is copy paste from delete invite that was approved

photonsFardad: It feels like incosistency that 2 deletes in 2 screens behave differently. sure I will change

apps/app/src/views/routes/Settings/components/users/UserItemActions.tsx
marisbest2: We should generally start naming mutations and fragments based on where they are being used in addition to what they do. This is inspired by the much stronger naming rule enforced by Relay. The point is that multiple places in code should be able to call `removeUserFromOrganization`, so either individual spot its used should be identifiable (and independently cached, but thats more for fragments)

apps/app/src/views/routes/Settings/components/users/UserItemActions.tsx
marisbest2: Its called `StyledToast`

apps/app/apps/app/src/gql/fragment-masking.ts
photonsFardad: I think auto added, I can delete see if destroys anything

apps/app/src/views/routes/Settings/components/users/UserItemActions.tsx
photonsFardad: Is there an issue with it?

apps/app/src/views/routes/Settings/components/users/UserItemActions.tsx
photonsFardad: done

apps/app/src/views/routes/Settings/components/users/UserItemActions.tsx
photonsFardad: done

apps/app/src/views/routes/Settings/components/users/RemoveUserActionItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+      middle
+      title
+    }
+    roles {
+      description
+      id
+      name
+    }
+  }
+`);
```
marisbest2, line 39: I dont think you need all of these? Looks like you only need `id` and `name`. 

Also, the fragment name should be `RemoveUserActionItemUserFragment`. And then you need to actually use that fragment

apps/app/src/views/routes/Settings/components/users/RemoveUserActionItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+      icon={<FiTrash fontSize="1.3em" color="red" />}
+      disabled={deleteLoading}
+      onClick={handleRemoveUser}
+    >
+      Remove User
+    </MenuItem>
+  );
+};
+
+export default RemoveUserMenuItem;
```
marisbest2, line 99: Prefer named exports. (we should set up a lint for this)

apps/app/src/views/routes/Settings/components/users/RemoveUserActionItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const { clinicalClient } = usePhoton();
+  const userData = useFragment(userFragment, user);
+  const [removeUser, { loading: deleteLoading }] = useMutation(RemoveUserFromOrganizationMutation, {
+    client: clinicalClient,
+    refetchQueries: ['UsersListQuery']
+  });
+
+  const handleRemoveUser = async () => {
+    const decision = await confirmWrapper('Remove user?', {
+      description: `user - ${userData.name?.full} will be removed from your organization`,
```
marisbest2, line 64:
```suggestion
      description: `${userData.name?.full ?? "User"} will be removed from your organization`,
```

marisbest2: Also, is `name` actually optional on user? Seems weird...

apps/app/src/views/routes/Settings/components/users/RemoveUserActionItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+`);
+
+const RemoveUserMenuItem: React.FC<RemoveUserMenuItemProps> = ({ user, onDelete }) => {
+  const toast = useToast();
+  const { colorMode } = useColorMode();
+  const { clinicalClient } = usePhoton();
+  const userData = useFragment(userFragment, user);
+  const [removeUser, { loading: deleteLoading }] = useMutation(RemoveUserFromOrganizationMutation, {
+    client: clinicalClient,
+    refetchQueries: ['UsersListQuery']
```
marisbest2, line 59: I think this needs `awaitRefetchQueries`

apps/app/src/views/routes/Settings/components/users/RemoveUserActionItem.tsx
```diff
@@ -0,0 +1,99 @@
+// RemoveUserMenuItem.tsx
```
marisbest2, line 1: This comment isnt needed, but also it doesnt match the file name

apps/app/src/views/routes/Settings/components/users/RemoveUserActionItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { MenuItem, useColorMode, useToast } from '@chakra-ui/react';
+import { FragmentType, graphql, useFragment } from 'apps/app/src/gql';
+import { FiTrash } from 'react-icons/fi';
+import { confirmWrapper } from '../../../../components/GuardDialog';
+import { useMutation } from '@apollo/client';
+import { usePhoton } from '@photonhealth/react';
+import { StyledToast } from 'apps/app/src/views/components/StyledToast';
+
+export const userFragment = graphql(/* GraphQL */ `
+  fragment UserFragment on User {
```
marisbest2, line 12: Fragment names need to be globally unique

apps/app/src/views/routes/Settings/components/users/RemoveUserActionItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+      colorScheme: 'red'
+    });
+
+    if (decision) {
+      await removeUser({
+        variables: {
+          userId: userData.id
+        }
+      });
+      onDelete();
```
marisbest2, line 77: I think you should probably call this AFTER the toast. `onDelete` could lead to unmounting this component which could end up weird.

apps/app/src/views/routes/Settings/components/users/RemoveUserActionItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+      toast({
+        position: 'top-right',
+        duration: 4000,
+        render: ({ onClose }) => (
+          <StyledToast onClose={onClose} type="success" description="Profile updated" />
```
marisbest2, line 82: Should say `User removed`

apps/app/src/views/routes/Settings/components/users/RemoveUserActionItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const decision = await confirmWrapper('Remove user?', {
+      description: `user - ${userData.name?.full} will be removed from your organization`,
+      cancelText: 'Cancel',
+      confirmText: 'Yes, Remove',
+      darkMode: colorMode !== 'light',
+      colorScheme: 'red'
+    });
+
+    if (decision) {
+      await removeUser({
```
marisbest2, line 72: What happens if this fails?

marisbest2 (changes requested):

thebadprince:
> Also, I think that the button text should be red in the menu, but defer to @thebadprince on that

Agree with @marisbest2. 

Also, I had a spec for the delete user modal, using the same as what we did for `Edit Role`. It's my fault that you didn't see this but it's in the same [Figma link](https://www.figma.com/file/2UzH7WjCFYAExlBNkKqgUz/%F0%9F%A4%96-Automation-Squad?type=design&node-id=868%3A20971&mode=design&t=X7cwMumZVwipry5U-1) as Edit Roles. 

image

apps/app/src/views/routes/Settings/components/users/RemoveUserActionItem.tsx
```diff
@@ -0,0 +1,119 @@
...
+  fragment RemoveUserActionUserFragment on User {
+    id
+    email
+    name {
+      first
+      full
+      last
+      middle
+      title
+    }
```
marisbest2, line 31: I think we only need `full` right?

apps/app/src/views/routes/Settings/components/users/RemoveUserActionItem.tsx
photonsFardad: done

marisbest2 (approved):

PR #868, opened 2024-01-23
Block accepting attestation multiple times on the FE

marisbest2:
I think this is why we were seeing so many errors about that.


samkotlove (approved):
Makes sense

PR #869, opened 2024-01-23
hopefully fix auth0 login_required bug

marisbest2:
We _think_ that the bug where you can be seemingly logged in but not actually is caused by `getAccessToken` returning null for the token. To repro, you can open two tabs, log in, then log out of one and navigate around in the other tab. In previous versions you would see a `Login Required` error message. Now you should get redirected to the login page.

We use `loginWithRedirect` so that if there's auto-login set up, it _should_ just re-execute the SSO handshake

### DIFF
--- a/packages/sdk/src/auth.ts
+++ b/packages/sdk/src/auth.ts
@@ -152,11 +152,12 @@ export class AuthManager {
     } catch (e) {
       if ((e as Error).message.includes('Consent required')) {
         token = await this.authentication.getTokenWithPopup(opts);
-      } else {
-        throw e;
       }
     }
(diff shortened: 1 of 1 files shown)


captDaylight (approved):

marisbest2:
> Where does the login required message come from then?

It comes from the response of `/authorize` which the auth0-spa library calls and handles and then throws `Error("Login Required")` which then gets picked up by the Apollo client because of the call to [`getAccessToken`](https://github.com/Photon-Health/client/blob/b45da50174ba7e55d2379dc2a8b547763c631cea/packages/sdk/src/lib.ts#L159)

PR #877, opened 2024-01-26
Solid query and mutation

captDaylight:
This ports over the `createQuery` and `createMutation` functions from [solid-apollo](https://github.com/merged-js/solid-apollo).

I've updated the `RecentOrders` to demonstrate how to use both functions.

### DIFF
--- a/packages/components/src/utils/createQuery.ts
+++ b/packages/components/src/utils/createQuery.ts
@@ -2,7 +2,7 @@
  * createQuery originates from the solid-apollo package
  * https://github.com/merged-js/solid-apollo/blob/main/src/createQuery.ts
  *
- * Modified slightly to fix types
+ * Modified slightly to fix types, enforces passing a client to work
  */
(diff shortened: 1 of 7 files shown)


packages/components/src/utils/createQuery.ts
```diff
@@ -0,0 +1,80 @@
...
+      return new Promise((resolve, reject) => {
+        const sub = observable.subscribe({
+          error: reject,
+          next: ({ data, error }) => {
+            if (error) {
+              reject(error);
+            }
+
+            if (!resolved) {
+              resolved = true;
```
jknapp25, line 65: does resolved get used?

captDaylight: great question, `resolved` is basically there to make sure that the promise gets resolved the first time the observable returns, and subsequent times it just reconciles the data

marisbest2 (approved):

jknapp25 (approved):

PR #878, opened 2024-01-26
Add urgent option

jknapp25:
Testing an urgent option added to the bottom of urgency options

image

### DIFF
--- a/apps/patient/src/utils/text.ts
+++ b/apps/patient/src/utils/text.ts
@@ -126,13 +126,18 @@ export const text = {
   tracking: 'Tracking #:',
   tryPhoton: 'Try Photon',
   urgencyOptions: [
-    '10:00 am',
-    '12:00 pm',
-    '2:00 pm',
-    '4:00 pm',
-    '6:00 pm',
-    'After hours',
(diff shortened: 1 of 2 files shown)


apps/patient/src/utils/text.ts
```diff
@@ -126,13 +126,18 @@ export const text = {
...
   tryPhoton: 'Try Photon',
   urgencyOptions: [
-    '10:00 am',
-    '12:00 pm',
-    '2:00 pm',
-    '4:00 pm',
-    '6:00 pm',
-    'After hours',
-    'Tomorrow'
+    { label: '10:00 am', description: null, icon: false },
```
jknapp25, line 129: Kind of lame that i changed all of this to an object because of just the urgent one, but I like this more than conditionally showing in the Urgency view

captDaylight: you could do  `{ label: '10:00 am' }` since you check for description below

also could enforce options as `{ label: string; description: string; icon: boolean }  | { label: string };`

captDaylight (approved):

PR #882, opened 2024-01-29
enforce and fix react hook rules

marisbest2:
https://legacy.reactjs.org/docs/hooks-rules.html
https://medium.com/@ryardley/react-hooks-not-magic-just-arrays-cd4f1857236e

### DIFF
--- a/packages/react/src/provider.tsx
+++ b/packages/react/src/provider.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import {
   ApolloClient,
   ApolloError,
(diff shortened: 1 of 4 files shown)


packages/react/src/provider.tsx
marisbest2, line 1: A few notes:
- Reading this file its pretty hard to tell what is a helper function and what is actually being exported. We should fix that
- I'm not sure if anyone is really using the react library, but if they are, we should really start exporting these as actual hooks (i.e. with the `use` naming). This might help avoid weird states where we're calling things out of order
- The deps exhaustiveness checks almost definitely solved some latent bugs, (especially where we were missing the `after` and `first` variables), but it also may lead to overfetching (e.g. where we include `location` in addition to `location.latitude` etc). But better to have the rule (as a warn)

captDaylight (approved):

PR #883, opened 2024-01-29
Fix User Tabs to make it only make the request when the tab is actually loading 

marisbest2:


apps/app/src/views/routes/Settings/index.tsx
```diff
@@ -109,7 +113,8 @@ export const Settings = () => {
 
   return (
     <Page header="Settings">
-      {loading ? (
+      {/* Show loading spinner until the correct tab index has been resolved */}
+      {loading || tabIndex == null ? (
```
captDaylight, line 117: `tabIndex == null` isn't strict equals so this should be totally fine, but since the type of tabIndex is `<number | undefined>` why not `tabIndex === undefined`?

marisbest2: I always do `== null`. I probably was taught once. I forget why

PR #933, opened 2024-02-06
in <SelectOrg />, pass orgId to login if theres only one organization

marisbest2:
Maybe something is hitting this? 
Would explain why:
- We don't see this as much as other people
- When we do see this, it redirects us to `<SelectOrg/>`

### DIFF
--- a/apps/app/src/views/routes/SelectOrg.tsx
+++ b/apps/app/src/views/routes/SelectOrg.tsx
@@ -43,7 +43,8 @@ export const SelectOrg = () => {
         login({
           appState: {
             returnTo: from
-          }
+          },
+          organizationId: organizations[0].id
         });
       }
     }


apps/app/src/views/routes/SelectOrg.tsx
```diff
@@ -43,7 +43,8 @@ export const SelectOrg = () => {
         login({
           appState: {
             returnTo: from
-          }
+          },
+          organizationId: organizations[0].id
```
captDaylight, line 47: do we need to optionally chain this?

marisbest2: I dont think so, because `organizations` has length === 1

captDaylight (approved):

PR #937, opened 2024-02-09
flicker

captDaylight:
fixes login flicker after picking org

https://www.loom.com/share/5ae97c567c4446ae9eba474e84c10c03

### DIFF
--- a/apps/app/src/views/routes/Login.tsx
+++ b/apps/app/src/views/routes/Login.tsx
@@ -63,7 +63,7 @@ export const Login = () => {
             ) : null}
             <HStack spacing="1" justify="center">
               <Text color="muted">Don't have an account?</Text>
-              <Link color="teal.500" href="mailto:rado@photon.health">
+              <Link color="teal.500" href="mailto:hello@photon.health">
                 Contact Sales
               </Link>
             </HStack>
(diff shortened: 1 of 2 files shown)


apps/app/src/views/routes/Login.tsx
```diff
@@ -62,8 +62,8 @@ export const Login = () => {
               </Alert>
             ) : null}
             <HStack spacing="1" justify="center">
-              <Text color="muted">{`Don't have an account?`}</Text>
-              <Link color="teal.500" href="mailto:rado@photon.health">
+              <Text color="muted">Don't have an account?</Text>
+              <Link color="teal.500" href="mailto:sales@photon.health">
```
samkotlove, line 66: I dont think this email is setup yet. You could do hello@photon.health

captDaylight: Otto just set it up https://photonhealth.slack.com/archives/C02NJPUFVPV/p1707508842316309?thread_ts=1707496542.298209&cid=C02NJPUFVPV

marisbest2:
I think we're going to need to revert the <SelectOrg /> thing, so it still needs to work with that

PR #959, opened 2024-02-15
fix default settings

photonsFardad:


apps/app/src/views/routes/PrescriptionForm.tsx
```diff
@@ -90,7 +90,7 @@ export const PrescriptionForm = () => {
       ref.current.patientId = patientId;
     }
   }, [ref.current, patientId]);
-
+  const orgSettings = user?.org_id in settings ? settings[user?.org_id] : settings.default;
```
samkotlove, line 93: would it be better to turn settings into a function, so you can just have it all handled over there? If the org id isnt handled we return the default

photonsFardad: sounds great

jknapp25 (approved):

PR #970, opened 2024-02-21
Rename external user id to be web component prop standard

captDaylight:


packages/elements/src/photon-client/photon-client-component.tsx
```diff
@@ -201,7 +201,7 @@ customElement(
       parse: false
     },
     externalUserId: {
-      attribute: 'userId',
+      attribute: 'user-id',
```
samkotlove, line 204: we will have to notify Max when he upgrades but like the consistency

marisbest2:
This is a breaking change. Either
- bump the minor version (bc we're still on 0.x)
- allow passing either userId or user-id (and maybe warn if they're using userId)

samkotlove:
> This is a breaking change. Either
> 
> * bump the minor version (bc we're still on 0.x)
> * allow passing either userId or user-id (and maybe warn if they're using userId)

You mean bump the major version?

samkotlove (approved):

PR #988, opened 2024-02-29
Add settings for drtelx and redbox

samkotlove:
Includes some consolidation of pharmacy ids as well


packages/settings/src/pharmacies.ts
```diff
@@ -0,0 +1,27 @@
+const ALTO_ID = 'phr_01G9CM93X1NFP1C9H9K50DPKHX';
```
captDaylight: should we make this `pharmacies.photon.ts`?

packages/settings/src/pharmacies.ts
```diff
@@ -0,0 +1,27 @@
+const ALTO_PHARMACY_ID = 'phr_01G9CM93X1NFP1C9H9K50DPKHX';
```
jknapp25, line 1: these are same across envs?

captDaylight: ha writing the same thing, maybe make it `pharmacies.photon.ts`

jknapp25:
need branding for drtelx?

packages/settings/src/pharmacies.ts
captDaylight: yup!

captDaylight (approved):

packages/settings/src/pharmacies.ts
samkotlove: The IDs are the same in all environments

PR #995, opened 2024-03-01
Polishing the item views

jknapp25:
Looks like a lot but it's mostly shuffling. Primarily work toward [consistency between our item views](https://photonhealth.slack.com/archives/C0302SQME1K/p1709242595632899) as well as some low hanging fruit like...

- show address a mail order was sent to ([ticket](https://www.notion.so/photons/Show-address-an-order-was-sent-to-on-order-page-a30c1799155c43eba71cad55fbd15672))
- updated select pharmacy ui on order page - hopefully this deters providers from immediately selecting pharm for a patient
- datadog button tracking on actions - to see what's actually being used on these views


apps/app/src/views/components/CopyText.tsx
```diff
@@ -11,7 +11,7 @@ const CopyText = ({ text, size = 'md' }: { text: string; size?: 'xs' | 'sm' | 'm
         whiteSpace={{ base: 'nowrap', sm: 'normal' }}
         overflow={{ base: 'hidden', sm: 'visible' }}
         textOverflow={{ base: 'ellipsis', sm: 'clip' }}
-        fontWeight="medium"
+        fontWeight={size === 'lg' ? 'medium' : undefined}
```
captDaylight, line 14: what is this for? I don't see us using `CopyText` size `lg` anywhere

jknapp25: In a previous PR i started using medium weight copytext for the headers, that made it different weight if it was used below (pharmacy id for example), i added this so it could go back to normal weight below. Quirky, but felt like passing weight in was a bit much.

PR #1006, opened 2024-03-05
Pharm change request

captDaylight:


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -96,6 +96,15 @@ export const Pharmacy = () => {
     setShowingAllPharmacies(false);
   };
 
+  const showToastError = () =>
+    toast({
+      title: isReroute ? 'Unable to reroute order' : 'Unable to submit pharmacy selection',
+      description: isReroute
+        ? 'Your prescription has already been received. Text us if you need to update your pharmacy.'
```
captDaylight, line 103: @jknapp25 @thebadprince could y'all look at this copy?

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -382,28 +391,19 @@ export const Pharmacy = () => {
...
-        description: 'Please refresh and try again',
-        ...TOAST_CONFIG.ERROR
-      });
-
+      showToastError();
       setSubmitting(false);
-
       console.error(JSON.stringify(error, undefined, 2));
     }
+    navigate(`/status?orderId=${order.id}&token=${token}`);
```
captDaylight, line 406: in this scenario, maybe something else has happened since they tried changing so they no longer can make a change. We should kick them out of this view when we show them the error toast

apps/patient/src/views/Pharmacy.tsx
thebadprince: Title: `Unable to change pharmacies`
description: `Your order is already being processed. Text us if you need it sent to a different pharmacy.`

I think this makes it vague enough (we're not saying their order is confirmed) but still more specific as to why they can't change pharmacies (order is being processed). There's also no internal lingo of "reroute" so it's more straight to the point. Lmk if that makes sense 🙏 

ps. Curious: what does the toast look like/how does it currently behave?

captDaylight: Screen Shot 2024-03-06 at 5 29 44 PM

thebadprince: tiny nit: i think we can use chakra's warning toast so this is orange instead of red. i prefer to use error if the action is destructive or there's a system error

jknapp25 (approved):

PR #1007, opened 2024-03-05
Ready by updates

jknapp25:
- update `Select pharmacy` cta to `Next`
- navigate to pharmacy selection without delay

### DIFF
--- a/apps/patient/src/views/Pharmacy.tsx
+++ b/apps/patient/src/views/Pharmacy.tsx
@@ -41,6 +41,10 @@ import capsulePharmacyIdLookup from '../data/capsulePharmacyIds.json';
 import { Pharmacy as EnrichedPharmacy } from '../utils/models';
 import costcoLogo from '../assets/costco_small.png';
 import { isGLP } from '../utils/isGLP';
+import dayjs from 'dayjs';
+import timezone from 'dayjs/plugin/timezone';
+
+dayjs.extend(timezone);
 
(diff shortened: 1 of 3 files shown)


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -41,6 +41,10 @@ import capsulePharmacyIdLookup from '../data/capsulePharmacyIds.json';
 import { Pharmacy as EnrichedPharmacy } from '../utils/models';
 import costcoLogo from '../assets/costco_small.png';
 import { isGLP } from '../utils/isGLP';
+import dayjs from 'dayjs';
```
captDaylight, line 44: is dayjs being used in this file?

apps/patient/src/views/ReadyBy.tsx
```diff
@@ -85,17 +72,10 @@ export const ReadyBy = () => {
 
     setOrder({
       ...order,
-      readyBy,
-      readyByTime
```
captDaylight, line 89: Why is readyByTime being removed here but used in the datadog action?

jknapp25: Oh! Good catch. Was thinking i don't use it later, but obviously still need to save it lol.

jknapp25: added back

captDaylight (approved):

PR #1021, opened 2024-03-12
Fix, Updating user roles not working

captDaylight:
https://www.notion.so/photons/Updating-user-roles-not-working-6ac08cda4c0f49e1baaa9c0ff20fd491?pvs=4

### DIFF
--- a/apps/app/src/views/routes/Settings/components/users/EditRolesAction.tsx
+++ b/apps/app/src/views/routes/Settings/components/users/EditRolesAction.tsx
@@ -94,11 +94,7 @@ const roleSchema = yup
             street1: yup.string().required('Address is required'),
             street2: yup.string(),
             city: yup.string().required('City is required'),
-            state: yupStateSchema.when('roles', {
-              is: (roles: { value: string; label: string }[]) => hasPrescriberRole(roles),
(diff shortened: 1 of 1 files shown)


apps/app/src/views/routes/Settings/components/users/EditRolesAction.tsx
```diff
@@ -94,11 +94,7 @@ const roleSchema = yup
             street1: yup.string().required('Address is required'),
             street2: yup.string(),
             city: yup.string().required('City is required'),
-            state: yupStateSchema.when('roles', {
-              is: (roles: { value: string; label: string }[]) => hasPrescriberRole(roles),
-              then: yup.string().required('Please enter a valid state'),
-              otherwise: yup.string()
-            }),
+            state: yupStateSchema,
```
captDaylight, line 97: @photonsFardad I can't remember why this was changed:
https://github.com/Photon-Health/client/pull/953

photonsFardad: I dont exactly remember. I think it had something to do with state is required when updating when its a prescriber role

photonsFardad (approved):

PR #1023, opened 2024-03-13
Autoselect "Open now" filter if rerouting after your pharmacy is closed

captDaylight:
If the pharmacy is closed we'll now send an SMS to let them know they can reroute if they need it sooner. This is a small efficiency to auto select the open now filter in the pharmacy search list.

### DIFF
--- a/apps/patient/src/views/Pharmacy.tsx
+++ b/apps/patient/src/views/Pharmacy.tsx
@@ -55,9 +55,9 @@ export const Pharmacy = () => {
   const [searchParams] = useSearchParams();
   const token = searchParams.get('token');
   const isReroute = searchParams.get('reroute');
+  const pharmacyClosed = searchParams.get('pharmacyClosed');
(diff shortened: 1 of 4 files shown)


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -55,9 +55,9 @@ export const Pharmacy = () => {
   const [searchParams] = useSearchParams();
   const token = searchParams.get('token');
   const isReroute = searchParams.get('reroute');
+  const pharmacyClosed = searchParams.get('pharmacyClosed');
```
jknapp25, line 58: what if we flip it and make it `openNow`? that way it's applicable to the filters and later we could do something similar with `open24Hours` if we want to navigate directly to this view and auto-enable filters

captDaylight: I like that

jknapp25 (approved):

PR #1025, opened 2024-03-14
Fix Geocoder race condition with standalone elements

captDaylight:
https://photonhealth.slack.com/archives/C05UCJ4FZAN/p1710433874075989

### DIFF
--- a/packages/components/src/systems/LocationSelect/index.tsx
+++ b/packages/components/src/systems/LocationSelect/index.tsx
@@ -9,6 +9,7 @@ import loadGoogleScript from '../../utils/loadGoogleScript';
 import getLocations, { Location } from '../../utils/getLocations';
 import autocompleteLocation from '../../utils/autocompleteLocation';
 import ComboBox from '../../particles/ComboBox';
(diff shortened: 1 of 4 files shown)


jknapp25 (approved):

packages/components/src/systems/LocationSelect/index.tsx
```diff
@@ -49,7 +52,8 @@ export default function LocationSelect(props: LocationSelectProps) {
       const {
         coords: { latitude, longitude }
       } = await getNavigatorLocation({ timeout: 5000 });
-      const locations = await getLocations({ latitude, longitude }, geocoder!);
+      await asyncInterval(() => !!geocoder(), 10, 20);
+      const locations = await getLocations({ latitude, longitude }, geocoder()!);
```
samkotlove, line 56: should (can?) we do the check for geocoder within getLocations?

captDaylight: I have it as a param on getLocations so that it's not creating an instance of the geocoder each time it's called, could be a broader solution for this in the future though.

PR #1037, opened 2024-03-18
Remove combine modal error

jknapp25:
This addresses the Blueberry comment [here](https://blueberrymed.slack.com/docs/T6AM2J4UD/F06QGB4328G?focus_section_id=temp:C:RUT638e4be724cd48db9b62ec036): `Error events are emitted with empty errors arrays.`

We are sending event messages, tho this situation we are emitting an event without details. 

I'd opt for removing this error since there technically isn't one, we're just letting them know that they have another order that this can be combined with. And I think the modal gives that context well.

### DIFF
--- a/packages/elements/package.json
+++ b/packages/elements/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@photonhealth/elements",
-  "version": "0.8.2",
+  "version": "0.8.3",
   "description": "",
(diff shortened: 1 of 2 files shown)


captDaylight (approved):
mazzive change

captDaylight:
Could you bump the patch number on the elements package before merging?

PR #1048, opened 2024-03-20
Med Ops can see the Prescribe form but not submit

captDaylight:
- `PhotonAuthorized` is now only looking for `read:patient` requirement. Developers can't see patients so the form won't work for them
- Added a new `Alert` component

https://www.notion.so/photons/Users-should-be-able-to-see-prescribe-flow-even-if-they-re-not-a-provider-56ac65efd259448781ab1182359e1914?pvs=4

If not med-ops or prescriber you'll see:
Screen Shot 2024-03-20 at 2 57 58 PM

If you're med-ops you'll see:
Screen Shot 2024-03-20 at 2 59 02 PM


jknapp25:
nit: contrast seems a little low, maybe a border would help?

image

jknapp25 (approved):
Very nice!

captDaylight:
@jknapp25 added the border, looks better :)

PR #1052, opened 2024-03-21
Fix Reroute Error + Can reroute on First Pharmacy Selection

captDaylight:
1. There was an error introduced with the reroute failed error that is fixed here
2. I simplified the logic about how the status page knows if an order reroute was just attempted, it also allows now for a reroute after the first selection


jknapp25 (approved):
luv it

PR #1053, opened 2024-03-21
Separate readyBy options by today/tomorrow

jknapp25:
In order to level out readyBy selected times across the day, we're trying out a full day of time options for tomorrow instead of assuming tomorrow means 9AM the next day.

https://github.com/Photon-Health/client/assets/3934326/74336358-3927-4efb-8964-c763e94fd3f5


apps/patient/src/views/ReadyBy.tsx
```diff
@@ -82,90 +88,137 @@ export const ReadyBy = () => {
...
-                              <Text color="gray.500" fontSize="sm">
-                                {option.description}
-                              </Text>
-                            )
-                          ) : null}
-                        </VStack>
-                      </HStack>
-                    </CardBody>
-                  </Card>
+                  <Fade key={activeTab + '-' + option.label} in={true}>
```
jknapp25, line 154: SlideFade would be ideal here, but that was a pain to get working. Give it a regular fade to make it clear that you changed tabs.

apps/patient/src/utils/general.ts
```diff
@@ -151,7 +151,7 @@ export const preparePharmacyHours = (pharmacy: types.Pharmacy): EnrichedPharmacy
   };
 };
 
-export const convertReadyByToUTCTimestamp = (readyBy: string): string => {
+export const convertReadyByToUTCTimestamp = (readyBy: string, readyByDay: string): string => {
```
captDaylight, line 154: could you make a todo comment to make these params typed? so something like `if (readyByDay === 'tomorrow') {` is more type safe

captDaylight (approved):

PR #1055, opened 2024-03-22
Test Rerouting on Neutron, setting enablePatientRerouting to true

captDaylight:


samkotlove (approved):
Are we going to swap the enablePatientrerouting flag default to always be true after testing?

captDaylight:
@samkotlove I think we should, are there any customers who explicitly don't want it?

PR #1069, opened 2024-03-27
Better Order Cancelled Page

captDaylight:
https://www.notion.so/photons/Create-more-robust-Order-Cancelled-Page-for-Patient-App-c6d89f6598914d4f8869e861b86e9596?pvs=4
Screen Shot 2024-03-27 at 3 12 59 PM

### DIFF
--- a/apps/patient/src/components/Prescriptions.tsx
+++ b/apps/patient/src/components/Prescriptions.tsx
@@ -1,0 +1,83 @@
+import {
+  Accordion,
+  AccordionButton,
+  AccordionItem,
+  AccordionPanel,
+  Box,
+  Card,
+  CardBody,
+  Container,
+  HStack,
+  Text,
+  VStack
+} from '@chakra-ui/react';
+import { FaPrescription } from 'react-icons/fa';
+import { text as t } from '../utils/text';
+import { formatDate } from '../utils/general';
+import { useOrderContext } from '../views/Main';
+
+export const Prescriptions = () => {
+  const { flattenedFills } = useOrderContext();
+
+  return (
+    <Container pb={32}>
+      <VStack spacing={4} align="span" pt={5}>
+        <Accordion allowToggle defaultIndex={[0]}>
+          {flattenedFills.map(({ id, treatment, prescription, count }) => (
+            <AccordionItem border="none" mb={3} key={id}>
+              <Card w="full" backgroundColor="white" borderRadius="lg">
+                <CardBody p={0}>
+                  <HStack>
+                    <AccordionButton
+                      p={5}
+                      _expanded={{ bg: 'transparent' }}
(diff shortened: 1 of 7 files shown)


apps/patient/src/utils/text.tsx
```diff
@@ -32,6 +32,12 @@ export const text = {
   open: 'Open',
   open24hrs: 'Open 24 Hours',
   orderCanceled: 'This order was canceled.',
```
jknapp25, line 34: still used?

apps/patient/src/utils/text.tsx
```diff
@@ -32,6 +32,12 @@ export const text = {
...
+  orderCanceledHeader: 'Your order was canceled',
```
jknapp25, line 35: i tend to like to exclude denoting elements like "header", "subheader" since I might use the text in multiple places, just preference

apps/patient/src/utils/text.tsx
jknapp25: ah, in the tab title 👍🏼

apps/patient/src/components/Prescriptions.tsx
jknapp25, line 1: maybe "PrescriptionsList" to match <entity><element> naming convention?

jknapp25:
Patient name on white background or gray? think we're good on text contrast?

image

captDaylight:
Screen Shot 2024-03-27 at 3 34 14 PM
@jknapp25 I'm going off of this one in the task link, @thebadprince ?

apps/patient/src/components/Prescriptions.tsx
captDaylight: Sounds great!

apps/patient/src/views/Canceled.tsx
```diff
@@ -1,17 +1,43 @@
...
+            </Heading>
+            <Box>
+              <Text display="inline">{t.orderCanceledSubheader}</Text>
+            </Box>
+          </VStack>
+        </Container>
+      </Box>
+
+      <Container>
+        <Text align="left" mt={4} fontSize="md" color="gray.400">
```
jknapp25, line 32: gray.500?

jknapp25:
@captDaylight that looks good to me, maybe gray.500 for "Patient" then?

apps/patient/src/utils/text.tsx
captDaylight: yup makes sense, updating

jknapp25 (approved):

apps/patient/src/views/Canceled.tsx
captDaylight: you got it my guy

captDaylight:
retested, works

PR #1072, opened 2024-03-28
Fix gap above today/tomorrow section on small screens

jknapp25:
Noticed in sessions that on prod where the blue "This is not a real prescription" banner shows and the readyBy options extend past the bottom of the page because it's an extra small screen (iPhone SE), that there's a gap above the sticky today/tomorrow section.

Removing this gap...
image


captDaylight (approved):

captDaylight:
I mean real slick would be to do somethnig like https://photonhealth.slack.com/archives/C05LVD54ZC0/p1711649787473229?thread_ts=1710790007.564709&cid=C05LVD54ZC0

PR #1080, opened 2024-04-03
Walgreens urgency pilot

jknapp25:
To test Walgreens urgent fast-pass we're floating any walgreens in a 25 mile radius up to the top of the list. This is specific to Blueberry at first because of pediatric urgency.

Example (sans blueberry):

image


apps/patient/src/components/PharmacyCard.tsx
```diff
@@ -91,8 +91,6 @@ const DistanceAddress = ({ distance, address }: DistanceAddressProps) => {
 interface PharmacyCardProps {
   pharmacy: EnrichedPharmacy;
   preferred?: boolean;
-  previous?: boolean;
```
jknapp25, line 94: unused tags

apps/patient/src/components/PharmacyCard.tsx
```diff
@@ -142,27 +138,15 @@ export const PharmacyCard = memo(function PharmacyCard({
...
+            {pharmacy?.isUrgent ? (
+              <Tag size="sm" bgColor="yellow.200">
+                <TagLabel>Ready in 30 minutes</TagLabel>
               </Tag>
             ) : null}
           </HStack>
           <VStack align="start" w="full" spacing={0}>
             <HStack spacing={2}>
-              {pharmacy?.logo ? <Image src={pharmacy.logo} width="auto" height="24px" /> : null}
+              {pharmacy?.logo ? <Image src={pharmacy.logo} width="auto" height="20px" /> : null}
```
jknapp25, line 149: Costco logo has white border, wags doesn't so it looked quite big. Double-checked, costco still looks good.

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -197,16 +198,43 @@ export const Pharmacy = () => {
...
         }
       }
     } catch {
       // no costcos found :(
       pharmaciesResult = [];
     }
 
+    // check if top ranked walgreens is in the area
+    try {
+      if (order.organization.name === 'Blueberry Pediatrics' && order.readyBy === 'Urgent') {
```
captDaylight, line 214: I think you could release this before Monday, in the costco case in the same file I flag it in the orgSettings:

``` ts
const enableTopRankedCostco = !isDemo && orgSettings.topRankedCostco;
// ... 
if (enableTopRankedCostco && containsGLP) {
```

You would then updated the org setting on monday to released it to Blueberry.

I could be convinced of your approach for flagging a customer too, but it would be good if they were consistent.

jknapp25: oh yeah, way better, I'll do that

packages/settings/src/lib/photon.ts
```diff
@@ -218,7 +219,8 @@ const organizationSettings: {
     ...defaultSettings,
     logo: 'blueberry_logo.png',
     accentColor: '#235AFF',
-    enableMedHistory: true
+    enableMedHistory: true,
+    topRankedWalgreens: true
```
captDaylight: you don't want to set this now right? you'll have a PR for this on Monday?

captDaylight: hmm didn't mean to make this a "review"

captDaylight (approved):

packages/settings/src/lib/photon.ts
jknapp25: removed

PR #1103, opened 2024-04-09
Filter pharmacy option if already in the list

jknapp25:
For top ranked pharmacies (Costco, Walgreens) we were potentially showing them twice in the list because they are two separate query results merged together. This dedupes.

### DIFF
--- a/apps/patient/src/utils/general.ts
+++ b/apps/patient/src/utils/general.ts
@@ -107,7 +107,7 @@ function isCloseEvent(event: types.PharmacyEvent): event is types.PharmacyCloseE
 }
 
 /**
- * Dedupes pharmacy options when merging new and existing pharmacies
+ * Dedupes and enriches pharmacy options when merging new and existing pharmacies
  * @param newPharmacies Pharmacies query result
  * @param existingPharmacies Pre-loaded pharmacies
  * @returns Deduped list of enriched pharmacy options
--- a/apps/patient/src/utils/general.ts
+++ b/apps/patient/src/utils/general.ts
@@ -106,6 +106,23 @@ function isCloseEvent(event: types.PharmacyEvent): event is types.PharmacyCloseE
   return event.type === 'close';
 }
 
+/**
+ * Dedupes pharmacy options when merging new and existing pharmacies
+ * @param newPharmacies Pharmacies query result
+ * @param existingPharmacies Pre-loaded pharmacies
+ * @returns Deduped list of enriched pharmacy options
+ */
+export const preparePharmacyOptions = (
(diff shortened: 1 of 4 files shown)


apps/patient/src/utils/general.ts
```diff
@@ -106,6 +106,23 @@ function isCloseEvent(event: types.PharmacyEvent): event is types.PharmacyCloseE
   return event.type === 'close';
 }
 
+/**
+ * Dedupes pharmacy options when merging new and existing pharmacies
+ * @param newPharmacies Pharmacies query result
+ * @param existingPharmacies Pre-loaded pharmacies
+ * @returns Deduped list of enriched pharmacy options
+ */
+export const preparePharmacyOptions = (
```
samkotlove, line 115: feel like `dedupePharmacyOptions` is a better name?

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -267,7 +265,10 @@ export const Pharmacy = () => {
       return;
     }
 
-    const preparedPharmacies: PharmacyWithHours[] = pharmaciesResult.map(preparePharmacy);
+    const preparedPharmacies: PharmacyWithHours[] = preparePharmacyOptions(
+      pharmaciesResult,
+      topRankedPharmacies
+    );
```
samkotlove, line 271: is it the top ranked pharmacies would be duplicated?

apps/patient/src/utils/general.ts
```diff
@@ -106,6 +106,23 @@ function isCloseEvent(event: types.PharmacyEvent): event is types.PharmacyCloseE
...
+  newPharmacies: types.Pharmacy[],
+  existingPharmacies: EnrichedPharmacy[] = []
+) => {
+  const existingPharmacyIds = new Set(existingPharmacies.map((p) => p.id));
+  const enrichedNewPharmacies = newPharmacies
+    .filter((p) => !existingPharmacyIds.has(p.id))
+    .map(preparePharmacy);
+  return [...existingPharmacies, ...enrichedNewPharmacies];
+};
```
samkotlove, line 124: could we get a unit test on this func?

apps/patient/src/utils/general.ts
jknapp25: It also processes hours info

apps/patient/src/views/Pharmacy.tsx
jknapp25: yes

marisbest2:
Though we should copy over the tests

PR #1108, opened 2024-04-15
Organize RX history in chronological order

captDaylight:
https://www.notion.so/photons/Organize-RX-history-in-chronological-order-17a04d5b149e472ead5d4cbded88c87c?pvs=4

UPDATE ⬇️ 

Apr-15-2024 17-01-27


packages/components/src/systems/PatientMedHistory/index.tsx
```diff
@@ -42,6 +43,25 @@ type PatientMedHistoryProps = {
...
+  active: boolean;
+  prescription: {
+    id: string;
+    writtenAt: string;
+  };
+  medication: {
+    id: string;
+    name: string;
+  };
+};
```
marisbest2, line 56: Why is this defined here and not coming from the SDK?

packages/components/src/systems/PatientMedHistory/index.tsx
```diff
@@ -42,6 +43,25 @@ type PatientMedHistoryProps = {
   newMedication?: Medication | SearchMedication;
 };
 
+type PatientMedication = {
+  active: boolean;
+  prescription: {
```
marisbest2, line 48: Schema says this is optional https://github.com/Photon-Health/terraform/blob/main/modules/graphql-api/appsync.graphql#L86

packages/components/src/systems/PatientMedHistory/index.tsx
```diff
@@ -42,6 +43,25 @@ type PatientMedHistoryProps = {
...
+  medication: {
+    id: string;
+    name: string;
+  };
+};
+
+type GetPatientResponse = {
+  patient: {
+    id: string;
+    medicationHistory: PatientMedication[];
```
marisbest2, line 61: Its a little bit of a nit, but the schema says this is optional. Idk what we actually return for a patient with no medication history. https://github.com/Photon-Health/terraform/blob/main/modules/graphql-api/appsync.graphql#L108

packages/components/src/systems/PatientMedHistory/index.tsx
```diff
@@ -62,17 +82,34 @@ export default function PatientMedHistory(props: PatientMedHistoryProps) {
...
-      query: GET_PATIENT_MED_HISTORY,
-      variables: { id: props.patientId }
-    });
-    if (data?.patient?.medicationHistory) {
-      setMedHistory(data.patient.medicationHistory);
-    } else {
+  const queryOptions = createMemo(() => ({
+    variables: { id: props.patientId },
+    client: client!.apollo
+  }));
```
marisbest2, line 88: Why separate this out?

jknapp25:
Latest at top or bottom? Having most recent below the fold feels a little less practical? Also, is it possible to init with a column sort on Written as DESC or ASC?

packages/components/src/systems/PatientMedHistory/index.tsx
```diff
@@ -62,17 +82,34 @@ export default function PatientMedHistory(props: PatientMedHistoryProps) {
...
+    if (medicationHistory) {
+      const sortedMedHistory = medicationHistory.slice().sort((a, b) => {
+        const dateA = a?.prescription?.writtenAt
+          ? new Date(a.prescription.writtenAt).getTime()
+          : -Infinity;
+        const dateB = b?.prescription?.writtenAt
+          ? new Date(b.prescription.writtenAt).getTime()
+          : -Infinity;
+        return dateA - dateB;
+      });
```
marisbest2, line 106: Are these date strings or dates already? I think you can just do `.sort()`?

marisbest2 (commented):
A few code comments.

Also agree with Josh, would probably put most recent at top

packages/components/src/systems/PatientMedHistory/index.tsx
captDaylight: because solid lifecycle wouldn't catch the patientId updating which would update the query, this will make that happen

packages/components/src/systems/PatientMedHistory/index.tsx
captDaylight: It's a date string at the moment or possibly undefined

captDaylight:
Updated to start with most recent first 👍

jknapp25 (approved):

marisbest2 (approved):

PR #1118, opened 2024-04-17
Add delivery options for Sana

jknapp25:
Surfacing two delivery options for Sana Care

image

### DIFF
--- a/apps/patient/src/components/BrandedPharmacyCard.tsx
+++ b/apps/patient/src/components/BrandedPharmacyCard.tsx
@@ -1,5 +1,4 @@
 import {
-  Box,
   Card,
   CardBody,
   Image,
@@ -18,14 +17,13 @@ import capsulePharmacyIdLookup from '../data/capsulePharmacyIds.json';
 
 interface Props {
   pharmacyId: string;
-  selectedId: string;
-  handleSelect: (id: string) => void;
+  selectedId?: string;
+  handleSelect?: (id: string) => void;
 }
 
 const PHARMACY_BRANDING = {
   [process.env.REACT_APP_AMAZON_PHARMACY_ID as string]: {
-    logo: amazonPharmacyLogo,
-    description: 'Save time. Save money. Stay healthy.'
+    logo: amazonPharmacyLogo
   },
   [process.env.REACT_APP_ALTO_PHARMACY_ID as string]: {
     logo: altoLogo,
@@ -51,40 +49,24 @@ export const BrandedPharmacyCard = ({ pharmacyId, selectedId, handleSelect }: Pr
   const brand = PHARMACY_BRANDING[pharmacyId];
   if (!brand) return null;
 
-  let tagline = null;
-  if (pharmacyId in capsulePharmacyIdLookup) {
-    const [firstWord, ...restOfSentence] = brand.description.split(' ');
-
-    tagline = (
-      <Text fontSize="sm" display="inline">
-        <Box as="span" color="gray.900" mr={1}>
-          {firstWord}
-        </Box>
-        <Box as="span" color="gray.500">
-          {restOfSentence.join(' ')}
-        </Box>
-      </Text>
-    );
-  } else {
-    tagline = (
-      <Text fontSize="sm" color="gray.500">
-        {brand.description}
-      </Text>
-    );
-  }
+  const tagline = brand.description ? (
+    <Text fontSize="sm" color="gray.500">
+      {brand.description}
+    </Text>
+  ) : null;
 
   return (
     <SlideFade offsetY="60px" in={true} key={`courier-pharmacy-${pharmacyId}`}>
       <Card
         bgColor="white"
         cursor="pointer"
-        onClick={() => handleSelect(pharmacyId)}
+        onClick={() => handleSelect && handleSelect(pharmacyId)}
         border="2px solid"
         borderColor={selectedId === pharmacyId ? 'brand.500' : 'white'}
         mx={isMobile ? -3 : undefined}
       >
         <CardBody p={3}>
-          <VStack align="start" spacing={2}>
+          <VStack align="start" spacing={1}>
             <Image src={brand.logo} width="auto" height="30px" />
             {tagline}
           </VStack>
--- a/apps/patient/src/components/BrandedPharmacyCard.tsx
+++ b/apps/patient/src/components/BrandedPharmacyCard.tsx
@@ -17,8 +17,8 @@ import capsulePharmacyIdLookup from '../data/capsulePharmacyIds.json';
 
 interface Props {
   pharmacyId: string;
-  selectedId?: string;
-  handleSelect?: (id: string) => void;
+  selectedId?: string | undefined;
+  handleSelect?: ((id: string) => void) | undefined;
 }
 
 const PHARMACY_BRANDING = {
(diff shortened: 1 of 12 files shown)


apps/patient/src/components/BrandedPharmacyCard.tsx
```diff
@@ -18,14 +17,13 @@ import capsulePharmacyIdLookup from '../data/capsulePharmacyIds.json';
...
-  selectedId: string;
-  handleSelect: (id: string) => void;
+  selectedId?: string;
+  handleSelect?: (id: string) => void;
 }
 
 const PHARMACY_BRANDING = {
   [process.env.REACT_APP_AMAZON_PHARMACY_ID as string]: {
-    logo: amazonPharmacyLogo,
-    description: 'Save time. Save money. Stay healthy.'
```
marisbest2, line 28: Why'd we drop this?

apps/patient/src/components/BrandedPharmacyCard.tsx
```diff
@@ -18,14 +17,13 @@ import capsulePharmacyIdLookup from '../data/capsulePharmacyIds.json';
 
 interface Props {
   pharmacyId: string;
-  selectedId: string;
-  handleSelect: (id: string) => void;
+  selectedId?: string;
+  handleSelect?: (id: string) => void;
```
marisbest2, line 21: Why is this optional?

apps/patient/src/components/BrandedPharmacyCard.tsx
```diff
@@ -18,14 +17,13 @@ import capsulePharmacyIdLookup from '../data/capsulePharmacyIds.json';
 
 interface Props {
   pharmacyId: string;
-  selectedId: string;
-  handleSelect: (id: string) => void;
+  selectedId?: string;
```
marisbest2, line 20: This should be `string | undefined`. You presumably always want the parent to send it down?

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -104,6 +105,10 @@ export const Pharmacy = () => {
 
   const enableMailOrder = !isDemo && orgSettings.mailOrderNavigate;
 
+  const isCapsuleTerritory =
+    order?.address?.postalCode !== undefined && order.address.postalCode in capsuleZipcodeLookup;
```
marisbest2, line 109: Generally prefer `!= null` to !== undefined`.

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -611,9 +616,18 @@ export const Pharmacy = () => {
...
-              {enableMailOrder ? (
+              {enableCourier || enableMailOrder ? (
                 <BrandedOptions
-                  options={orgSettings.mailOrderNavigateProviders ?? []}
+                  options={[
+                    ...(enableCourier && order?.address?.postalCode
+                      ? [
+                          capsuleZipcodeLookup[
+                            order.address.postalCode as keyof typeof capsuleZipcodeLookup
+                          ]?.pharmacyId
```
marisbest2, line 626: If the postal code is not in the lookup, you might end up with an `undefined` in there. I dont think this is possible though?

marisbest2: I would probably do something like

```ts
const capsulePharmacy = capsuleZipcodeLookup[
                            order.address.postalCode as keyof typeof capsuleZipcodeLookup
                          ]?.pharmacyId;
```
And then 
```tsx
options={[...(enableCourier && order?.address?.postalCode && capsulePharmacy ? [capsulePharmacy] : []) ...)
```

Still kinda ugly but I think a bit better

marisbest2: You can also do:
```ts
const options = [
    enableCourier && order?.address?.postalCode && capsulePharmacy, 
    enableMailOrder && orgSettings.mailOrderNavigateProviders
].filter(x => x)
```

packages/settings/src/lib/photon.ts
```diff
@@ -210,8 +210,9 @@ const organizationSettings: {
     ...defaultSettings,
     logo: 'sana_care_logo.png',
     accentColor: '#FBCC45',
-    mailOrderNavigate: false,
+    mailOrderNavigate: true,
     mailOrderNavigateProviders: [AMAZON_PHARMACY_ID],
+    enableCourierNavigate: true,
```
marisbest2, line 215: Can we do this in a separate PR I think, just in case this doesnt work for some reason and we want to walk it up to photon

apps/patient/src/views/Status.tsx
```diff
@@ -178,6 +185,9 @@ export const Status = () => {
   // Demo pharmacies are already prepared
   const pharmacyWithHours = pharmacy ? (isDemo ? pharmacy : preparePharmacy(pharmacy)) : undefined;
 
+  const isDeliveryPharmacy =
+    fulfillmentType === types.FulfillmentType.MailOrder || fulfillmentType === 'COURIER';
```
marisbest2, line 189: Either use strings for both or use the enum for both

apps/patient/src/components/BrandedPharmacyCard.tsx
jknapp25: just to distinguish a bit more from same day, [chat w/ jomi](https://photonhealth.slack.com/archives/C0302SQME1K/p1713388912122989?thread_ts=1713387121.069459&cid=C0302SQME1K)

apps/patient/src/components/BrandedPharmacyCard.tsx
jknapp25: these aren't used on Status page

apps/patient/src/components/BrandedPharmacyCard.tsx
jknapp25: status page doesn't use the selectedId stuff since the pharmacy is just presentational, so i just made it optional

apps/patient/src/views/Pharmacy.tsx
jknapp25: updated

apps/patient/src/components/BrandedPharmacyCard.tsx
```diff
@@ -51,40 +49,24 @@ export const BrandedPharmacyCard = ({ pharmacyId, selectedId, handleSelect }: Pr
...
+    </Text>
+  ) : null;
 
   return (
     <SlideFade offsetY="60px" in={true} key={`courier-pharmacy-${pharmacyId}`}>
       <Card
         bgColor="white"
         cursor="pointer"
-        onClick={() => handleSelect(pharmacyId)}
+        onClick={() => handleSelect && handleSelect(pharmacyId)}
```
captDaylight, line 63: can also use optional chaining here 
`onClick={() => handleSelect?.(pharmacyId)}`

captDaylight (approved):

apps/patient/src/components/BrandedPharmacyCard.tsx
marisbest2: I had the same thing here lol

apps/patient/src/components/BrandedPharmacyCard.tsx
jknapp25: reverted to original

apps/patient/src/components/BrandedPharmacyCard.tsx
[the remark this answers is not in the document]
jknapp25: did optional chain

PR #1119, opened 2024-04-18
Copy refinements on delivery vs local

jknapp25:
Subheader and branded pharmacy subtext updates.

image

### DIFF
--- a/apps/patient/src/components/PickupOptions.tsx
+++ b/apps/patient/src/components/PickupOptions.tsx
@@ -15,7 +15,7 @@ interface PickupOptionsProps {
   handleSetPreferred: (id: string) => void;
   loadingMore: boolean;
   showingAllPharmacies: boolean;
-  courierEnabled: boolean;
+  showHeading: boolean;
   enableOpenNow: boolean;
   enable24Hr: boolean;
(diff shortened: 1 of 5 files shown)


apps/patient/src/components/PickupOptions.tsx
```diff
@@ -15,7 +15,7 @@ interface PickupOptionsProps {
   handleSetPreferred: (id: string) => void;
   loadingMore: boolean;
   showingAllPharmacies: boolean;
-  courierEnabled: boolean;
+  showHeading: boolean;
```
marisbest2, line 18: Why are we changing from the specific `courierEnabled` to the generic `showHeading`? What if we want a different header in the future?

marisbest2 (approved):

apps/patient/src/components/PickupOptions.tsx
jknapp25: This was a bug. If only mail order was enabled, it would not show the header on pickup options, like this...

image

Also note, we were piping `enableMailOrder` into the `courierEnabled` prop

image

marisbest2: So it should be "showDeliveryOptions" is what you're saying?

PR #1124, opened 2024-04-18
Fix err fetching catalogs in advanced search

jknapp25:
On `/prescriptions/new`, we've been seeing `TypeError: Cannot read properties of null (reading &#x27;catalogs&#x27;)`, this was due to incorrect query being used in med search.

### DIFF
--- a/packages/elements/src/photon-med-search/photon-med-search-component.tsx
+++ b/packages/elements/src/photon-med-search/photon-med-search-component.tsx
@@ -18,11 +18,9 @@ import { MedicationConceptDropdown } from './components/MedicationConceptDropdow
 import { MedicationFilterDropdown } from './components/MedicationFilterDropdown';
 
 const GET_CATALOGS = gql`
-  query GetProducts($medId: String!) {
-    medicationProducts(id: $medId) {
+  query GetCatalogs {
+    catalogs {
       id
-      name
-      controlled
     }
   }
 `;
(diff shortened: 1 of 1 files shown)


packages/elements/src/photon-med-search/photon-med-search-component.tsx
```diff
@@ -77,7 +77,7 @@ customElement(
 
     onMount(async () => {
       const { data } = await client!.apollo.query({ query: GET_CATALOGS });
-      if (data.catalogs.length > 0) {
+      if (data?.catalogs.length > 0) {
```
marisbest2, line 80: Does `onMount` get called more than once?

marisbest2: There are a few possibilities of whats happening here:
- the query occasionally fails. In which case we should investigate why
- this component hits the `onMount` lifecycle multiple times, in which case we should investigate why.
- There might be some other ways

captDaylight: @marisbest2 I'm not seeing multiple onmounts

packages/elements/src/photon-med-search/photon-med-search-component.tsx
```diff
@@ -77,7 +77,7 @@ customElement(
 
     onMount(async () => {
       const { data } = await client!.apollo.query({ query: GET_CATALOGS });
```
captDaylight, line 79: @marisbest2 @jknapp25 

OH it's because `GET_CATALOGS` above looks like:
```
const GET_CATALOGS = gql`
  query GetProducts($medId: String!) {
    medicationProducts(id: $medId) {
      id
      name
      controlled
    }
  }
`;
const GET_PRODUCTS = gql`
  query GetProducts($medId: String!) {
    medicationProducts(id: $medId) {
      id
      name
      controlled
    }
  }
`;
```
which is an exact copy of the one below it.... there isn't a catalogs in the query   :|

PR #1125, opened 2024-04-18
Add github action for e2e test on patient demo

jknapp25:
This is to keep the demo from failing when we're iterating quickly on the patient app.

The DataDog synthetic test is [defined here](https://app.datadoghq.com/synthetics/details/7dw-d5w-7ss?from_ts=1713478936440&to_ts=1713482536440&live=true).

Todo:

- [ ] add app key to github


marisbest2:
Did we add the app key?

PR #1136, opened 2024-04-19
Set pkg version for apps to commit hash

jknapp25:


apps/app/config-overrides.js
```diff
@@ -33,6 +33,24 @@ module.exports = {
       }
     });
 
+    // Add Datadog source map upload code
```
jknapp25, line 36: think i had missed this piece on the clinical side before

apps/app/src/index.tsx
```diff
@@ -22,7 +24,7 @@ datadogRum.init({
   site: 'datadoghq.com',
   service: pkg.name,
   env: process.env.REACT_APP_ENV_NAME,
-  version: pkg.version,
```
captDaylight, line 25: nice, I guess we aren't updating the app version so this wouldn't have been helpful as it was 👍

PR #1141, opened 2024-04-23
Test sourcemap upload without full path

jknapp25:
I was instructed by a DD support agent to try without the full dist path since DD can scan subdirectories


captDaylight (approved):

marisbest2:
I'm skeptical that this is what we want to do, but we can try it. In the monorepo, `patient` and `app` are two totally separate apps, they just happen to live in the same repo. But maybe... 🤷

PR #1144, opened 2024-04-26
Add Walmart med price

jknapp25:
For the Walmart pilot we're surface med prices if the prescribed meds are in a select list. This will only surface a price if all the medications in the order are present in the list.

Example with 2 meds ($4 + $10) in the list:

image


apps/patient/src/components/PharmacyCard.tsx
```diff
@@ -26,6 +26,7 @@ interface PharmacyCardProps {
   onSetPreferred?: () => void;
   selectable?: boolean;
   showDetails?: boolean;
+  price: number;
```
marisbest2, line 29: This should be optional I assume?

apps/patient/src/utils/general.ts
```diff
@@ -207,3 +211,40 @@ export const convertReadyByToUTCTimestamp = (readyBy: string, readyByDay: string
...
+        (!med.form || fill.treatment.name.toLowerCase().includes(med.form)) &&
+        med.daysSupply === fill?.prescription?.daysSupply &&
+        med.dispenseQuantity === fill?.prescription?.dispenseQuantity
+    );
+
+    // If a matching fill is not found, set allFillsMatch to false
+    if (!match) {
+      allFillsMatch = false;
+      break;
+    }
```
marisbest2, line 242: Just return `null` here I think?

apps/patient/src/utils/general.ts
```diff
@@ -207,3 +211,40 @@ export const convertReadyByToUTCTimestamp = (readyBy: string, readyByDay: string
...
+ */
+export const getTotalWalmartOrderPrice = (fills: types.Fill[]): number => {
+  let totalPrice = 0;
+  let allFillsMatch = true;
+
+  for (const fill of fills) {
+    // Check if the fill is in the walmartMedLookup list
+    const match = walmartMedLookup.find(
+      (med) =>
+        new RegExp(`${med.name}.*${med.strength}`, 'i').test(fill.treatment.name) &&
```
marisbest2, line 231: Can we not just do an exact match here? Theoretically worried about partial matches. For example

`Acetaminophen 500mg - Caffeine 60mg - Pyrilamine maleate 15mg` (excedrin)

would match `Acetaminophen 15mg` despite not being the same thing. (obviously a contrived example)

apps/patient/src/utils/general.ts
```diff
@@ -207,3 +211,40 @@ export const convertReadyByToUTCTimestamp = (readyBy: string, readyByDay: string
...
+      allFillsMatch = false;
+      break;
+    }
+
+    // If a matching fill is found, add its price to the total price
+    totalPrice += match.price;
+  }
+
+  // If all fills match, return the total price; otherwise, return 0
+  return allFillsMatch ? totalPrice : 0;
```
marisbest2, line 249: `0` is a valid price. We should return `null` here if we can't figure out a price

apps/patient/src/views/Main.tsx
```diff
@@ -21,6 +25,7 @@ interface OrderContextType {
   setOrder: (order: Order) => void;
   logo: any;
   isDemo: boolean;
+  totalWalmartPrice: number;
```
marisbest2, line 28: this should be `number | null`

apps/patient/src/views/Main.tsx
```diff
@@ -42,6 +47,8 @@ export const Main = () => {
     isDemo ? countFillsAndRemoveDuplicates(demoOrder.fills) : []
   );
 
+  const [totalWalmartPrice, setTotalWalmartPrice] = useState(0);
```
marisbest2, line 50: useState<number | null>(null)

apps/patient/src/components/BrandedPharmacyCard.tsx
```diff
@@ -46,8 +46,8 @@ export const BrandedPharmacyCard = ({ pharmacyId, handleSelect }: Props) => {
   ) : null;
 
   return (
-    <Container>
-      <Box py={2} bgColor="white" cursor="pointer" onClick={() => handleSelect?.(pharmacyId)}>
+    <Container p={3}>
+      <Box bgColor="white" cursor="pointer" onClick={() => handleSelect?.(pharmacyId)}>
```
marisbest2, line 50: What's with the style changes in this PR?

marisbest2 (dismissed):

marisbest2:
Do we want/need a disclaimer on the pricing?

apps/patient/src/components/BrandedPharmacyCard.tsx
jknapp25: inconsistent padding on pickup vs branded pharmacies here

apps/patient/src/utils/general.ts
jknapp25: first pass was keeping just number, but I'll update

apps/patient/src/utils/general.ts
captDaylight: I think that might be hard to ensure an exact match and the example might be an edge case, IMO this should be good to go because this is at max a 2 or 3 week test that will affect 200 - 400 orders. I think getting this out today is more the higher priority since we have an upcoming meeting with Walmart, hoping to show some positive results

captDaylight (approved):

PR #1146, opened 2024-04-30
Two Auth0 Providers

captDaylight:
We have an issue where a customer is has an [Auth0 instance that's in conflict with our Element's instance](https://photonhealth.slack.com/archives/C06DCU1DTMJ/p1713253957783859). What's causing the issue is that both instances are picking up on the redirect url's params. So to solve this I've added a hardcoded param to our elements that won't do any checks if it's coming from the parent.

Any customer that is using their own Auth0 login will need to set `skipRedirectCallback` to look for the presence of `photon=true`:

``` tsx
    <Auth0Provider
      domain="customer-auth-domain"
      clientId="customer-auth-client-id"
      authorizationParams={{
        redirect_uri: `${window.location.origin}`
      }}
      skipRedirectCallback={window.location.href.includes('photon=true')}
    >
```

Then, when as high up in the customer's authenticated tree as possible, the'll place our `photon-client`:

``` tsx
      {!isAuthenticated ? (
        <LoginButton />
      ) : (
        <photon-client
          ...
          auto-login="true"
          redirect-uri={window.location.origin} 
        >
          {/* ETC */}
        </photon-client>
     }
```

🎥 Demo Video
https://www.loom.com/share/58814af997ea4367a705d253ee68a9c3

👀 Demo Branch
https://github.com/Photon-Health/client/tree/auth0
If you go here to `apps/auth` you can install and then `npm run dev` to recreate the scenario I'm showing in the video above.

### References

https://github.com/auth0/auth0-react/pull/416
This is the PR that allows for multiple contexts. Since we aren't using React, we don't need the context on the parent React component. However in the code there [was a hint to the answer.](https://github.com/auth0/auth0-react/blob/93d54c3863f95ca3f0f3ae51d6a859dbe6463f4b/src/auth0-provider.tsx#L176)

https://github.com/auth0-samples/auth0-link-accounts-sample/tree/react-variant/React
Here's a demo of the above PR

### DIFF
--- a/packages/elements/package.json
+++ b/packages/elements/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@photonhealth/elements",
-  "version": "0.9.2",
+  "version": "0.9.1-rc.1",
   "description": "",
   "main": "dist/index.js",
   "types": "./dist/index.d.ts",
--- a/packages/elements/package.json
+++ b/packages/elements/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@photonhealth/elements",
-  "version": "0.9.1",
+  "version": "0.9.2",
   "description": "",
   "main": "dist/index.js",
   "types": "./dist/index.d.ts",
--- a/packages/elements/src/utils.ts
+++ b/packages/elements/src/utils.ts
@@ -14,12 +14,12 @@ export const validateProps = (props: Record<string, any>, required: string[]) =>
 const CODE_RE = /[?&]code=[^&]+/;
 const STATE_RE = /[?&]state=[^&]+/;
 const ERROR_RE = /[?&]error=[^&]+/;
-const PARENT_RE = /[?&]photon-parent-auth=[^&]+/; // if this is present, these are auth params for a different Auth0 instance
+const IS_PHOTON_RE = /[?&]photon=true[^&]+/; // if this is present, these are auth params for a different Auth0 instance
 
 export const hasAuthParams = (searchParams = window.location.search): boolean =>
   (CODE_RE.test(searchParams) || ERROR_RE.test(searchParams)) &&
   STATE_RE.test(searchParams) &&
-  !PARENT_RE.test(searchParams);
+  IS_PHOTON_RE.test(searchParams);
 
 export const toTitleCase = (str: string) => {
   return str.replace(/\w\S*/g, function (txt) {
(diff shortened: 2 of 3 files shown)


jknapp25 (approved):
LGTM!

packages/elements/src/utils.ts
```diff
@@ -14,9 +14,12 @@ export const validateProps = (props: Record<string, any>, required: string[]) =>
 const CODE_RE = /[?&]code=[^&]+/;
 const STATE_RE = /[?&]state=[^&]+/;
 const ERROR_RE = /[?&]error=[^&]+/;
+const PARENT_RE = /[?&]photon-parent-auth=[^&]+/; // if this is present, these are auth params for a different Auth0 instance
```
samkotlove, line 17: So we are requiring our customer to add `photon-parent-auth`, can't we just do the inverse and add a `photon=true` flag so our customers don't have to do anything?

captDaylight: @sam So If I understand correctly:
- we programmatically append `photon=true` to the redirect uri passed into the `<photon-client>`
- in Photon client, we would check for the presence of `photon=true`
- on their side they would still need to add `skipRedirectCallback={window.location.href.includes('photon=true')}`

packages/elements/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@photonhealth/elements",
-  "version": "0.9.1",
+  "version": "0.9.2",
```
marisbest2, line 3: This should be a minor bump

packages/elements/src/utils.ts
```diff
@@ -14,9 +14,12 @@ export const validateProps = (props: Record<string, any>, required: string[]) =>
 const CODE_RE = /[?&]code=[^&]+/;
 const STATE_RE = /[?&]state=[^&]+/;
 const ERROR_RE = /[?&]error=[^&]+/;
+const IS_PHOTON_RE = /[?&]photon=true[^&]+/; // if this is present, these are auth params for a different Auth0 instance
```
marisbest2, line 17: Why do we use a regex rather than parsing the search params?

packages/elements/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@photonhealth/elements",
-  "version": "0.9.1",
+  "version": "0.9.1-rc.1",
```
captDaylight, line 3: @samkotlove candidate

PR #1149, opened 2024-05-03
Chakra link styling fix

captDaylight:
Link used to spread the whole distance of the parent:
https://www.notion.so/photons/Chakra-link-style-is-messed-up-on-Patient-Page-cce51b87e691408d9aaa174a8e5c41d6?pvs=4

May-03-2024 10-46-30

Also, `TicketModal` form was giving me this error:
```
Type '{ children: Element; onSubmit: (e?: FormEvent<HTMLFormElement> | undefined) => void; style: { width: string; }; }' is missing the following properties from type 'Pick<DetailedHTMLProps<FormHTMLAttributes<HTMLFormElement>, HTMLFormElement>, "children" | "id" | ... 260 more ... | "key">': placeholder, onPointerEnterCapture, onPointerLeaveCapture
``` 
so I modified it with some undefined values...

### DIFF
--- a/apps/app/src/views/components/TicketModal.tsx
+++ b/apps/app/src/views/components/TicketModal.tsx
@@ -124,7 +124,13 @@ export const TicketModal = ({
             >
               {({ handleSubmit, errors, touched, setFieldTouched }) => {
                 return (
-                  <Form onSubmit={handleSubmit} style={{ width: '100%' }}>
+                  <Form
+                    onSubmit={handleSubmit}
+                    style={{ width: '100%' }}
+                    placeholder={undefined}
+                    onPointerEnterCapture={undefined}
+                    onPointerLeaveCapture={undefined}
+                  >
                     <VStack spacing={4} width="100%">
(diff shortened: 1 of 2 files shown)


apps/app/src/views/components/TicketModal.tsx
```diff
@@ -124,7 +124,13 @@ export const TicketModal = ({
             >
               {({ handleSubmit, errors, touched, setFieldTouched }) => {
                 return (
-                  <Form onSubmit={handleSubmit} style={{ width: '100%' }}>
+                  <Form
+                    onSubmit={handleSubmit}
+                    style={{ width: '100%' }}
```
jknapp25, line 129: `w="full"` should work

jknapp25 (approved):

apps/app/src/views/components/TicketModal.tsx
captDaylight: ah this is actually a `formik` form

PR #1187, opened 2024-05-23
Fix unnecessary page reload on session success

samkotlove:


packages/elements/src/store.ts
```diff
@@ -15,7 +15,11 @@ import { GraphQLError } from 'graphql';
 import jwtDecode from 'jwt-decode';
 
 const defaultOnRedirectCallback = (appState?: any): void => {
-  window.location.replace(appState?.returnTo || window.location.pathname);
```
samkotlove, line 18: This was forcing a page load every time, whcih was necessary because the session wasn't available on the elements yet, which is why i check session in the file below

packages/elements/src/store.ts
```diff
@@ -15,7 +15,11 @@ import { GraphQLError } from 'graphql';
...
+  if (appState?.returnTo) {
+    window.location.replace(appState?.returnTo);
+  } else {
+    window.history.pushState(null, window.document.title, window.location.pathname);
```
samkotlove, line 21: instead just drop the query params we added without reloading

packages/sdk/src/auth.ts
```diff
@@ -217,7 +217,9 @@ export class AuthManager {
    */
   public async handleRedirect(url?: string): Promise<RedirectLoginResult<any> | undefined> {
     try {
-      return this.authentication.handleRedirectCallback(url);
+      const result = await this.authentication.handleRedirectCallback(url);
+      await this.authentication.checkSession();
```
samkotlove, line 221: this hydrates the auth0client with the session token so the elements can render.

packages/elements/src/store.ts
```diff
@@ -209,6 +213,7 @@ export class PhotonClientStore {
         try {
           const result = await this.sdk.authentication.handleRedirect(url);
           defaultOnRedirectCallback(result?.appState);
+          await this.authentication.checkSession();
```
captDaylight, line 216: Does this one need to be there too?

packages/sdk/src/auth.ts
captDaylight: Wouldn't this need to check session before handling the redirect? or I am maybe not fully understanding the flow

samkotlove: Not quite, `handleRedirectCallback` does the work to grab the code from the url, validate, and set the cookie

samkotlove: Removed this one. The above explaination still holds, but this one was unnecesssary

captDaylight:
can you add a release candidate on element version?

captDaylight (approved):

PR #1201, opened 2024-05-29
Add order readiness info to status page

jknapp25:
Now we'll show:
- `readyBy` time before received
- "estimated ready time" after it has been marked received

image


packages/sdk/src/types.ts
```diff
@@ -428,6 +428,8 @@ export type OrderFulfillment = {
   state: FulfillmentState;
   trackingNumber?: Maybe<Scalars['String']>;
   type: FulfillmentType;
+  minutesUntilReady?: Maybe<Scalars['Int']>;
+  createdAt?: Scalars['AWSDateTime'];
```
marisbest2, line 432: This isn't optional right?

packages/sdk/src/types.ts
```diff
@@ -428,6 +428,8 @@ export type OrderFulfillment = {
   state: FulfillmentState;
   trackingNumber?: Maybe<Scalars['String']>;
   type: FulfillmentType;
+  minutesUntilReady?: Maybe<Scalars['Int']>;
```
marisbest2, line 431: I think you need readyBy etc here too right?

apps/patient/src/views/Status.tsx
```diff
@@ -15,6 +26,67 @@ import { formatAddress, getFulfillmentType, preparePharmacy } from '../utils/gen
 import { orderStateMapping as m, text as t } from '../utils/text';
 import { useOrderContext } from './Main';
 import { HorizontalStatusStepper } from '../components/HorizontalStatusStepper';
+import dayjs from 'dayjs';
+
+function formatReadyTime(baseTime: string, minutesUntilReady: number): ReactElement {
```
marisbest2, line 31: This should be a react component not a function. 

Use like 
```
<ReadyForPickup baseTime={createdAt} minutesUntilReady={minutesUntilReady} />
```

As currently set up, it'll rerender every time, whereas if its a component, React will only render it when props change.

apps/patient/src/views/Status.tsx
```diff
@@ -15,6 +26,67 @@ import { formatAddress, getFulfillmentType, preparePharmacy } from '../utils/gen
...
+    <Text>
+      Ready{' '}
+      <b>
+        {dayPrefix} at {readyTime.format(timeFormat)}
+      </b>
+    </Text>
+  );
+}
+
+function formatReadyBy(readyBy: string, readyByDay: string): ReactElement {
```
marisbest2, line 59: Same here, should be a component

apps/patient/src/views/Status.tsx
marisbest2, line 31: The naming here really is bad. This returns the **confirmed** ready time. But we have a different field called `readyByTime` which is the **desired** pickup time.

We should fix the naming on both ends probs, but in the meantime this should probably be `<ReadyForPickupTime>` or something like that.

apps/patient/src/graphql/queries.ts
```diff
@@ -6,6 +6,9 @@ export const GET_ORDER = gql`
       id
       state
       isReroutable
+      readyBy
+      readyByDay
+      readyByTime
```
marisbest2, line 11: Are we using `readyByTime`?

apps/patient/src/graphql/queries.ts
```diff
@@ -22,6 +25,8 @@ export const GET_ORDER = gql`
         type
         state
         trackingNumber
+        minutesUntilReady
+        createdAt
```
marisbest2, line 29: Is `minutesUntilReady` actually relative to `createdAt`? It looks to me from reading the code that `minutesUntilReady` is relative to that being set whereas `createdAt` happens when the order is sent?

marisbest2 (changes requested):
A few changes and also a big question - Is `minutesUntilReady` actually relative to `createdAt`?

packages/sdk/src/types.ts
jknapp25: it's on the original order already

apps/patient/src/graphql/queries.ts
jknapp25: no, good catch

apps/patient/src/views/Status.tsx
marisbest2: pretty much, as a function, this will rerender every time because it won't be part of the shadow DOM. The code isn't bad performing, but its not ideal.

If its a component, the code WILL be part of the shadow DOM, so it'll only rerender when the parent rerenders (never) or the props change (also never).

apps/patient/src/views/Status.tsx
```diff
@@ -15,6 +26,67 @@ import { formatAddress, getFulfillmentType, preparePharmacy } from '../utils/gen
 import { orderStateMapping as m, text as t } from '../utils/text';
 import { useOrderContext } from './Main';
 import { HorizontalStatusStepper } from '../components/HorizontalStatusStepper';
+import dayjs from 'dayjs';
+
+function formatReadyTime(baseTime: string, minutesUntilReady: number): ReactElement {
+  const currentTime = dayjs(baseTime);
```
marisbest2, line 32: This naming is confusing to me also. I keep reading `currentTime` as "now" when its really the `baseTime` just as a `dayjs`.

jknapp25: yeah this went thru a few revisions, goods points, I'll address

apps/patient/src/components/ReadyText.tsx
```diff
@@ -0,0 +1,106 @@
+import dayjs, { Dayjs } from 'dayjs';
+import { Text } from '@chakra-ui/react';
+import { OrderFulfillment } from 'packages/sdk/dist/types';
+
+export type Maybe<T> = T | null;
```
marisbest2, line 5: Presumably this exists elsewhere (e.g. in `packages/sdk/dist/types`). `Maybe<T>` should also be `T | null | undefined`

apps/patient/src/components/ReadyText.tsx
```diff
@@ -0,0 +1,106 @@
...
+    <>
+      {showPatientDesiredreadyBy && (
+        <PatientDesiredReadyBy readyBy={readyBy} readyByDay={readyByDay} />
+      )}
+      {showEstimatedReadyAt && (
+        <PharmacyEstimatedReadyAt pharmacyEstimatedReadyAt={fulfillment.pharmacyEstimatedReadyAt} />
+      )}
+    </>
+  );
+};
```
marisbest2, line 35: Its hard to parse this. I _think_ this is:
- If deliveryPharmacy = do nothing
- If SENT = show desired
- If RECEIVED = show estimated ready

Is that right?
If yes, I think we can simplify this to:

```suggestion
  if (isDeliveryPharmacy) return null;
  
  if ((!fulfillment || fulfillment?.state === 'SENT') && readyBy && readyByDay) {
    return <PatientDesiredReadyBy readyBy={readyBy} readyByDay={readyByDay} />
  }
  
  if (fulfillment?.state === 'RECEIVED' && fulfillment?.pharmacyEstimatedReadyAt) {
    return <PharmacyEstimatedReadyAt pharmacyEstimatedReadyAt={fulfillment.pharmacyEstimatedReadyAt} />
  }
```

This makes it clearer that only one of these is being returned and when.

apps/patient/src/components/ReadyText.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const roundUpTo15MinInterval = (pharmacyEstimatedReadyAt: Date): Dayjs => {
+  const dayJS = dayjs(pharmacyEstimatedReadyAt);
```
marisbest2, line 38:
```suggestion
  const estimatedReadyAtAsDayJs = dayjs(pharmacyEstimatedReadyAt);
```

`dayJS` and `dayjs` are VERY close to each other

apps/patient/src/components/ReadyText.tsx
```diff
@@ -0,0 +1,106 @@
...
+    return dayJS.minute(roundedMinutes);
+  }
+};
+
+interface PharmacyEstimatedReadyAtProps {
+  pharmacyEstimatedReadyAt: Date;
+}
+const PharmacyEstimatedReadyAt = ({ pharmacyEstimatedReadyAt }: PharmacyEstimatedReadyAtProps) => {
+  const readyAt = roundUpTo15MinInterval(pharmacyEstimatedReadyAt);
+  const isTomorrow = readyAt.isSame(dayjs().add(1, 'day'), 'day');
```
marisbest2, line 53: Consider: https://day.js.org/docs/en/plugin/is-tomorrow

apps/patient/src/components/ReadyText.tsx
```diff
@@ -0,0 +1,106 @@
...
+    <Text>
+      Ready{' '}
+      {isTomorrow ? (
+        <b>Tomorrow at {readyAt.format(timeFormat)}</b>
+      ) : (
+        <>
+          at <b>{readyAt.format(timeFormat)}</b>
+        </>
+      )}
+    </Text>
```
marisbest2, line 66:
```suggestion
    <Text>
      Ready{isTomorrow ? <b> Tomorrow </b> : " "} at <b>{readyAt.format(timeFormat)}</b>
    </Text>
```

Something like this maybe cleaner. Might need to ensure the spaces are correct.

apps/patient/src/components/ReadyText.tsx
```diff
@@ -0,0 +1,106 @@
...
+}
+const PatientDesiredReadyBy = ({ readyBy, readyByDay }: PatientDesiredReadyByProps) => {
+  if (readyBy === 'Urgent') {
+    return (
+      <Text>
+        Need order <b>as soon as possible</b>
+      </Text>
+    );
+  } else if (readyBy === 'After hours') {
+    return readyByDay === 'Today' ? (
```
marisbest2, line 82: What happens if they said they need it tomorrow, but now its tomorrow? i.e. on Monday they said tomorrow and now its tuesday.

apps/patient/src/components/ReadyText.tsx
```diff
@@ -0,0 +1,106 @@
...
+        <b>
+          {hour} {period}
+          {readyByDay === 'Tomorrow' ? ' tomorrow' : ''}
+        </b>
+      </Text>
+    );
+  }
+};
+
+export default ReadyText;
```
marisbest2, line 106: Generally, prefer named exports (unless theres a reason that the library we're using doesnt like that, but I dont think thats the case here)

packages/sdk/src/types.ts
marisbest2: This still says optional

apps/patient/src/components/ReadyText.tsx
jknapp25: oh nice, way better

apps/patient/src/components/ReadyText.tsx
jknapp25: imported

apps/patient/src/components/ReadyText.tsx
jknapp25: oh good catch

apps/patient/src/components/ReadyText.tsx
```diff
@@ -0,0 +1,128 @@
...
+import isTomorrow from 'dayjs/plugin/isTomorrow';
+
+dayjs.extend(isTomorrow);
+
+interface ReadyTextProps {
+  readyBy?: string;
+  readyByDay?: string;
+  readyByTime?: string;
+  isDeliveryPharmacy?: boolean;
+  fulfillment?: Maybe<OrderFulfillment> | undefined;
```
marisbest2, line 13:
```suggestion
  fulfillment?: Maybe<OrderFulfillment>;
```

marisbest2 (approved):

apps/patient/src/components/ReadyText.tsx
jknapp25: cool

PR #1202, opened 2024-05-30
Remove Walmart med price pilot

jknapp25:
Now that we have enough data around surfacing med price for walmart, and given it didn't change conversion, there's no need to keep this around.


apps/patient/src/components/PharmacyInfo.tsx
```diff
@@ -90,9 +89,7 @@ export const PharmacyInfo = ({
   preferred = false,
   showDetails = true,
   boldPharmacyName = true,
-  isStatus = false,
-  price = null,
-  selected = false
```
marisbest2, line 95: Do we not need the selected?

apps/patient/src/utils/general.ts
```diff
@@ -7,11 +7,9 @@ import timezone from 'dayjs/plugin/timezone';
 import utc from 'dayjs/plugin/utc';
 import costcoLogo from '../assets/costco_small.png';
 import walgreensLogo from '../assets/walgreens_small.png';
-import walmartLogo from '../assets/walmart_small.svg';
```
marisbest2, line 10: Do we want to keep the Walmart logo? I get that price didnt make a difference but I feel like it's nice to be able to see the logo?

apps/patient/src/components/PharmacyInfo.tsx
jknapp25: it was only being used to surface the cash price disclaimer, now that that's getting removed, there's no reason for that prop

apps/patient/src/utils/general.ts
captDaylight: was thinking the same thing, but maybe this is something we don't give away for free to a pharmacy? But it is a nicer looking interface

marisbest2: Yeah that makes sense to me. Idk

marisbest2 (approved):

PR #1210, opened 2024-06-04
FIX: Mail Order Only Is Not showing Up in Pharmacy select Card

captDaylight:
If "Send to Patient" isn't in the tabs of the Pharmacy Select card, the tab body didn't show up until the tab is clicked.

Ticket Ref
https://www.notion.so/photons/Mail-Order-Only-isn-t-showing-up-88562f784b8c4664a322d7719d1a3eb2?pvs=4


jknapp25:
It's showing the tab names rn

image
image

jknapp25 (approved):

PR #1214, opened 2024-06-06
Track selected pharmacy rank in list

jknapp25:
To start getting a feel for how far down users are going into the list to find their pharmacy.

The reason why we're doing this with Google Analytics is so that we can pull this into Metabase.

### DIFF
--- a/apps/patient/src/views/Pharmacy.tsx
+++ b/apps/patient/src/views/Pharmacy.tsx
@@ -434,7 +434,7 @@ export const Pharmacy = () => {
   ): void => {
     // Get pharmacy index in list
     const index = pharmacies.findIndex((p) => p.id === selectedPharmacyId);
-    if (index != -1) {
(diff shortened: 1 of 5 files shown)


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -427,6 +428,22 @@ export const Pharmacy = () => {
     setShowFooter(true);
   };
 
+  const trackSelectedPharmacyRank = (
+    selectedPharmacyId: string,
+    pharmacies: EnrichedPharmacy[]
+  ): void => {
+    // Get pharmacy index in list
+    const index = pharmacies.findIndex((p) => p.id === selectedPharmacyId);
+    if (index != -1) {
```
marisbest2, line 437: !==

marisbest2 (approved):

PR #1219, opened 2024-06-12
Dropdown fix for mobile

captDaylight:
[Dropdown being hidden on mobile. ](https://photonhealth.slack.com/archives/C033T9N949M/p1718152280392819) I thought this might be more painful, glad it isn't.


packages/elements/src/photon-dropdown/photon-dropdown.tsx
```diff
@@ -214,6 +214,9 @@ export const PhotonDropdown = <T extends { id: string }>(props: {
           if (props.onOpen) {
             await props.onOpen();
           }
+          setTimeout(() => {
```
jknapp25, line 217: is there a delay?

jknapp25 (approved):

packages/elements/src/photon-dropdown/photon-dropdown.tsx
captDaylight: ah well, it seems to fix on my computer's simulator but I'm not quite seeing it on my phone yet with the branch version, so testing out some timeouts

PR #1224, opened 2024-06-13
Remove unnecessary GA setup tag

jknapp25:
Apparently since we're using react-ga4, we don't need to setup the usual google analytics tag in index.html. It handles that for us.


captDaylight (approved):
BYE BYE

PR #1228, opened 2024-06-14
Remove boilerplate code and add quickstart instructions to elements npm readme

enteladuka:
Removes boilerplate code in the readme and adds quick install instruction and a link to our docs for full details on elements


captDaylight (approved):

packages/elements/README.md
```diff
@@ -1,44 +1,25 @@
...
+
+```html
+<script src="https://cdn.rx.dev/elements/@latest/dist/index.js" type="module"></script>
+```
 
-- First Parameter is the WebComponent Tag Name. This is the name under which your component will be registered and usable from HTML. In this example, `photon-my-first-component`
-- Second Parameter is the default values of properties. If you have properties a user isn’t supposed to specify, you can declare the defaults here. Any properties that are not kebab case (e.g first-word) are transformed to kebab-case when passed from HTML/JSX (per WebComponent Spec Guidelines)
-- Third Parameter is a function of (props, options) ⇒ JSX.Element. NOTE: Do not de-structure `props` as in SolidJS this will make your properties (which are a Proxy object) lose reactivity. The second parameter, `options` contains a singular property `element` which gives you raw access to your DOM element created by this web-component. This is considered advanced functionality and rarely used (though I use it in our codebase and have examples of when it’s appropriate).
+Full documentation of available elements can be found here: [Photon Elements Documentation](https://docs.photon.health/docs/elements)
```
captDaylight, line 25: Maybe say also "and example configurations" or something

enteladuka: done!

PR #1235, opened 2024-06-20
Disable STP for DrTelx

jknapp25:


packages/settings/src/lib/photon.ts
```diff
@@ -250,7 +250,6 @@ const organizationSettings: {
   org_zc1RzzmSwd8eE94U: {
     logo: 'sesame_logo.jpg',
     accentColor: '#5224C7',
-    sendToPatient: true,
```
jknapp25, line 253: Sesame: stp defaults to tru, not needed

samkotlove (approved):

PR #1238, opened 2024-06-20
Adv med dropdown fix

captDaylight:
This is second attempt of this https://github.com/Photon-Health/client/pull/1219 [dropdown being hidden on mobile. ](https://photonhealth.slack.com/archives/C033T9N949M/p1718152280392819)

BEFORE
Screen Shot 2024-06-20 at 4 07 18 PM

NOW
Screen Shot 2024-06-20 at 4 06 14 PM

### DIFF
--- a/packages/elements/src/photon-med-search/components/MedicationFilterDropdown.tsx
+++ b/packages/elements/src/photon-med-search/components/MedicationFilterDropdown.tsx
@@ -1,7 +1,7 @@
 import { Medication, SearchMedication } from '@photonhealth/sdk/dist/types';
-import { createEffect, createSignal, createMemo } from 'solid-js';
+import { createEffect, createSignal, createMemo, Show, For } from 'solid-js';
 import { gql } from '@apollo/client';
-import { usePhotonClient } from '@photonhealth/components';
+import { usePhotonClient, ComboBox, Text } from '@photonhealth/components';
 import { PhotonDropdown } from '../../photon-dropdown';
 
 const GET_FORMS = gql`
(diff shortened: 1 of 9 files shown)


packages/elements/src/photon-med-search/components/MedicationFilterDropdown.tsx
```diff
@@ -90,25 +90,39 @@ export const MedicationFilterDropdown = (props: MedicationFormDropdownProps) =>
     }
   });
 
+  createEffect(() => {
+    if (selected()?.id) {
+      setFilterId(selected()!.id);
+      props.setMedicationId(selected()!.id);
```
marisbest2, line 96: This feels like it could end up with two different event loops. I think it might be a good idea to use [`batch`](https://www.solidjs.com/tutorial/reactivity_batch)

packages/elements/src/photon-med-search/photon-med-search-component.tsx
```diff
@@ -124,8 +125,9 @@ customElement(
 
     return (
       <div ref={ref}>
+        <style>{photonStyles}</style>
         <p class="font-sans text-lg text-gray-700 pb-2">{props.title}</p>
-        <div class="flex flex-col xs:flex-row gap-4">
```
marisbest2, line 128: Why was this ever `flex-row`?

marisbest2 (commented):
Two comments but looks good

packages/elements/src/photon-med-search/photon-med-search-component.tsx
captDaylight: no idea, I was confused by this too

marisbest2 (approved):

PR #1243, opened 2024-06-25
Fix refill count incorrect on patient app

jknapp25:
We had an extra `countFillsAndRemoveDuplicates` call which reset the refill count back to 0, removed that

### DIFF
--- a/apps/patient/src/views/Main.tsx
+++ b/apps/patient/src/views/Main.tsx
@@ -64,8 +64,7 @@ export const Main = () => {
       console.log('handleOrderResponse', order);
       setOrder(order);
 
-      const preppedFills = countFillsAndRemoveDuplicates(order.fills);
-      setFlattenedFills(countFillsAndRemoveDuplicates(preppedFills));
+      setFlattenedFills(countFillsAndRemoveDuplicates(order.fills));
 
       datadogRum.setGlobalContextProperty('organizationId', order.organization.id);
       datadogRum.setGlobalContextProperty('orderId', orderId);
@@ -179,6 +178,8 @@ export const Main = () => {
     setOrder,
     logo
   };
+
+  console.log(flattenedFills);
 
   return (
     <ChakraProvider theme={theme(order?.organization.id)}>
--- a/apps/patient/src/views/Main.tsx
+++ b/apps/patient/src/views/Main.tsx
@@ -179,8 +179,6 @@ export const Main = () => {
     logo
   };
 
-  console.log(flattenedFills);
-
   return (
     <ChakraProvider theme={theme(order?.organization.id)}>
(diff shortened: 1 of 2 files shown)


marisbest2:
One thing:

I would consider also rewriting `countFillsAndRemoveDuplicates` to be idempotent. Something like this _should_ work (though I would test it, and if you don't like `reduce` possibly rewrite as loops)

```ts
export const countFillsAndRemoveDuplicates = (fills: (FillWithCount | Fill)[]): FillWithCount[] => {
  const counts = fills.reduce((acc, f) => {
    if (!(f.treatment.id in acc)) {
      acc[f.treatment.id] = 0;
    }
    acc[f.treatment.id] += "count" in f ? f.count : 1;
    return acc;
  }, {} as Record<string, number>);

  const distinctFills: Map<string, FillWithCount> = fills.reduce((acc: Map<string, FillWithCount>, f) => {
    if (!acc.has(f.treatment.id)) {
      acc.set(f.treatment.id, { ...f, count: counts[f.treatment.id] })
    }
    return acc;
  }, new Map<string, FillWithCount>());

  return Array.from(distinctFills.values());
};
```

This is a case where typescript actually fails us. Because `FillWithCount` is a super-type of `Fill` it'll actually safely fit.

We can alternatively do 
```ts
export const countFillsAndRemoveDuplicates = (fills: (Fill & { count?: never }[]): FillWithCount[] 
```
Which would throw an error if you tried to call it twice. 

But that's less ideal as seen below:
```ts
// passes, which is bad
declare const f2: (x: Fill[]) => FillWithCount[];
f2(f2([])) 


type FillWithoutCount = Fill & { count?: never };

declare const f: (x: FillWithoutCount[]) => FillWithCount[];

f(f([])); /* Argument of type 'FillWithCount[]' is not assignable to parameter of type 'FillWithoutCount[]'.
  Type 'FillWithCount' is not assignable to type 'FillWithoutCount'.
    Type 'FillWithCount' is not assignable to type '{ count?: undefined; }'.
      Types of property 'count' are incompatible.
        Type 'number' is not assignable to type 'undefined'.(2345)*/


// But this also passes, which is bad
const a: Fill[] = f([]);
f(a)

File: apps/patient/src/views/Main.tsx [hunk 1]
```diff
@@ -180,6 +179,8 @@ export const Main = () => {
     logo
   };
 
+  console.log(flattenedFills);
```
marisbest2, line 182: Make sure we remove this before we merge

marisbest2 (approved):

apps/patient/src/views/Main.tsx
jknapp25: removed

PR #1267, opened 2024-07-17
Capsule x Sesame next day

jknapp25:
Turning on Capsule for Sesame as part of our monetization test.
 - enabled courier for Sesame
 - limited to non-GLP1's
 - updated pharmacy icons to match [designs](https://www.figma.com/design/BkeblOMtaBc6BM5TDZ0VPJ/%F0%9F%93%B2-Patient-App?node-id=1675-5680&t=WHSwkfxMHKc4vqfH-0)

Example
image


apps/patient/src/components/BrandedPharmacyCard.tsx
```diff
@@ -1,56 +1,68 @@
...
   if (!brand) return null;
 
-  const tagline = brand.description ? (
-    <Text fontSize="sm" color="gray.500">
-      {brand.description}
-    </Text>
-  ) : null;
+  const pharmacy = { id: pharmacyId, name: brand.name, logo: brand.logo };
+
+  const selectable = !!handleSelect;
```
marisbest2, line 51: I personally find it easier to read `handleSelect == null` but 🤷

apps/patient/src/components/BrandedPharmacyCard.tsx
```diff
@@ -1,56 +1,68 @@
...
-    <Box bgColor="white" w="full" cursor="pointer" onClick={() => handleSelect?.(pharmacyId)}>
-      <VStack align="start" spacing={2}>
-        <Image src={brand.logo} width="auto" height="30px" />
-        {tagline}
-      </VStack>
-    </Box>
+    <Card
+      bgColor="white"
+      border="2px solid"
+      borderColor={selected && selectable ? 'brand.500' : 'white'}
```
marisbest2, line 57: Why is this only shown with selected style if its also `selectable`?

apps/patient/src/components/BrandedPharmacyCard.tsx
```diff
@@ -1,56 +1,68 @@
...
+      borderRadius="lg"
+      onClick={() => selectable && handleSelect(pharmacyId)}
+      mx={{ base: -3, md: undefined }}
+      cursor={selectable ? 'pointer' : undefined}
+    >
+      <CardBody p={3}>
+        <PharmacyInfo pharmacy={pharmacy} tagline={brand.description} boldPharmacyName={false} />
```
marisbest2, line 64: I'm kinda surprised this even typechecks. I guess its because we have a bunch of optional types. 

Its going to be annoying later than we're only providing `id` `name` and `logo` to this component. 

But 🤷 , deal with it later

marisbest2 (approved):

apps/patient/src/components/BrandedPharmacyCard.tsx
marisbest2: Specifically here I'd do one of two things:
- make `PharmacyInfo` accept a `id`, `name` and `logo` (instead of `pharmacy`). That way you know those are the only pieces you have.
- Make `BrandedPharmacyCard` accept a `Pharmacy` rather than `pharmacyId`. Then pass in the pharmacy to `PharmacyInfo`

jknapp25: Yeah this is the pain of not having correct delivery pharmacy types in our db

apps/patient/src/components/BrandedPharmacyCard.tsx
jknapp25: oh good catch, I think that was residual from when we didn't have PharmacyInfo,. removing

apps/patient/src/components/BrandedPharmacyCard.tsx
jknapp25: removed selectable, made handleSelect required

PR #1293, opened 2024-08-02
infinite radius pharmacy search

ritabulman:
removed default `radius` values when getting pharmacies by location


whittle-photon (commented):
Why is radius made optional in two places and removed in all the others? The PR description says it removes the default radius—is it still possible to specify a radius at all?

ritabulman:
so I removed `radius` where it was being passed in as a default value, and changed `radius` to optional where it could potentially be passed in as a parameter. 

removed `radius` where it was harcoded as an argument because it didn't seem to be pulling a value from any UI element

PR #1302, opened 2024-08-09
Fix optional externalId bug

samkotlove:


marisbest2 (approved):

marisbest2:
You need to bump the package

marisbest2:
You need to bump the package version

PR #1317, opened 2024-08-16
fix edit patient bug to make sure name is actually updated

marisbest2:
This was a pain. Because `PatientInfo` owns its own data and because `components` and `elements` don't share a store, the only way to force-retrigger is a rerender is to clear out the patientId and then reset it. But because solid does batching, you also need to set a timeout.

Also added a drop of types


jknapp25 (approved):
Tested, looks good 👍

PR #1322, opened 2024-08-16
codegen the patient app

marisbest2:
Add codegen to the patient app.

- We're always using `patient-api` for the patient app so its easy to point at
- The codegen will hot-reload if you run `npm run patient` but you can also run `nx run patient:codegen:watch` (or just `nx run patient:codegen` if you don't want to watch
- For now, this keeps the `graphql-request` library as the client. This provides no caching or anything, so we should eventually revisit that
- We currently co-locate all our queries and mutations. This makes some stuff easy (i.e. the types are consistent across pages) but other stuff hard. For a simple app like the patient app, I think the current approach is fine. But as the app gets more complex we might want to move to Apollo or Relay
- By generating our own types, this PR also removes the dependency on the sdk's package because we weren't actually using it

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 node_modules
 dist
 apps/app/src/gql
+__generated__
--- a/.github/workflows/prettier.yml
+++ b/.github/workflows/prettier.yml
@@ -22,5 +22,8 @@ jobs:
       - name: Install Dependencies
         run: npm ci
 
+      - name: Codegen
+        run: npx run-many --target=codegen --all
+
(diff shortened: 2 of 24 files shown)


jknapp25:
seeing some type issues when running locally

image

marisbest2:
@jknapp25 Probably have to run the codegen

marisbest2:
You might need to run the codegen or maybe `nx reset` before `npm run patient`. It _should_ run the codegen on start

jknapp25 (approved):
this is great! way cleaner

PR #1336, opened 2024-08-20
Set top ranked pharmacy search radius to 15mi

jknapp25:
Right now we are returning costcos that are real far away (see below) for top ranked pharmacies like Costco we should limit the radius in search.

image (40)

### DIFF
--- a/apps/patient/src/api/internal.ts
+++ b/apps/patient/src/api/internal.ts
@@ -54,7 +54,7 @@ export const getPharmacies = async ({
       GET_PHARMACIES,
       {
         location: {
-          radius: searchParams.radius ?? 100,
+          radius: 100,
           ...searchParams
         },
         limit,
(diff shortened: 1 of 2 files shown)


apps/patient/src/api/internal.ts
```diff
@@ -53,7 +54,7 @@ export const getPharmacies = async ({
       GET_PHARMACIES,
       {
         location: {
-          radius: 100,
+          radius: searchParams.radius ?? 100,
```
marisbest2, line 57: These look like they're the same?

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -245,7 +245,7 @@ export const Pharmacy = () => {
       }
       try {
         const topRankedCostco: EnrichedPharmacy[] = await getPharmacies({
-          searchParams: { latitude, longitude },
+          searchParams: { latitude, longitude, radius: 15 },
```
marisbest2, line 248: This is what matters

apps/patient/src/api/internal.ts
jknapp25: not sure i understand

jknapp25: got it

marisbest2 (approved):

PR #1345, opened 2024-09-05
fixing spacing in the treatments tab

billoran8:
as per [this](https://www.notion.so/photons/Update-templates-catalog-layout-d7b8c3feb46e49148bf2a4eb422cc121) ticket, there are some spacing issues in the catalogs section.

the overflow issue has been resolved by switching from `vw` units to `100%`. `vw` will always be the view width, which sometimes can be too big - we instead want it to fill the parent container, which is what changing to `100%` does.

additionally, there was some annoying vertical discrepancy between the split contents due to differences in handling the first child vs the second. i've standardized it so they both have the same amount of margin.

### DESKTOP
treatments_before_desktop

**issues:** 

1. the vertical alignment of the table and the `add treatment` section on the right
2. the overflow of the table

### MOBILE
treatments_before_mobile

### DESKTOP 
treatments_after_desktop

**solutions:** 

1. the alignment of the table and `add treatment section` is uniform
2. the overflow of the table has been resolved

### MOBILE
treatments_after_mobile

### DIFF
--- a/apps/app/src/views/components/SplitLayout.tsx
+++ b/apps/app/src/views/components/SplitLayout.tsx
@@ -18,7 +18,6 @@ export const SplitLayout = (props: any) => {
         width={isMobileAndTablet ? '100%' : 'full'}
         alignSelf={isMobileAndTablet ? 'center' : 'flex-start'}
         mt={2}
-        px={isMobileAndTablet ? 4 : 0}
       >
         {children[0]}
       </Box>
--- a/apps/app/src/views/components/SplitLayout.tsx
+++ b/apps/app/src/views/components/SplitLayout.tsx
@@ -10,14 +10,14 @@ export const SplitLayout = (props: any) => {
     <Stack
       flexDir={isMobileAndTablet ? 'column' : 'row'}
       gap={5}
-      width="95vw"
+      width="100%"
       maxW="1400"
       justify="center"
     >
       <Box
-        width={isMobileAndTablet ? '100vw' : 'full'}
+        width={isMobileAndTablet ? '100%' : 'full'}
         alignSelf={isMobileAndTablet ? 'center' : 'flex-start'}
-        pt={2}
+        mt={2}
         px={isMobileAndTablet ? 4 : 0}
       >
         {children[0]}
@@ -29,6 +29,7 @@ export const SplitLayout = (props: any) => {
           shadow="md"
           minW="400px"
           alignSelf={isMobileAndTablet ? 'center' : 'flex-start'}
+          mt={2}
           p={6}
         >
           {children[1]}
--- a/apps/app/src/views/routes/Settings/views/TreatmentTab.tsx
+++ b/apps/app/src/views/routes/Settings/views/TreatmentTab.tsx
@@ -2,6 +2,7 @@ import { useMutation } from '@apollo/client';
 import {
   Alert,
   AlertIcon,
+  Box,
   Modal,
   ModalBody,
   ModalCloseButton,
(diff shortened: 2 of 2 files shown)


apps/app/src/views/components/SplitLayout.tsx
```diff
@@ -10,14 +10,14 @@ export const SplitLayout = (props: any) => {
     <Stack
       flexDir={isMobileAndTablet ? 'column' : 'row'}
       gap={5}
-      width="95vw"
+      width="100%"
```
billoran8, line 13: this should be `%` instead of `vw` as we want the stack to fill the width of the parent, not the viewport width.

apps/app/src/views/components/SplitLayout.tsx
```diff
@@ -29,6 +29,7 @@ export const SplitLayout = (props: any) => {
           shadow="md"
           minW="400px"
           alignSelf={isMobileAndTablet ? 'center' : 'flex-start'}
+          mt={2}
```
billoran8, line 32: the second child not having any padding or margin causes it to be about 6 pixels higher than the first. i'm adding it here to be equivilant to the first child.

apps/app/src/views/components/SplitLayout.tsx
```diff
@@ -10,14 +10,14 @@ export const SplitLayout = (props: any) => {
...
+      width="100%"
       maxW="1400"
       justify="center"
     >
       <Box
-        width={isMobileAndTablet ? '100vw' : 'full'}
+        width={isMobileAndTablet ? '100%' : 'full'}
         alignSelf={isMobileAndTablet ? 'center' : 'flex-start'}
-        pt={2}
+        mt={2}
```
billoran8, line 20: changing this from padding to margin to be consistent with the second child

jknapp25:
Can we drop the sidekick treatment search and just go with parity with the templates tab ui?

image

marisbest2 (approved):
Looks great

billoran8:
good q whittle, lemme check with jomi

apps/app/src/views/components/SplitLayout.tsx
```diff
@@ -10,14 +10,14 @@ export const SplitLayout = (props: any) => {
...
         px={isMobileAndTablet ? 4 : 0}
```
thebadprince, line 21: Screenshot 2024-09-05 at 3 05 02 PM

Think this would look even sweeter if we can remove the left and right padding on mobile/tablet 🙏

thebadprince:
> Can we drop the sidekick treatment search and just go with parity with the templates tab ui? (aka stick search in modal)
> 
> image

Agree here but we can prob hold off on it. Still TBD what we'll do with the concept of a Catalog — whether we even keep it or not or if there's a config at the org level where an org can opt to have a formulary or not. Either way, we'll need to update the designs so idk if we should change much right now aside from styling at the moment.

After we finish med search, we can revisit how all of Settings ought to be (which reminds me that I prob need to do a design pass on all the tabs so we can make the whole thing consistent design-wise).

ANYWAY, last thing that bugs me about this page is the black check mark inside the selection but this is at the component level:

image

Other than that, LGTM @billoran8 !

billoran8:
alright! i'll remove the left and right padding on mobile and update this PR when that's ready for review.

i'll make a ticket for changing the black checkmark to white and tag you in it @thebadprince .

re: the future of catalogs - maybe a low prio ticket to spitball and brainstorm and spike on how we see it evolving when it becomes more of a customer need?

billoran8:
i've removed the padding from the splitlayout on device/mobile and will make sure i get jomi's approval

thebadprince:
LGTM!! 🚀

PR #1354, opened 2024-09-10
Limit med search to 3 or more chars in search string

jknapp25:
Search can tend to take a bit when you're only searching for "s". Limiting to 3 or more worked well for our prototypes. Responses are about to get larger with us returning generic as well when searching branded.

### DIFF
--- a/packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
+++ b/packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
@@ -257,10 +257,11 @@ const Component = (props: ComponentProps) => {
   const tryLoadTreatmentOptions = async (searchTerm: string) => {
     setLoadingTreatmentOptions(true);
 
-    const treatmentOptions = searchTerm
-      ? await loadTreatmentOptions(client!.sdk.apolloClinical, searchTerm)
-      : // Set treatment options to empty array if search term is empty
-        [];
+    const treatmentOptions =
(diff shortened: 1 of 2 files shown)


packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
```diff
@@ -257,10 +257,11 @@ const Component = (props: ComponentProps) => {
...
 
-    const treatmentOptions = searchTerm
-      ? await loadTreatmentOptions(client!.sdk.apolloClinical, searchTerm)
-      : // Set treatment options to empty array if search term is empty
-        [];
+    const treatmentOptions =
+      searchTerm?.length >= 3 // 3 min chars for smaller responses
+        ? await loadTreatmentOptions(client!.sdk.apolloClinical, searchTerm)
+        : // Set treatment options to empty array if search term is empty
+          [];
```
marisbest2, line 264: Nit, I would differentiate between `[]` which means "no results" and `null` which can mean "not searching"

Also will make it easier to solve the fact that it first shows "No Data"

marisbest2 (approved):

packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
jknapp25: that's good feedback, need to think more about this because it affects the 3 dropdowns, moving forward with this small change for the moment.

PR #1357, opened 2024-09-12
WIP, no need to look, just adding notes as I go

captDaylight:


captDaylight:
moved the extension into it's own repo

PR #1369, opened 2024-09-13
Update costco logic to always show instead of just for GLP1 orders

samkotlove:


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -132,6 +132,9 @@ export const Pharmacy = () => {
   const hasTopRankedCostco = topRankedPharmacies.some((p) => p.name === 'Costco Pharmacy');
   const enableMailOrder =
     !isDemo &&
+    // If we're loading we shouldn't show pharmacies of mail order (we don't even know if we want to show until
+    // we see if there are top ranked costcos)
+    !loadingPharmacies &&
```
samkotlove, line 137: This is only semi-related, ui fix. Without this two weird behaviors arise:
- You can see mail order before local pharmacies load
- With costco, it was showing mail order until pharmacies load, after which it would hide the mail order

The latter couldve also been fixed by checking if toprankedwasenabled

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -344,7 +347,7 @@ export const Pharmacy = () => {
         let topRankedPharmacies: EnrichedPharmacy[] = [];
 
         // check if top ranked costco is enabled and there are GLP treatments
-        if (enableTopRankedCostco && containsGLP) {
+        if (enableTopRankedCostco) {
```
samkotlove, line 350: This is the intended change

jknapp25 (approved):

apps/patient/src/views/Pharmacy.tsx
samkotlove: Coming back, doing the latter instead to avoid other design considerations

PR #1376, opened 2024-09-17
Allow adding whole organizations to the new medication search

samkotlove:
Next se'll need to be able to turn on whole orgs and not just individual users

### DIFF
--- a/packages/elements/src/photon-multirx-form/photon-prescribe-workflow.tsx
+++ b/packages/elements/src/photon-multirx-form/photon-prescribe-workflow.tsx
@@ -66,6 +66,11 @@ const newMedSearchTesters = [
   'Caitlyn Call',
   'Melissa Lotto'
 ];
+// Note: this has to be environment specific
+const newMedSearchOrgs = [
+  // Piction
+  'org_v5It8IoY0RH1Rw80'
+];
 
 export type Address = {
   city: string;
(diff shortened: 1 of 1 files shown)


packages/elements/src/photon-multirx-form/photon-prescribe-workflow.tsx
```diff
@@ -164,9 +169,9 @@ export function PrescribeWorkflow(props: PrescribeProps) {
   });
   createEffect(() => {
     if (authenticated()) {
-      const enableNewMedicationSearch = newMedSearchTesters.includes(
-        client?.authentication.state.user?.name
-      );
+      const enableNewMedicationSearch =
+        newMedSearchTesters.includes(client?.authentication.state.user?.name) ||
+        newMedSearchOrgs.includes(client?.authentication.state.user?.org_id);
```
marisbest2, line 174: Not to be annoying, but can we put this in `package/settings`?

samkotlove: Don't know that it can, elements doesnt seem to ahve access to `@client/settings`

jknapp25 (approved):

packages/elements/src/photon-multirx-form/photon-prescribe-workflow.tsx
captDaylight: This would only apply to customers who embed with elements. Piction for example looks to be using our site, so  the way we typically did this in the past is we did this for them was to add a parameter to the prescribe flow https://github.com/Photon-Health/client/blob/00d6a407467f0938f842f66c58678bd5297f05ff/packages/elements/src/photon-multirx-form/photon-prescribe-workflow-component.tsx#L30C9-L30C25

For features that will be requirements, we just asked them to bump the version

PR #1379, opened 2024-09-17
Release med search on clinical app

captDaylight:
Style can be tweaked, but functionality allows for a closable educational banner.

This will release the new med search to all App users

Screen Shot 2024-09-17 at 6 31 47 PM

### DIFF
--- a/packages/components/src/particles/Banner/index.tsx
+++ b/packages/components/src/particles/Banner/index.tsx
@@ -1,21 +1,46 @@
-import { Show, JSXElement } from 'solid-js';
+import { createSignal, JSXElement, onMount, Show } from 'solid-js';
 import Icon from '../Icon';
 
 export type BannerProps = {
   children: JSXElement;
   status: 'success' | 'info';
   withoutIcon?: boolean;
+  closable?: boolean;
+  name?: string;
 };
 
 export default function Banner(props: BannerProps) {
+  const [isVisible, setIsVisible] = createSignal(true);
+
+  onMount(() => {
+    const storedState = localStorage.getItem(`banner-${props.name}`);
+    if (storedState === 'closed') {
+      setIsVisible(false);
+    }
+  });
+
+  const closeBanner = () => {
+    setIsVisible(false);
+    if (props.name) {
+      localStorage.setItem(`banner-${props.name}`, 'closed');
+    }
+  };
+
   return (
-    <div class="text-blue-600 flex items-center gap-2 bg-blue-50 py-3 px-3 sm:px-4 rounded-lg">
-      <Show when={!props.withoutIcon}>
-        <div class="flex-shrink-0">
-          <Icon name="informationCircle" size="sm" />
-        </div>
-      </Show>
-      {props.children}
-    </div>
+    <Show when={isVisible()}>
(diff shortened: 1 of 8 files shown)


packages/components/src/particles/Banner/index.tsx
```diff
@@ -1,21 +1,46 @@
...
   withoutIcon?: boolean;
+  closable?: boolean;
+  name?: string;
 };
 
 export default function Banner(props: BannerProps) {
+  const [isVisible, setIsVisible] = createSignal(true);
+
+  onMount(() => {
+    const storedState = localStorage.getItem(`banner-${props.name}`);
```
marisbest2, line 16: this seems kinda wild?

But if we're going to do it, I'd make it something like `photon-banner-`

marisbest2 (approved):

packages/components/src/particles/Banner/index.tsx
```diff
@@ -1,21 +1,46 @@
-import { Show, JSXElement } from 'solid-js';
+import { createSignal, JSXElement, onMount, Show } from 'solid-js';
 import Icon from '../Icon';
 
 export type BannerProps = {
   children: JSXElement;
   status: 'success' | 'info';
   withoutIcon?: boolean;
+  closable?: boolean;
+  name?: string;
```
billoran8, line 9: not a defect just a thought - seems like this is more of an `identifier` than a name? feel free to disregard i can just see the word `name` obscuring the true purpose of this prop

billoran8 (approved):
not gonna mark it as a defect but i'd love to start exploring some form of testing - maybe we could start exploring [jest snapshot tests](https://jestjs.io/docs/snapshot-testing) to verify that our UIs don't change in unexpected ways? or do we sort of just check the storybook?

do we sort of just assume that things have been manually verified? are there sets of edge-cases in the space that we all know about when manually verifying?

packages/components/src/particles/Banner/index.tsx
marisbest2: Agree with Bill

packages/components/src/particles/Banner/index.tsx
captDaylight: updated

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -221,7 +221,7 @@ export const AddPrescriptionCard = (props: {
         >
           <Show when={props.enableNewMedicationSearch}>
             <div class="mb-2">
-              <Banner status="info" withoutIcon closable name="new-medication-search-banner">
+              <Banner status="info" withoutIcon closable id="new-medication-search-banner">
```
billoran8, line 224: lol now that i think about it is `id` a reserved property? it hink it might be

packages/components/src/particles/Banner/index.tsx
captDaylight: That makes sense, updated!

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
captDaylight: naw I think it's all good imo, I think it's still idiomatic

PR #1388, opened 2024-09-20
singlecare med price UI rebase

marisbest2:
Adds
- Ability to search pharmacies by cash price
- Discount card section on status page
- Modal for discount card details


apps/patient/src/api/internal.ts
```diff
@@ -34,29 +49,48 @@ export const getPharmacies = async ({
...
+      : graphQLClient.GetPharmaciesByLocation;
+    return formatPharmacies(
+      await fn({
+        location: {
+          radius: 100,
+          ...searchParams
+        },
+        openAt: isOpenNow ? now : undefined,
+        is24hr,
+        ...(!includePrice ? { limit, offset, name } : {})
```
marisbest2, line 70: Why dont we have limit/offset here?

captDaylight: I didn't see it in on the services query, @jknapp25 is that something we plan on adding before releasing this?

apps/patient/src/graphql/queries.ts
```diff
@@ -128,53 +181,38 @@ export const GET_PHARMACIES = gql`
...
+  }
+  ${PHARMACY_FIELDS}
+`;
+
+// TODO THIS IS TEMPORARY, I don't know what the query looks like yet
+export const GET_DISCOUNT_CARD = gql`
+  query GetDiscountCard($id: ID!) {
+    discountCard: order(id: $id) {
+      id
+      price: id
```
captDaylight, line 211: @marisbest2 It looks like `id` got added to all of these

marisbest2: Yeah, this is just a stopgap

apps/patient/src/api/internal.ts
jknapp25: it got added on getPharmacyPrices...

apps/patient/src/api/internal.ts
```diff
@@ -34,29 +49,48 @@ export const getPharmacies = async ({
...
+  try {
+    const response = await graphQLClient.GetDiscountCard({
+      id
     });
 
-    if (response?.pharmaciesByLocation?.length > 0) {
-      return response.pharmaciesByLocation;
+    if (!response?.discountCard?.bin) {
+      throw new Error('No discount card returned');
     }
```
marisbest2, line 88: Why is this an error?

apps/patient/src/api/internal.ts
marisbest2: So whats the deal here?

apps/patient/src/components/CouponDetails.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { FiInfo } from 'react-icons/fi';
+import { CouponModal } from '../components';
+import { text as t } from '../utils/text';
+import { useOrderContext } from '../views/Main';
+
+export const CouponDetails: FC = () => {
+  const [couponModalOpen, setCouponModalOpen] = useState<boolean>(false);
+  const { order } = useOrderContext();
+
+  const { price, bin, pcn, group, memberId } = order.discountCards[0];
```
marisbest2, line 12: This is a place where typescript falls over. `order.discountCards[0]` can be undefined and this would fall over. So we should first test if length > 0

apps/patient/src/components/CouponDetails.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <Text fontWeight="semibold" fontSize="md">
+          {t.showThisCoupon}
+        </Text>
+      </Box>
+      <HStack w="full" align="start">
+        <VStack w="30%" align="start">
+          <Text>{t.bin}</Text>
+          <Text>{t.pcn}</Text>
+          <Text>{t.group}</Text>
+          <Text>{t.memberId}</Text>
```
marisbest2, line 34: Given that these are actual things, is there a reason we need to pull this from `t`? Can't we just do `BIN`, `PCN` etc

apps/patient/src/components/CouponDetails.tsx
```diff
@@ -0,0 +1,58 @@
...
+        </VStack>
+        <VStack w="70%" align="start">
+          <Text as="b">{bin}</Text>
+          <Text as="b">{pcn}</Text>
+          <Text as="b">{group}</Text>
+          <Text as="b">{memberId}</Text>
+        </VStack>
+      </HStack>
```
marisbest2, line 42: This feels a bit weird. I would probably do it as a `<VStack><HStack/><HStack/><HStack/><HStack/></VStack>`

marisbest2: Where each `<HStack>` is effectively a row.

apps/patient/src/utils/models.ts
```diff
@@ -2,18 +2,29 @@ import {
...
 export type Order = NotMaybe<GetOrderQuery['order']>;
 export type Fill = Order['fills'][number];
+export interface DiscountCard {
+  id: string;
+  price: string;
+  bin: string;
+  pcn: string;
+  group: string;
+  memberId: string;
+}
```
marisbest2, line 19: I would pull this off `Order` rather than hardcoding like this.

jknapp25: oh think this is leftover from pre-codegen

samkotlove (approved):

apps/patient/src/components/CouponDetails.tsx
jknapp25: yeah that's better, was doing a quick and dirty thing to make sure they aligned, width works

apps/patient/src/components/CouponDetails.tsx
jknapp25: oh thought i had a check for that

PR #1394, opened 2024-09-25
Add template name in search

captDaylight:
https://photonhealth.slack.com/archives/C02V18YAEE9/p1727284924730709

### DIFF
--- a/packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
+++ b/packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
@@ -115,7 +115,9 @@ function getFilteredData(
   // Filter the catalog data and treatment options based on the search text
   return catalogData.filter((item) => {
     const itemName =
-      'treatment' in item ? item.treatment.name.toLowerCase() : item.name.toLowerCase();
+      'treatment' in item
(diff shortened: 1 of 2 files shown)


jknapp25 (approved):

packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
```diff
@@ -115,7 +115,9 @@ function getFilteredData(
   // Filter the catalog data and treatment options based on the search text
   return catalogData.filter((item) => {
     const itemName =
-      'treatment' in item ? item.treatment.name.toLowerCase() : item.name.toLowerCase();
+      'treatment' in item
+        ? `${item?.name?.toLowerCase() || ''} ${item.treatment.name.toLowerCase()} `
```
marisbest2, line 119: Is the space here going to screw things up if there is no name?

I think you should just move this into the `every` below

captDaylight: the space doesn't have an effect, I've tested terms in both
Screen Shot 2024-09-25 at 2 19 06 PM
Screen Shot 2024-09-25 at 2 19 55 PM

PR #1420, opened 2024-10-04
Save patientSelectedPrice flag

jknapp25:
We updated the discount cards table with this flag so that we can better track price selections.

- Now passing `patientSelectedPrice` into rerouteOrder and setOrderPharmacy
- Updated status page to ensure we're showing a discount card for the current pharmacy
- Also did a lil pre-work for when we'll have multiple discount cards. It's not clear yet when we'll do this, maybe in the next month or so, I'm fine with removing, but felt nice to have

### DIFF
--- a/apps/patient/src/views/Pharmacy.tsx
+++ b/apps/patient/src/views/Pharmacy.tsx
@@ -548,7 +548,7 @@ export const Pharmacy = () => {
     trackSelectedPharmacyRank(selectedId, allPharmacies);
 
     try {
-      const patientSelectedPrice = !!allPharmacies.find((p) => p.id === selectedId)?.price;
+      const patientSelectedPrice = sortBy === 'price';
       const result = isReroute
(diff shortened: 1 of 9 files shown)


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -548,14 +548,16 @@ export const Pharmacy = () => {
     trackSelectedPharmacyRank(selectedId, allPharmacies);
 
     try {
+      const patientSelectedPrice = !!allPharmacies.find((p) => p.id === selectedId)?.price;
```
captDaylight, line 551: Is `allPharmacies` only a list of pharmacies with prices? or could that also be pharmacies from the distance list?

jknapp25: reverted to the way we're doing it for DD right now, `sortBy === 'price'`

captDaylight (approved):

PR #1427, opened 2024-10-10
Add catalog params to photon-prescribe-workflow

samkotlove:
Two additional params are added:

`catalog-id`
This allows constraining a caller to a specific catalog

`allow-off-catalog-search`
This gives control to allowing or blocking searching for items _not_ on the catalog

These will be used in mail order configurations, so a customer can programmatically constrain the meds allowed to be prescribed.

### DIFF
--- a/packages/elements/src/stores/catalog.ts
+++ b/packages/elements/src/stores/catalog.ts
@@ -9,7 +9,7 @@ const CATALOG_FIELDS = gql`
     id
   }
 `;
-const CatalogtFieldsMap = { CatalogFields: CATALOG_FIELDS };
+const CatalogFieldsMap = { CatalogFields: CATALOG_FIELDS };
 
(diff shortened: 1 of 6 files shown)


packages/elements/src/stores/catalog.ts
```diff
@@ -4,6 +4,13 @@ import { GraphQLError } from 'graphql';
 import gql from 'graphql-tag';
 import { createStore } from 'solid-js/store';
 
+const CATALOG_FIELDS = gql`
+  fragment CatalogFields on Catalog {
+    id
+  }
+`;
+const CatalogtFieldsMap = { CatalogFields: CATALOG_FIELDS };
```
jknapp25, line 12: nit: small typo

marisbest2 (approved):

PR #1428, opened 2024-10-10
Enable mail order options for Sana Care

jknapp25:
We want to enable a few mail order options for Sana Care.

Todo:
- Get actual walmart mail order logo

Todo as follow-up:
- Get rounded variant of costplus logo
- Get walmart logo with blue background and round


marisbest2:
Is there a neutron version for this?

And will weird things happen if the walmart envvar isnt set?

jknapp25:
> Is there a neutron version for this?
> 
> And will weird things happen if the walmart envvar isnt set?

neutron version? like for double-checking?

got the walmart id from rado

PR #1445, opened 2024-10-15
Test: Swap Cash Price with Distance search

jknapp25:
Simple test to swap the order of cash price with distance pharmacy searching to see how it impacts conversion

### DIFF
--- a/apps/patient/src/components/PickupOptions.tsx
+++ b/apps/patient/src/components/PickupOptions.tsx
@@ -187,11 +187,6 @@ export const PickupOptions = ({
           {t.showMore}
         </Button>
       ) : null}
-      {!loadingMore && showingAllPharmacies ? (
-        <Text color="gray.500" textAlign="center">
-          {t.showingAll}
-        </Text>
-      ) : null}
     </VStack>
   );
 };
--- a/apps/patient/src/views/Pharmacy.tsx
+++ b/apps/patient/src/views/Pharmacy.tsx
@@ -100,7 +100,7 @@ export const Pharmacy = () => {
 
   // sorting
   type SortBy = 'price' | 'distance';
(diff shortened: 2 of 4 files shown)


apps/patient/src/components/PickupOptions.tsx
```diff
@@ -187,11 +187,6 @@ export const PickupOptions = ({
           {t.showMore}
         </Button>
       ) : null}
-      {!loadingMore && showingAllPharmacies ? (
-        <Text color="gray.500" textAlign="center">
-          {t.showingAll}
```
jknapp25, line 192: patients are clicking this non-clickable text to get more pharmacies, jomi suggested removing

ritabulman (approved):

billoran8 (commented):
any concerns about having changed the package lock? doesn't look like anything here should have touched that - are we all on the same version of node and npm?

billoran8 (approved):

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -757,45 +756,31 @@ export const Pharmacy = () => {
...
-                    }}
-                    border="2px"
-                    borderColor="gray.100"
-                    backgroundColor="white"
-                    onClick={() => setSortBy('distance')}
-                    borderRadius="xl"
-                  >
-                    Distance
-                  </Button>
+                  {(['distance', 'price'] as SortBy[]).map((sort) => (
```
marisbest2, line 759: note: you actually want `as const` rather than `as SortBy[]`. 

If you add a 3rd `SortBy` or remove one of them, this will still typecheck. But if you use `as const` and `distance` or `price` is no longer in `SortBy` then it'll typefail

PR #1446, opened 2024-10-16
Discount card test follow-up

jknapp25:
- restrict frontend to only show 1 coupon - saw one order with this, i believe it was reroute to the same pharmacy
- hide toggle for multirx - backend was limiting this already by returning no pharmacies
- don't show prices for glp1 - this was an original requirement that i missed
- don't register `cash_price_displayed` action unless toggled to price search - need this change now that we're not defaulting to cash price search

### DIFF
--- a/apps/patient/src/components/Coupons.tsx
+++ b/apps/patient/src/components/Coupons.tsx
@@ -27,11 +27,10 @@ export const Coupons = () => {
   return (
     <VStack w="full" alignItems="stretch" spacing={4}>
       <Heading as="h4" size="md">
-        {discountCards.length > 1 ? 'Coupons' : 'Coupon'}
+        Coupon
       </Heading>
-      {discountCards.map((card) => (
-        <Coupon key={card.id} coupon={card} />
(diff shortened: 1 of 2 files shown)


apps/patient/src/components/Coupons.tsx
```diff
@@ -27,11 +27,10 @@ export const Coupons = () => {
...
     <VStack w="full" alignItems="stretch" spacing={4}>
       <Heading as="h4" size="md">
-        {discountCards.length > 1 ? 'Coupons' : 'Coupon'}
+        Coupon
       </Heading>
-      {discountCards.map((card) => (
-        <Coupon key={card.id} coupon={card} />
-      ))}
+      {/* Show one coupon only */}
+      <Coupon coupon={discountCards[0]} />
```
marisbest2, line 33: Note: we should maybe enforce this on the backend (later)

ritabulman (approved):

marisbest2 (approved):

PR #1451, opened 2024-10-18
Swap distance with cash price

jknapp25:
Returning cash price search to be the first that surfaces in pharmacy selection


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -674,27 +674,6 @@ export const Pharmacy = () => {
     }
   };
 
-  // TODO: remove after we start storing prices (<1wk)
```
jknapp25, line 677: don't need this stuff anymore

samkotlove (approved):

PR #1454, opened 2024-10-21
Add Event when adding a Draft Prescription

captDaylight:
When the draft is added to the order, we will emit the event `photon-draft-prescription-created` with the draft prescription information.


packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -115,13 +148,7 @@ export const AddPrescriptionCard = (props: {
       const addDraftPrescription = async () => {
         props.actions.updateFormValue({
           key: 'draftPrescriptions',
-          value: [
-            ...(props.store.draftPrescriptions?.value || []),
-            {
-              id: String(Math.random()),
```
marisbest2, line 121: Do we not use this?

marisbest2 (approved):

PR #1457, opened 2024-10-24
Bump Typescript to latest (5.6.3) and Nx to v20

captDaylight:
We're on nx 20 and typescript 5 🥳 

<del>The issue with Nx going from v19 to v20 (we were at 15 so an improvement for sure) is that it's expecting a package version that doesn't exist. I made a ticket for it https://github.com/nrwl/nx/issues/28624.</del>

https://www.notion.so/photons/Upgrade-client-repo-TS-Node-Nx-1188bfbbf4ec80a49da4c5e0344b4764?pvs=4

this is gonna blow up your local environment, so...

```bash
# first, try
# use a more recent node
nvm use 20
# Install new dependencies
npm install   
# Run migrations
npx nx migrate --run-migrations


# if that doesn't work nuke
# blow away modules and dist
find . -name "dist" -type d -prune -exec rm -rf '{}' +
find . -name "node_modules" -type d -prune -exec rm -rf '{}' +
# clear cache
npm cache clean --force
# pray it works
# install
npm i
npx nx migrate --run-migrations
```


apps/app/src/index.tsx
```diff
@@ -10,8 +10,7 @@ import theme from './configs/theme';
 import { App } from './App';
 
 import pkg from '../package.json';
-
-import('@photonhealth/elements').catch(() => {});
+import '@photonhealth/elements';
```
marisbest2, line 13: Is this safe?

packages/elements/src/stores/catalog.ts
```diff
@@ -72,7 +72,7 @@ const createCatalogStore = () => {
     const { data, errors } = await client.clinical.catalog.getCatalogs({
       fragment: CatalogFieldsMap
     });
-    setStore('catalogs', {
+    return setStore('catalogs', {
```
marisbest2, line 75: Whats the deal here?

packages/elements/tsconfig.json
```diff
@@ -10,7 +10,9 @@
     "jsx": "preserve",
     "jsxImportSource": "solid-js",
     "resolveJsonModule": true,
+    "skipLibCheck": true,
```
marisbest2, line 13: you might want to turn this back on

packages/sdk/package.json
```diff
@@ -31,8 +31,10 @@
   },
   "exports": {
     ".": {
+      "types": "./dist/lib.d.ts",
       "import": "./dist/lib.mjs",
-      "require": "./dist/lib.js"
+      "require": "./dist/lib.js",
+      "default": "./dist/lib.js"
```
marisbest2, line 37: Whats this?

marisbest2 (approved):
Looks reasonable to me!

packages/elements/src/stores/catalog.ts
captDaylight: ah... I could have swore this was giving me an error, I'm going crazy reverted

packages/elements/tsconfig.json
captDaylight: I agree this isn't 100% ideal, but I'm getting super unrelated node type errors that are blocking. let's just keep in mind that we want to switch these back to false at some point

packages/sdk/package.json
captDaylight: scattershot to make it work

PR #1459, opened 2024-10-28
remove console logs and unnecessary comment

billoran8:
was heavily testing some stuff and didn't want to create prescriptions while verifying behavior but forgot to uncomment it out in the larger pr. whoops.

### DIFF
--- a/packages/elements/src/photon-multirx-form/photon-prescribe-workflow.tsx
+++ b/packages/elements/src/photon-multirx-form/photon-prescribe-workflow.tsx
@@ -460,9 +460,6 @@ export function PrescribeWorkflow(props: PrescribeProps) {
     // if we have alerts we'll want the prescriber to acknowledge them
     // first, unless we're overriding them
 
-    console.log('overrideScreenAlerts');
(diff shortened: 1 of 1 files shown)


packages/elements/src/photon-multirx-form/photon-prescribe-workflow.tsx
```diff
@@ -460,17 +460,15 @@ export function PrescribeWorkflow(props: PrescribeProps) {
     // if we have alerts we'll want the prescriber to acknowledge them
     // first, unless we're overriding them
 
-    console.log('overrideScreenAlerts');
-    console.log(overrideScreenAlerts());
```
samkotlove, line 464: we should add linting rules on this

samkotlove (approved):

packages/elements/src/photon-multirx-form/photon-prescribe-workflow.tsx
billoran8: 100% agree, i'll make a ticket

PR #1461, opened 2024-10-29
Fix issue with combine orders with clinical alerting

billoran8:
updating some colors; adding some borders; fixing issue with skipping alerts when there's a recent order; also adding some datadog dispatches to get us metrics about what causes people to close or acknoweldge alerts (though i will also be adding similar metrics to the backend)

### DIFF
--- a/packages/components/src/particles/Banner/index.tsx
+++ b/packages/components/src/particles/Banner/index.tsx
@@ -59,7 +59,7 @@ export default function Banner(props: BannerProps) {
 
   return (
     <Show when={isVisible()}>
-      <div class={bannerClasses(props.withBorder ?? false)}>
+      <div class={bannerClasses()}>
         <div class="flex w-full">
           <Show when={!props.withoutIcon}>
             <div class="flex-shrink-0 mr-2">
--- a/packages/components/src/particles/Banner/index.tsx
+++ b/packages/components/src/particles/Banner/index.tsx
@@ -31,19 +31,19 @@ export default function Banner(props: BannerProps) {
     }
   };
 
-  const bannerClasses = (includeBorder: boolean) => {
+  const bannerClasses = () => {
     return clsx('flex place-items-start justify-between gap-2 py-3 px-3 sm:px-4 rounded-lg', {
-      'border-2 rounded border-solid': includeBorder,
+      'border-2 rounded border-solid': props.withBorder,
       'text-blue-600 bg-blue-50': props.status === 'info',
-      'border-blue-600': props.status === 'info' && includeBorder,
+      'border-blue-600': props.status === 'info' && props.withBorder,
       'text-red-600 bg-red-50': props.status === 'error',
-      'border-red-600': props.status === 'error' && includeBorder,
+      'border-red-600': props.status === 'error' && props.withBorder,
       'text-green-600 bg-green-50': props.status === 'success',
-      'border-green-600': props.status === 'success' && includeBorder,
+      'border-green-600': props.status === 'success' && props.withBorder,
       'text-yellow-600 bg-yellow-50': props.status === 'warning',
-      'border-yellow-600': props.status === 'warning' && includeBorder,
+      'border-yellow-600': props.status === 'warning' && props.withBorder,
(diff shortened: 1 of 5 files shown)


packages/components/src/particles/Banner/index.tsx
```diff
@@ -46,13 +53,13 @@ export default function Banner(props: BannerProps) {
       'text-red-600': props.status === 'error',
       'text-green-600': props.status === 'success',
       'text-yellow-600': props.status === 'warning',
-      'text-gray-600': props.status === 'suggestion'
+      'text-slate-600': props.status === 'suggestion'
```
billoran8, line 56: there's only one section using this setting and it's for the popup when acknowleding alerts. making it darker to help it pop a lil bit more.

packages/components/src/systems/ScreeningAlerts/ScreeningAlert.tsx
```diff
@@ -28,7 +28,7 @@ const STATUS_TO_CLASS_MAP: Record<string, BannerStatus> = {
  */
 const getStatus = (severity: string, type: string): BannerStatus => {
   if (type === 'ALLERGEN') {
-    return 'suggestion';
+    return 'error';
```
billoran8, line 31: we want allergens red and not grey

packages/components/src/systems/ScreeningAlerts/ScreeningAlertByEntity.tsx
```diff
@@ -64,7 +64,7 @@ export const ScreeningAlertByEntity = (props: {
   otherAlertsByEntity: AlertsForEntity[];
 }) => {
   return (
-    <Banner withoutIcon status="suggestion">
+    <Banner withoutIcon withBorder status="suggestion">
```
billoran8, line 67: we want the alert in the acknowledgement window ot have borders as well

packages/elements/src/photon-multirx-form/photon-prescribe-workflow.tsx
```diff
@@ -522,6 +521,8 @@ export function PrescribeWorkflow(props: PrescribeProps) {
               }
             });
 
+            combineOrSubmit();
```
billoran8, line 524: was calling the wrong function; `combineOrSubmit` delegates to appropriate downstream functions like `combine`; `submitForm` is the nuclear option and i was taking it.

packages/components/src/systems/ScreeningAlerts/ScreeningAlertAcknowledgement.tsx
```diff
@@ -38,6 +39,31 @@ export function ScreeningAlertAcknowledgementDialog(
...
+      props.alerts.reduce((acc, value, index) => {
+        acc[index] = value.description;
+        return acc;
+      }, {} as Record<string, string>),
+      ref
+    );
+    props.onIgnoreWarningAndCreateAnyway();
+  }
+
+  function onRevisitPrescriptions() {
```
billoran8, line 56: here's an event that'll show us which warnigns are encouraging people to cancel creation and modify the prescriptions

packages/components/src/particles/Banner/index.tsx
```diff
@@ -30,13 +31,19 @@ export default function Banner(props: BannerProps) {
     }
   };
 
-  const bannerClasses = () => {
+  const bannerClasses = (includeBorder: boolean) => {
```
marisbest2, line 34: Just for curiosity, why not just use `props.withBorder`?

marisbest2 (approved):

packages/components/src/particles/Banner/index.tsx
billoran8: there is absolutely no reason why! i just tend to like doing things this way because in my head it makes em easier to test this function in isolation if we were ever that psychotic

billoran8: since it's not being tested int hat way and likely never will lemme just check the props and remove some complesxity

captDaylight (approved):

PR #1466, opened 2024-10-31
persist url search params when session expires

ritabulman:
for this ticket: https://www.notion.so/photons/URL-params-not-persisted-on-redirect-1288bfbbf4ec8055b43bf981aa95df6e

session expires while url contains search params: user gets returned to url with search params after logging back in
user goes to the login page: user is direct to `/prescriptions` after login
user is logged out and tries to access url with search params: user is directed to that url with search params after login
session expires while patient is at `/patients`: user is returned to `/patients` after logging back in

### DIFF
--- a/apps/app/src/views/components/Auth.tsx
+++ b/apps/app/src/views/components/Auth.tsx
@@ -30,7 +30,10 @@ export const Auth = (props: AuthProps) => {
     );
 
   return (
-    <Button colorScheme="blue" onClick={() => login({ appState: { returnTo } })}>
+    <Button
+      colorScheme="blue"
+      onClick={() => login({ appState: { returnTo: window.location.origin + returnTo } })}
+    >
       Log in
     </Button>
   );
--- a/apps/app/src/views/components/Auth.tsx
+++ b/apps/app/src/views/components/Auth.tsx
@@ -30,10 +30,7 @@ export const Auth = (props: AuthProps) => {
     );
 
   return (
-    <Button
-      colorScheme="blue"
-      onClick={() => login({ appState: { returnTo: window.location.origin + returnTo } })}
-    >
+    <Button colorScheme="blue" onClick={() => login({ appState: { returnTo } })}>
       Log in
     </Button>
   );
--- a/apps/app/src/views/routes/Login.tsx
+++ b/apps/app/src/views/routes/Login.tsx
@@ -41,7 +41,7 @@ export const Login = () => {
     return <Navigate to="/" replace />;
   }
 
-  const from = `${(location.state?.from?.pathname || '/') + (location.state?.from?.search || '')}`;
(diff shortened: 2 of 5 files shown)


marisbest2:
How do we test this for multiple cases? Does this code ever work as expected?

apps/app/src/views/routes/Login.tsx
```diff
@@ -41,7 +41,7 @@ export const Login = () => {
     return <Navigate to="/" replace />;
   }
 
-  const from = `${(location.state?.from?.pathname || '/') + (location.state?.from?.search || '')}`;
+  const from = `${(location.pathname || '/') + (location.search || '')}`;
```
captDaylight, line 44: maybe use the template tag without the `+`
```js
`${location.pathname || '/'}${location.search || ''}`;
```

apps/app/src/views/components/Auth.tsx
```diff
@@ -30,7 +30,10 @@ export const Auth = (props: AuthProps) => {
     );
 
   return (
-    <Button colorScheme="blue" onClick={() => login({ appState: { returnTo } })}>
+    <Button
+      colorScheme="blue"
+      onClick={() => login({ appState: { returnTo: window.location.origin + returnTo } })}
```
captDaylight, line 35: can you make this a template literal? 
Does it need the origin?

apps/app/src/views/routes/Main.tsx
```diff
@@ -82,17 +82,15 @@ export const Main = () => {
...
+    const pathname = location.pathname;
     const queryString = query.toString() ? `?${query.toString()}` : '';
 
     return (
       <Navigate
         to={`/login${queryString}`}
         state={{
-          from: {
-            ...location,
-            ...(pathname ? { pathname } : {})
```
captDaylight, line 94: anywhere else that from.pathname is being used?

packages/sdk/src/auth.ts
```diff
@@ -151,7 +151,13 @@ export class AuthManager {
       }
     }
     if (!token) {
-      await this.authentication.loginWithRedirect(opts);
+      const redirectOpts: RedirectLoginOptions = {
+        ...opts,
+        appState: {
+          returnTo: window.location.pathname + window.location.search
```
captDaylight, line 157: template literal plz

captDaylight: also returnTo is relative here, but above it included the origin

apps/app/src/views/components/Auth.tsx
ritabulman: I removed the origin and it still works ty

packages/sdk/src/auth.ts
ritabulman: removed origin above

captDaylight (approved):

PR #1496, opened 2024-12-02
shortlink action

marisbest2:
Shortlink action. See https://github.com/Photon-Health/lambdas/pull/1823 and https://www.notion.so/photons/Track-shortlink-opens-1508bfbbf4ec802da387e0af63a44109

### DIFF
--- a/apps/patient/src/views/Main.tsx
+++ b/apps/patient/src/views/Main.tsx
@@ -40,7 +40,6 @@ export const Main = () => {
   const isDemo = searchParams.get('demo');
   const orderId = searchParams.get('orderId');
   const phone = searchParams.get('phone');
-  const useV2 = searchParams.get('v2');
 
   const [order, setOrder] = useState<Order | undefined>(isDemo ? demoOrder : undefined);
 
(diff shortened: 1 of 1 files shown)


apps/patient/src/views/Main.tsx
```diff
@@ -55,6 +54,22 @@ export const Main = () => {
   const location = useLocation();
   const [faqModalIsOpen, setFaqModalIsOpen] = useState(false);
 
+  useEffect(() => {
+    if (location.pathname === '/' && token) {
```
billoran8, line 58: why do we only want to emit the event if the path is a slash? wouldn't we rather have this useEffect run on every page load (maybe add the url in as metadata lol?)

billoran8: i would just add a comment explaining why this is necessary, this seems like one of those innocuous things that might be easy to misunderstand/delete then subsequently polute our data

billoran8 (approved):

apps/patient/src/views/Main.tsx
```diff
@@ -55,6 +54,27 @@ export const Main = () => {
   const location = useLocation();
   const [faqModalIsOpen, setFaqModalIsOpen] = useState(false);
 
+  useEffect(
+    function triggerDatadogShortlinkOpenEvent() {
```
billoran8, line 58: i love this name (and the comment!)

billoran8 (approved):
yuh

PR #1499, opened 2024-12-04
High res pharmacy logos 🤓

jknapp25:
A bunch of our pharmacy logos were low res, swapping those out so folks like me can see em!


billoran8 (approved):
fine with this for now but any thought about storing files on s3 or some other location? feels weird to use github to host images imo

ritabulman (approved):

PR #1501, opened 2024-12-04
Display default info for mail order pharmacy, move tracking link into the pharmcy info section

mrrsnn:
We currently don't have any default behavior for mail order pharmacies that don't have any information in the `PHARMACY_BRANDING` constant. 

## Before
Screenshot 2024-12-04 at 10 31 02 AM

## After
Screenshot 2024-12-04 at 12 11 13 PM
Screenshot 2024-12-04 at 12 26 35 PM
Screenshot 2024-12-04 at 12 11 23 PM

### DIFF
--- a/apps/patient/src/views/StatusV2.tsx
+++ b/apps/patient/src/views/StatusV2.tsx
@@ -1,4 +1,4 @@
-import { Box, Button, Container, Heading, Link, Text, VStack } from '@chakra-ui/react';
+import { Box, Button, Container, Heading, VStack } from '@chakra-ui/react';
 import { getSettings } from '@client/settings';
 import queryString from 'query-string';
 import { useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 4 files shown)


apps/patient/src/views/StatusV2.tsx
```diff
@@ -255,10 +223,18 @@ export const StatusV2 = () => {
               </Heading>
               <Card>
                 <VStack w="full" spacing={0}>
-                  {pharmacyInfo}
+                  {displayPharmacy && (
```
marisbest2, line 226: If `displayPharmacy` is empty (not sure when that would happen, but 🤷) we shouldn't show this whole Pharmacy block (i.e. don't show the `Heading` and the `Card`)

mrrsnn: ah yeah good catch

marisbest2 (approved):

PR #1503, opened 2024-12-05
add call pharmacy button

ritabulman:
add button to order status page to call pharmacy when an order is no longer rerouteable


apps/patient/src/views/StatusV2.tsx
```diff
@@ -42,6 +42,11 @@ export const StatusV2 = () => {
     window.open(url);
   };
 
+  const handleCallPharmacy = () => {
+    if (!pharmacy?.name || !pharmacy?.phone) return;
```
jknapp25, line 46: why is name needed?

apps/patient/src/views/StatusV2.tsx
```diff
@@ -158,6 +163,22 @@ export const StatusV2 = () => {
     </Button>
   );
 
+  const callPharmacyButton = (
+    <Button
+      mt={2}
```
jknapp25, line 168: Vstack spacing should take care of vertical margin

apps/patient/src/views/StatusV2.tsx
```diff
@@ -158,6 +163,22 @@ export const StatusV2 = () => {
...
+      mx="auto"
+      size="md"
```
jknapp25, line 170: it's default medium right?

jknapp25 (approved):

PR #1504, opened 2024-12-06
Template Quantity can allows for values between 0 and 1 also

captDaylight:
https://www.notion.so/photons/Template-Editor-Doesn-t-allow-for-Fractional-Amounts-1308bfbbf4ec80e091c9fdbbde53609d?pvs=4

### DIFF
--- a/apps/app/src/views/routes/Settings/components/templates/TemplateForm/utils.ts
+++ b/apps/app/src/views/routes/Settings/components/templates/TemplateForm/utils.ts
@@ -9,7 +9,12 @@ export const TEMPLATE_SCHEMA = yup.object({
     })
     .required('Please select a treatment...'),
   dispenseAsWritten: yup.boolean().nullable().optional(),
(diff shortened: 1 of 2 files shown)


apps/app/src/views/routes/Settings/components/templates/TemplateForm/utils.ts
```diff
@@ -9,7 +9,12 @@ export const TEMPLATE_SCHEMA = yup.object({
     })
     .required('Please select a treatment...'),
   dispenseAsWritten: yup.boolean().nullable().optional(),
-  dispenseQuantity: yup.number().min(1, 'Must be 1 or more...').default(1).nullable().optional(),
```
billoran8, line 12: the fact that this specific check leads me to believe that there might be some downstream impacts here - any concerns?

captDaylight: we have it working like this in solidjs elements for months now, this was just incorrectly implemented in the react side

billoran8 (approved):

PR #1513, opened 2024-12-10
🍴 Price v Insurance Fork

jknapp25:
We're placing a decision point after readyBy and before pharmacy selection to determine if the patient actually wants to pay with cash or insurance. This test will likely result in fewer price selections, but should help us gauge intent of the patient better. Also might increase % adjudicated 🤞.

**New Payment Method view**

image

**How it looks on /pharmacy**

image

### DIFF
--- a/apps/patient/src/components/PharmacyFilters.tsx
+++ b/apps/patient/src/components/PharmacyFilters.tsx
@@ -1,4 +1,5 @@
 import { Button, HStack, Text } from '@chakra-ui/react';
+import { useOrderContext } from '../views/Main';
 
 interface PharmacyFiltersProps {
   enableOpenNow: boolean;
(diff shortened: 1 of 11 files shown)


captDaylight:
can me get a screenshot

apps/patient/src/components/PharmacyFilters.tsx
```diff
@@ -37,8 +44,22 @@ export const PharmacyFilters = ({
...
+          size="sm"
+          borderRadius="lg"
+          variant="filter"
+          isActive={enablePrice}
+          onClick={() => {
+            if (setEnablePrice) setEnablePrice(false);
+            setEnablePrice(!enablePrice);
+          }}
+        >
+          Price
```
captDaylight, line 60: (throw away comment) I thought we were going to consider calling this "coupon" at some point?

jknapp25: oh we do call it a coupon on the status page, but upfront it's simplified to just a price filter

captDaylight (approved):

PR #1525, opened 2024-12-12
Fix incorrect text on order placed sub-header

jknapp25:
Also worked on spacing, text color/weight, justification, etc. to reflect design spec.

image

### DIFF
--- a/apps/patient/src/components/statusV2/Step.tsx
+++ b/apps/patient/src/components/statusV2/Step.tsx
@@ -9,36 +9,18 @@ export interface StepProps {
 }
 
 export const Step = ({ color, icon, complete, iconProps }: StepProps) => {
+  const uiColor = complete
+    ? color === 'primary'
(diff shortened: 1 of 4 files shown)


apps/patient/src/components/statusV2/Step.tsx
```diff
@@ -9,36 +9,18 @@ export interface StepProps {
 }
 
 export const Step = ({ color, icon, complete, iconProps }: StepProps) => {
+  const uiColor = complete
```
billoran8, line 12: nit: nested ternaries like this are so hard for me to digest! maybe extract this into a `getUiColor` function and refactor it into something a lil bit easier to read?

billoran8: gj removing that duplicated logic in any case though

billoran8 (approved):

PR #1528, opened 2024-12-12
Load patient data and add to formStore for the prescribe workflow web component 

mrrsnn:
To avoid any confusion between internal and external patient ids used in query params and passed down as props, we will load the patient data into the patient store, then update the forms store patient value with the new data (already done in PatientCard) so that the source of truth for id is in `formStore.patient.value.id`

To avoid lingering stale data between sessions that use the same form store, we will reset the form store onCleanup in the `photon-prescribe-workflow-component`

### DIFF
--- a/packages/elements/src/photon-multirx-form/photon-prescribe-workflow-component.tsx
+++ b/packages/elements/src/photon-multirx-form/photon-prescribe-workflow-component.tsx
@@ -1,20 +1,9 @@
 import { RecentOrders } from '@photonhealth/components';
 import { customElement } from 'solid-element';
 import { createFormStore } from '../stores/form';
-import { PatientStore } from '../stores/patient';
 import { PrescribeProps, PrescribeWorkflow } from './photon-prescribe-workflow';
-import { createEffect, onMount } from 'solid-js';
-import { usePhoton } from '../context';
 
 const Component = (props: PrescribeProps) => {
-  const client = usePhoton();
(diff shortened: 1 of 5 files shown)


captDaylight:
success criteria:

- `patient-id` param accepts both internal and external ids
- you cannot submit order with stale patient data from a previous order

packages/elements/src/photon-multirx-form/photon-prescribe-workflow-component.tsx
```diff
@@ -1,9 +1,20 @@
...
 import { PrescribeProps, PrescribeWorkflow } from './photon-prescribe-workflow';
+import { createEffect, onMount } from 'solid-js';
+import { usePhoton } from '../context';
 
 const Component = (props: PrescribeProps) => {
+  const client = usePhoton();
+  const { actions: patientActions, store: patientStore } = PatientStore;
+  onMount(() => {
+    if (props.patientId) {
+      patientActions.getSelectedPatient(client!.getSDK(), props.patientId);
```
captDaylight, line 14: is there not logic somewhere that's already doing this? we're presumably already doing this since the patient card already loads if you pass in a patient id

captDaylight:
my understanding was that previously the source of truth was `formStore.patient.value.id` 

The problem is that we had stale data, the issue really is more that we need to roll back my change (which you did in rxMutation and orderMutation) but also prevent the potential of stale data in the formStore.

captDaylight (approved):

PR #1535, opened 2024-12-19
Add provider and date to duplicate order modal

captDaylight:
[Ref this ticket](https://www.notion.so/photons/Investigate-Duplicate-Orders-Potentially-Update-Logic-1528bfbbf4ec80b69129ea041681e9f6?pvs=4)

I’ve confirmed that the duplicate modal is displaying between different providers with the same organization.
My thinking is to add the provider name and date to the modal to make it very clear that it was prescribed perhaps by someone else and when.

Before:
Screen Shot 2024-12-19 at 4 08 24 PM

After: 
Screen Shot 2024-12-19 at 4 26 26 PM


mrrsnn (approved):
LGTM

PR #1588, opened 2025-02-18
Hide Glp1 prices temporarily

jknapp25:
Miscommunication w/ Sesame. Recalibrating, then we'll relaunch. 
Bit of a setback but we're marching on, lot's of refinements to pricing on-deck to test right now anyway.

### DIFF
--- a/apps/patient/src/views/ReadyBy.tsx
+++ b/apps/patient/src/views/ReadyBy.tsx
@@ -32,6 +32,7 @@ import { FixedFooter, PoweredBy } from '../components';
 import { text as t } from '../utils/text';
 import { useOrderContext } from './Main';
 import { RxLightningBolt } from 'react-icons/rx';
+import { isGLP } from '../utils/isGLP';
 
 const checkDisabled = (option: string): boolean => {
   const currentTime = dayjs();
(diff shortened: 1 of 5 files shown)


apps/patient/src/views/ReadyBy.tsx
```diff
@@ -92,8 +93,10 @@ export const ReadyBy = () => {
       readyByTime
     });
 
+    const containsGLP = flattenedFills.some((fill) => isGLP(fill.treatment.name));
+
     // Redirect to payment method selection if applicable
-    if (isOrgWithCouponsEnabled(order.organization.id) && !isMultiRx) {
+    if (isOrgWithCouponsEnabled(order.organization.id) && !isMultiRx && !containsGLP) {
```
billoran8, line 99: this seems to be a Very Important conditional and looks to be constantly expanding - we may wanna rethink this! (somethign tod o later)

billoran8 (approved):

apps/patient/src/views/ReadyBy.tsx
jknapp25: spreading the feature flag seeds are we?

billoran8: u know how i doooooo

billoran8: 🌱 🌱 🌱

PR #1590, opened 2025-02-20
Order exceptions

ritabulman:
grab exceptions on order object and display updates to header/status page if exceptions are present

### DIFF
--- a/apps/patient/src/components/InsuranceAlert.tsx
+++ b/apps/patient/src/components/InsuranceAlert.tsx
@@ -31,15 +31,15 @@ export const InsuranceAlert = ({ exceptions }: InsuranceAlertProps) => {
       <Card
         bgColor="white"
         border="1px solid"
-        borderColor="orange.500"
+        borderColor={`${details.color}.500`}
         borderRadius="2xl"
-        background="#FFFAEB"
+        background={details.color === 'orange' ? '#FFFAEB' : '#EFF8FF'}
         mx={{ base: -3, md: undefined }}
-        color="orange.500"
+        color={`${details.color}.500`}
       >
         <CardBody p={3}>
           <HStack>
-            <Icon color="orange.500" as={FiInfo} fontWeight="bold" />
+            <Icon color={`${details.color}.500`} as={FiInfo} fontWeight="bold" />
             <Text fontWeight="semibold">{details?.text}</Text>
           </HStack>
         </CardBody>
--- a/apps/patient/src/components/InsuranceAlert.tsx
+++ b/apps/patient/src/components/InsuranceAlert.tsx
@@ -1,22 +1,32 @@
-import React from 'react';
 import { Card, CardBody, HStack, Icon, Text, VStack } from '@chakra-ui/react';
+import { OrderExceptionType } from 'apps/patient/src/__generated__/graphql';
+import { FiInfo } from 'react-icons/fi';
 
-function insuranceAlertText(props: OrderStatusHeaderProps) {
-  if (props.exception === 'PHARMACY_NEEDS_INSURANCE_INFO') {
-    return 'Please bring your insurance card to pick up your prescription.';
+function getInsuranceAlertDetails(exceptions: OrderExceptionType[]) {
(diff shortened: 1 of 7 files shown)


apps/patient/src/views/StatusV2.tsx
```diff
@@ -201,6 +202,12 @@ export const StatusV2 = () => {
     [fulfillments, order.fulfillment]
   );
 
+  const exception = order.exceptions[0]
```
jknapp25, line 205: what's `order.exceptions[0]` ?

apps/patient/src/views/StatusV2.tsx
```diff
@@ -218,17 +225,12 @@ export const StatusV2 = () => {
           <Container py={6}>
             <VStack spacing={4} width="full" alignItems="stretch">
               <HolidayAlert>Holiday may affect pharmacy hours.</HolidayAlert>
+              <InsuranceAlert exceptions={order.exceptions} />
```
jknapp25, line 228: seeing as the alert only displays one, can we pass in one exception here?

apps/patient/src/components/InsuranceAlert.tsx
```diff
@@ -0,0 +1,49 @@
+import { Card, CardBody, HStack, Icon, Text, VStack } from '@chakra-ui/react';
+import { OrderExceptionType } from 'apps/patient/src/__generated__/graphql';
+import { FiInfo } from 'react-icons/fi';
+
+function getInsuranceAlertDetails(exceptions: OrderExceptionType[]) {
+  if (exceptions.some((e) => e === 'PHARMACY_NEEDS_INSURANCE_INFO')) {
```
jknapp25, line 6: maybe just a map would simplify like...

{
   PHARMACY_NEEDS_INSURANCE_INFO: {
      text: 'Please bring your insurance card to pick up your prescription.',
      color: 'blue'
    },
    PHARMACY_DOES_NOT_ACCEPT_INSURANCE: {
      text: 'Your pharmacy does not accept your insurance on file. Use a discount card to pay cash price or change your pharmacy.',
      color: 'orange'
    }
}

apps/patient/src/views/StatusV2.tsx
ritabulman: this is looking for any exception at the order level (which right now would just be one), which would take precedence over the pharmacy closed attribute and fulfillment level exceptions

jknapp25: roger that! do you mind adding a lil comment to denote that?

apps/patient/src/components/statusV2/Header.tsx
```diff
@@ -22,12 +22,37 @@ export interface OrderStatusHeaderProps {
...
+    | 'PHARMACY_UNREACHABLE'
+    | 'EXTERNAL_TRANSFER'
+    | 'PHARMACY_NEEDS_INSURANCE_INFO'
+    | 'PHARMACY_DOES_NOT_ACCEPT_INSURANCE';
   pharmacyEstimatedReadyAt?: Date;
   patientDesiredReadyAt?: Date | 'URGENT';
 }
 
-function headerText(status: OrderStatusHeaderProps['status']) {
+function headerText(props: OrderStatusHeaderProps) {
```
jknapp25, line 36: nit: exception and status map here might also be nice

apps/patient/src/components/statusV2/Header.tsx
```diff
@@ -22,12 +22,37 @@ export interface OrderStatusHeaderProps {
...
     | 'RX_CLARIFICATION'
     | 'OTC'
-    | 'MEDICAL_DEVICE';
+    | 'MEDICAL_DEVICE'
+    | 'DEMOGRAPHIC_MISMATCH'
+    | 'PHARMACY_CLOSED'
+    | 'PHARMACY_UNREACHABLE'
+    | 'EXTERNAL_TRANSFER'
+    | 'PHARMACY_NEEDS_INSURANCE_INFO'
+    | 'PHARMACY_DOES_NOT_ACCEPT_INSURANCE';
```
jknapp25, line 31: do we have these exception types centralized that we could reuse something like ExceptionType here?

PR #1599, opened 2025-02-24
Fix stale patient data being sent with new orders

captDaylight:
https://photonhealth.slack.com/archives/C03TRQ4NLE5/p1739919711496799

Stale customer data can be sent in orders while the new patient information is being fetched. Previous attempts [here](https://github.com/Photon-Health/client/pull/1518/files) and here with [the action reset](https://github.com/Photon-Health/client/pull/1528/files).

This also seems to fix an issue we've been seeing[ in our data](https://www.notion.so/photons/We-get-warnings-Fill-s-for-one-or-more-prescriptions-do-not-belong-to-patient-pat_-provided-on-t-19f8bfbbf4ec80aabdcceb79e69238d3?pvs=4):
Screenshot 2025-02-21 at 3 08 47 PM

### DIFF
--- a/packages/elements/index.html
+++ b/packages/elements/index.html
@@ -5,6 +5,26 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Photon Components</title>
     <script src="./src/index.ts" type="module"></script>
+
+    <script>
+      window.addEventListener('photon-order-created', (e) => {
+        console.log(e);
+        const parent = document.getElementById('parent');
+        const one = document.getElementById('one');
+        one.remove();
+
+        const two = document.createElement('photon-prescribe-workflow');
+        two.id = 'two';
+        two.patientId = 'pat_01J8T7B7WF4YSSQX526MR0S0RM';
+        two.pharmacyId = 'phr_01GA9HPXAF9BC9PC8XM25Q836N';
+        two.templateIds = 'tmp_01HF7H3J7J37458TEMT6R8JW84';
+        two.enableOrder = true;
+        two.hideTemplates = true;
+        setTimeout(() => {
+          parent.appendChild(two);
+        }, 1000);
+      });
+    </script>
   </head>
   <body style="margin: 0px">
     <photon-client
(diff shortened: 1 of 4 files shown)


packages/elements/index.html
```diff
@@ -5,6 +5,26 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Photon Components</title>
     <script src="./src/index.ts" type="module"></script>
+
```
billoran8, line 8: did we mean to commit this

captDaylight: ah no, didn't mean to commit, will remove

packages/elements/src/photon-multirx-form/photon-prescribe-workflow.tsx
```diff
@@ -580,8 +588,13 @@ export function PrescribeWorkflow(props: PrescribeProps) {
               />
               <Show
                 when={
-                  props.formStore.patient?.value?.address ||
-                  (props.formStore.patient?.value?.id && !props.enableOrder)
+                  // if patientId is passed in, we need to ensure it matches the patient id in our store
```
billoran8, line 591: is the user experience here just that when a prescriber has an old patient cached that they just will have to refresh the page?

billoran8: is there any sort of metric or tracking or event we can trigger to see how often this actually winds up happening?

mrrsnn (approved):

packages/elements/src/photon-multirx-form/photon-prescribe-workflow.tsx
captDaylight: no way for us to track if it's in someone else's application

captDaylight: no, they don't need to refresh the page, we just need to dump our cached data

billoran8: we should consider making an endpoint that can submit metrics to datadog from other people's aplications; thoughts?

PR #1617, opened 2025-03-06
Add Remo org settings/branding to photon settings file

mrrsnn:


billoran8 (approved):
future thgouht: let's upload images to s3! feels weird to have assets like this in a repo imo

PR #1635, opened 2025-03-13
Update URL Whitelist for SPA Clients in Developer Tab

mrrsnn:
### Client Whitelist Updating
- Updates the SPA Client whitelist URLs
- Automatically sanitizes out blank spaces between and around commas
- Validates that all comma separated entries are in URL format
- Also changed the inputs for client name, id, and secret so that they look visibly disabled. They already are disabled in production, but look editable! 

https://github.com/user-attachments/assets/79fa3006-f7b6-4852-aa68-5f35950c14fd


apps/app/src/views/routes/Settings/components/client-credentials/AppNameView.tsx
```diff
@@ -18,7 +18,7 @@ export const AppNameView = (props: AppNameViewProps) => {
     <FormControl id="app_name">
       <FormLabel>Application Name</FormLabel>
       <InputGroup variant="outline">
-        <Input value={name} isReadOnly placeholder={name} />
+        <Input value={name} disabled variant="filled" placeholder={name} />
```
billoran8, line 21: merge issue?

billoran8: nevermind

jknapp25, line 21: you can set read only style in chakra config if you want it in one place

apps/app/src/views/routes/Settings/components/client-credentials/SPAClientEditForm.tsx
```diff
@@ -0,0 +1,133 @@
...
+      id
+    }
+  }
+`);
+
+const spaClientValidationSchema = yup.object({
+  whiteListedUrls: yup
+    .string()
+    .required('White Listed URLs are required')
+    .test('urls', 'Must be comma separated valid URLs', (value) => {
```
billoran8, line 31: if we're doing front-end validation let's make it so we throw an exception if the URLs have any sort of query params, which we don't permit! just to save us from issues in teh future

apps/app/src/views/routes/Settings/components/client-credentials/AppNameView.tsx
jknapp25: also, the hover darkening made me think it was still editable

apps/app/src/views/routes/Settings/components/client-credentials/SPAClientEditForm.tsx
mrrsnn: Good idea!

billoran8 (approved):

PR #1643, opened 2025-03-17
Fix running components locally

jakefromphoton:
localhost:3000 had errors when running locally via `npx nx run components:start`

 - fix 401 error from not setting clinical-api env in sdk
 - use better test patientId

### DIFF
--- a/packages/components/src/systems/Client/index.tsx
+++ b/packages/components/src/systems/Client/index.tsx
@@ -1,4 +1,4 @@
-import { PhotonClient } from '@photonhealth/sdk';
+import { Env, PhotonClient } from '@photonhealth/sdk';
 import { createEffect, JSXElement } from 'solid-js';
 import { PhotonContext } from '../../context';
 import { PhotonClientStore } from '../../store';
@@ -21,6 +21,7 @@ type ClientProps = {
   context?: any;
   createStore?: typeof createStore;
   connection?: string;
+  env?: Env;
 };
 
 export default function Client(props: ClientProps) {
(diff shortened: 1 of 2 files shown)


packages/components/src/systems/Client/index.tsx
```diff
@@ -79,3 +81,20 @@ export default function Client(props: ClientProps) {
     </Provider>
   );
 }
+
+function getEnv(props: ClientProps): Env {
+  const isEnv = (env: Env) =>
+    props.domain?.includes(env) || props.audience?.includes(env) || props.uri?.includes(env);
```
billoran8, line 87: nit: i like using `??` instead for purity of intent - ?? is explicitly 'coalesce' instead of `||` being or

jakefromphoton: Makes sense to me - if `domain` is non-null and doesn't include `env`, then ?? will just return false and we can not bother checking the other props.

packages/components/src/systems/Client/index.tsx
```diff
@@ -79,3 +81,23 @@ export default function Client(props: ClientProps) {
     </Provider>
   );
 }
+
+function getEnv(props: ClientProps): Env {
+  const isEnv = (env: Env) =>
+    props.domain?.includes(env) ??
```
billoran8, line 87: uh, look at that purity of purpose! the intent is so clear!! i see EVERYTHING now

billoran8 (approved):

PR #1666, opened 2025-03-24
Fix CI error for npm publish

jakefromphoton:
Using fix for a known npm build issue documented here: 

https://github.com/vitejs/vite/discussions/15532#discussioncomment-8527272

### DIFF
--- a/packages/elements/package.json
+++ b/packages/elements/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@photonhealth/elements",
-  "version": "0.14.6",
+  "version": "0.14.10",
   "description": "",
   "main": "dist/index.js",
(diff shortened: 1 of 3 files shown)


packages/elements/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@photonhealth/elements",
-  "version": "0.14.6",
+  "version": "0.14.10",
```
captDaylight, line 3: why 6 => 10?

captDaylight (approved):

packages/elements/package.json
jakefromphoton: Didn't figure out how to trigger the job without the version bump for a bit (the check compares package.json version of previous commit)

PR #1668, opened 2025-03-24
Fix/med history expand

jakefromphoton:
* Mainly fixes clicking items with the same Treatment ID (e.g. 2+ prescriptions of "amoxicillin 500mg") expanding all rows with that Treatment ID
* Refactors
  *  type mappings, prop naming, code organization
  * Remove unused `active` and `comment` fields


jakefromphoton:
merged into https://github.com/Photon-Health/client/pull/1673

PR #1670, opened 2025-03-25
Disable med refill button

jakefromphoton:
- mainly due to z-index conflicts between IconButton, sticky grid header, and recent orders duplicate popup


whittle-photon (approved):
Thanks!

PR #1697, opened 2025-03-28
Remove Effective Date input from Prescription form

captDaylight:
Create Rx mutation doesn't change, just defaults to the same day it was created


jakefromphoton:
Does this remove the ability for prescribers to set the effectiveDate / doNotFillBeforeDate ? If so, any context to share on why we're removing this feature? Sounds like the backend mutation was ignoring this field.

PR #1699, opened 2025-03-28
Remove duplicate PhotonClientStore from elements

jakefromphoton:
Having two version of this, one in `packages/components` and one in `packages/elements`, is highly confusing. So, let's consolidate it and only have the one in components. 

This aligns with current thinking b/t Paul and I that `elements` should have minimal business logic, and be focused on exporting html web components for (1) customers who embed Photon functionality and (2) our clinical app.

* Updated imports across `elements`
* Migrated logic from the `elements` version into `components`:
  1. Auto login
  2. Redirect handler logic
  3. Removed createStore reactivity hack/workaround from 2 years ago  -- seems fine now


packages/components/src/store.ts
```diff
@@ -15,7 +15,11 @@ import jwtDecode from 'jwt-decode';
 import type { GraphQLFormattedError } from 'graphql';
 
 const defaultOnRedirectCallback = (appState?: any): void => {
-  window.location.replace(appState?.returnTo || window.location.pathname);
+  if (appState?.returnTo) {
+    window.location.replace(appState?.returnTo);
+  } else {
+    window.history.pushState(null, window.document.title, window.location.pathname);
+  }
```
jakefromphoton, line 22: took this from `elements`

packages/components/src/store.ts
```diff
@@ -209,6 +214,7 @@ export class PhotonClientStore {
         try {
           const result = await this.sdk.authentication.handleRedirect(url);
           defaultOnRedirectCallback(result?.appState);
+          await this.authentication.checkSession();
```
jakefromphoton, line 217: took this from `elements`

packages/components/src/store.ts
```diff
@@ -280,13 +286,16 @@ export class PhotonClientStore {
...
+      let permissions: Permission[] = [];
+      if (this.autoLogin || authenticated) {
+        // getAccessToken is triggering the SSO screen to appear, so for auto-login=false, we don't want to call it
+        try {
+          const token = await this.sdk.authentication.getAccessToken();
+          const decoded: { permissions: Permission[] } = jwtDecode(token);
+          permissions = decoded?.permissions || [];
+        } catch (err) {
+          permissions = [];
+        }
```
jakefromphoton, line 298: took this from `elements`

packages/components/src/store.ts
captDaylight: have you tested this locally?

packages/components/src/systems/Client/index.tsx
```diff
@@ -37,9 +35,7 @@ export default function Client(props: ClientProps) {
     env: props.env ?? getEnv(props)
   });
 
-  const store = props?.createStore
```
captDaylight, line 40: looks like this is only really here for dev in /components, lgtm

captDaylight (approved):

packages/components/src/store.ts
jakefromphoton: yes

PR #1713, opened 2025-04-02
Run playwright e2e on PRs

jakefromphoton:


apps/app/e2e/create_prescription.spec.ts
```diff
@@ -1,22 +1,27 @@
...
+  await page.getByRole('link', { name: /New Patient/ }).click();
 
-  await page.getByRole('button', { name: 'Log in' }).click();
-  await expect(page.getByText('Log in to continue to Photon Clinical App')).toBeVisible();
-  await page.getByRole('button', { name: 'Continue with Google' }).click();
+  const patientNumber = getRandomInt(0, 100_000_000);
 
-  await page.waitForURL(/accounts\.google\.com/, { timeout: 10_000 });
+  await page.getByLabel('First Name').fill('Jimbob');
+  await page.getByLabel('Last Name').fill(`McTesterson_${patientNumber}`);
```
captDaylight, line 12: Ah this is where all those patients are coming from. Is the idea that each time this test is run a new user is created?

captDaylight: also don't patients need an address? I thought that would make prescribing fail...

apps/app/project.json
```diff
@@ -62,6 +62,13 @@
         "command": "env-cmd -f .env.boson npx playwright test"
       }
     },
+    "e2e:ui": {
```
captDaylight, line 65: in some ways, it might be easier to run this test on elements to simulate all the various org settings that could occur... like disabling on enabling combine, orders, etc.

apps/app/e2e/create_prescription.spec.ts
jakefromphoton: I created a separate Org today for the e2e test users so our patients wont get spammed further

And yes an address is needed to make the prescription. The test doesn't go that far yet though, i want to do that in a separate PR since this one is getting big

captDaylight (approved):

PR #1745, opened 2025-04-11
Show mail order tab for Ognomy and add pharmacies

davalopar:
Ognomy want to see gift health for vials and truepill lilly and amazon for auto injectors in their mail order tab.
For context they also send these pharmacy ids into their embed.

Longer term we want to enforce med <> pharmacy relationship but this will reduce the customer issues we're getting from them when the providers send it to the wrong place (i.e. a pharmacy that cannot fulfill it)


packages/settings/src/lib/photon.ts
```diff
@@ -304,7 +305,9 @@ const organizationSettings: {
   // Ognomy
   org_2B1yxV6yS9ROqV1g: {
     logo: 'ognomy_logo.webp',
-    accentColor: '#0a7ade'
+    accentColor: '#0a7ade',
+    mailOrder: true,
+    mailOrderProviders: [GIFTHEALTH_PHARMACY_ID, TRUEPILL_LILLY_PHARMACY_ID, AMAZON_PHARMACY_ID],
```
jknapp25: amazon on patient or provider side? if you want to surface amazon to the patient, use `mailOrderNavigateProviders`

jknapp25: ah saw your comment

jknapp25: do you mind double-checking that this'll hook into our integrated mail order system correctly? this is the first time we've used amazon on the provider side

jknapp25:
first one??? 💃🪩⭐️

jknapp25 (approved):

packages/settings/src/lib/photon.ts
davalopar: checked this in boson and it works as expected

PR #1748, opened 2025-04-15
Org settings edit form

mrrsnn:
- Gathers `settings` from the subresolver of the org graphql query
- Uses the clinical upload server at `https://clinical-api.{env}.health/uploads/logo` to upload logos to S3, saves the Cloudfront cdn url as the `brandLogo` form field
- Uses the clinical resolvers mutation to update org settings on save

https://github.com/user-attachments/assets/87595b74-c858-4b7f-9a1b-37be264d695f


captDaylight:
where's teh video bubble on your video

packages/react/src/provider.tsx
```diff
@@ -643,6 +644,7 @@ const stub = (): never => {
 };
 
 const PhotonClientContext = createContext<PhotonClientContextInterface>({
+  env: 'boson',
```
captDaylight: hardcoded?

apps/app/src/views/components/FileUpload.tsx
```diff
@@ -0,0 +1,141 @@
...
+      toast({
+        title: 'File uploaded',
+        description: 'File uploaded successfully',
+        status: 'success'
+      });
+    } catch (error) {
+      console.error(error);
+      toast({
+        title: 'Error uploading file',
+        description: 'Please try again',
```
captDaylight: add error to description

captDaylight (approved):

packages/react/src/provider.tsx
mrrsnn: the context type was screaming at me and needed a temp value before it's assigned the value from process.env in the Provider component

PR #1753, opened 2025-04-16
Replace client side org settings

mrrsnn:


apps/app/codegen.ts
```diff
@@ -3,7 +3,15 @@ import { CodegenConfig } from '@graphql-codegen/cli';
 const config: CodegenConfig = {
   // schema: 'http://clinical-api.tau.health:8080/graphql',
   schema: 'http://clinical-api.boson.health/graphql',
-  documents: ['apps/app/src/**/Settings/**/*.tsx', 'apps/app/src/views/components/TicketModal.tsx'],
+  documents: [
+    'apps/app/src/**/Settings/**/*.tsx',
```
captDaylight, line 7: just keep in mind, if we add files here there can only be service queries and mutations in these files, just fyi

apps/app/src/views/components/Nav.tsx
```diff
@@ -38,27 +38,44 @@ import {
...
 
+import { graphql } from 'apps/app/src/gql';
+import { useQuery } from '@apollo/client';
+
+const orgSettingsQuery = graphql(/* GraphQL */ `
+  query NavOrgSettingsQuery {
+    organization {
+      settings {
+        providerUx {
+          federatedAuth
```
captDaylight, line 56: nothing we can do , but this is annoying

packages/settings/src/lib/boson.ts
```diff
@@ -43,14 +43,12 @@ const defaultSettings: OrganizationSettings = {
...
-  return defaultSettings;
+export const boson = function (organizationId: string | undefined): MailOrderPharmacyConfigs {
+  const orgConfig = organizationId ? organizationSettings[organizationId] : undefined;
+  return {
+    provider: orgConfig?.mailOrderProviders ?? defaultSettings.mailOrderProviders,
+    patient: orgConfig?.mailOrderNavigateProviders ?? defaultSettings.mailOrderNavigateProviders
+  };
 };
 
 /**
```
captDaylight, line 54: can we not get rid of most of this in the file? everything except the mail order stuff?

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -50,13 +49,30 @@ import { isGLP } from '../utils/isGLP';
 import { Pharmacy as EnrichedPharmacy } from '../utils/models';
 import { datadogRum } from '@datadog/browser-rum';
 import { GetPharmaciesByLocationQuery, Pharmacy as PharmacyType } from '../__generated__/graphql';
+import { getOrgMailOrderPharms } from '@client/settings';
 
 const GET_PHARMACIES_COUNT = 5; // Number of pharmacies to fetch at a time
 
+const pricingEnabledOrgs = new Set([
```
captDaylight, line 56: where is this coming from?

apps/patient/src/configs/theme.ts
```diff
@@ -1,10 +1,7 @@
 import { extendTheme } from '@chakra-ui/react';
 import tinycolor from 'tinycolor2';
-import { getSettings } from '@client/settings';
-
-function generateChakraTheme(key?: string) {
-  const { accentColor } = getSettings(key);
 
+function generateChakraTheme({ accentColor = '#3182ce' }: { accentColor?: string } = {}) {
```
captDaylight, line 4: did this end up fixing the color thing? can you post some images of what it looks like?

captDaylight (approved):

packages/settings/src/lib/boson.ts
mrrsnn: Yep gonna remove that after I use the contents to create our SQL Scripts!

apps/patient/src/views/Pharmacy.tsx
mrrsnn: For Josh's GLP-1 testing. We only have it enabled for Sesame currently. I asked @jknapp25 and he said to do this for now because we're not sure if we're going to expose an `enablePricing` setting for orgs yet

apps/patient/src/configs/theme.ts
mrrsnn: Nope, I forgot to ask Jomi why we don't currently use those colors

PR #1764, opened 2025-04-21
hardcoding novo pharmacy pilot

billoran8:


apps/patient/src/components/BrandedOptions.tsx
```diff
@@ -7,16 +7,20 @@ interface Props {
   options: string[];
   location: string;
   selectedId: string;
-  amazonPharmacyOverride: string | undefined;
+  brandedOptionOverrides: BrandedOptionOverrides;
```
billoran8, line 10: instead of passing down a single string, we're passing down an object, which should make it easier to add more to this set.

longer term this entire setup is difficult to work with and i think we should consider redeveloping pharmacy search.

apps/patient/src/components/PharmacyInfo.tsx
```diff
@@ -269,19 +273,38 @@ export const PharmacyInfo = ({
   const directionsUrl = `http://maps.google.com/?q=${pharmacy?.name}, ${pharmacyFormattedAddress}`;
 
   let amazonPharmacyElementOverride = undefined;
-  if (amazonPharmacyOverride) {
+  const isAmazonPharmacy = pharmacy.id === process.env.REACT_APP_AMAZON_PHARMACY_ID ? true : false;
```
billoran8: this entire widget should be redone to more naturally express the attributes - there's a lot of overrides and booleans here that maybe should be on a pharmacy/displaySettings type object to have less conditionals that have to be managed up above.

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -193,10 +193,97 @@ export const Pharmacy = () => {
         .filter((offer) => offer.deliveryEstimate !== undefined);
 
       if (amazonOffers.length > 0 && amazonOffers[0]?.deliveryEstimate?.deliveryPromise) {
-        setAmazonPharmacyOverride(amazonOffers[0]?.deliveryEstimate?.deliveryPromise);
+        const newBrandedOptionsOverride = {
+          ...brandedOptionsOverride,
+          amazonPharmacyOverride: amazonOffers[0]?.deliveryEstimate?.deliveryPromise
+        };
+
+        if (JSON.stringify(newBrandedOptionsOverride) !== JSON.stringify(brandedOptionsOverride)) {
```
billoran8: only set the state if it's different; otherwise these two setStates will trigger in a loop.

prob a better way to do this - how?

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -193,10 +193,97 @@ export const Pharmacy = () => {
...
+          setBrandedOptionsOverride(newBrandedOptionsOverride);
+        }
       }
     }
 
+    const determineNovocareExperimentSegment = (order: Order): string | undefined => {
```
billoran8: this is all copied from the original amazonPharmacy hardcoded experiment with

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -193,10 +193,97 @@ export const Pharmacy = () => {
...
+
+      const deliveryType = getDeliveryType();
+
+      if (deliveryType) {
+        const newBrandedOptionsOverride = {
+          ...brandedOptionsOverride,
+          novocareExperimentOverride: deliveryType
+        };
+
+        if (JSON.stringify(newBrandedOptionsOverride) !== JSON.stringify(brandedOptionsOverride)) {
```
billoran8: same as above, only changing the state if it's different to prevent pingponging. better way to do this?

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -637,6 +725,22 @@ export const Pharmacy = () => {
               }
             }
 
+            if (brandedOptionsOverride?.novocareExperimentOverride) {
```
billoran8, line 635: adding some metrics. wanted to add more (user selected pharmacy closest to their home address, user selected one with a "available in 30 mins" tag, etc etc) but it's really annoying to have to do that at this level

apps/patient/src/assets/novo_circle.png
mrrsnn, line 1: have Paul upload this to s3 and make sure that the brandLogo is set to the cdn url in org settings

apps/patient/src/components/PharmacyInfo.tsx
```diff
@@ -235,7 +235,11 @@ interface PharmacyInfoProps {
   showHours?: boolean;
   isCurrentPharmacy?: boolean;
   orderFulfillment?: OrderFulfillment;
-  amazonPharmacyOverride?: string;
+  brandedOptionOverride?: BrandedOptionOverrides;
+}
+
+interface BrandedOptionOverrides {
+  [key: string]: string | undefined;
 }
```
mrrsnn: Seems like we're defining this same interface across files. Might be nice to put it one spot

apps/patient/src/components/PharmacyInfo.tsx
jfab3: Can this shorten to `const isAmazonPharmacy = pharmacy.id === process.env.REACT_APP_AMAZON_PHARMACY_ID` ?

apps/patient/src/components/PharmacyInfo.tsx
```diff
@@ -269,19 +273,38 @@ export const PharmacyInfo = ({
...
-          {amazonPharmacyOverride}
+          {brandedOptionOverride?.amazonPharmacyOverride}
         </Text>
       </HStack>
     );
   }
 
+  let novocarePharmacyElementOverride = undefined;
+  const isNovocarePharmacy =
+    pharmacy.id === process.env.REACT_APP_NOVOCARE_PHARMACY_ID ? true : false;
```
jfab3: Same as above

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -193,10 +193,97 @@ export const Pharmacy = () => {
...
+      const deliveryType = getDeliveryType();
+
+      if (deliveryType) {
+        const newBrandedOptionsOverride = {
+          ...brandedOptionsOverride,
+          novocareExperimentOverride: deliveryType
+        };
+
+        if (JSON.stringify(newBrandedOptionsOverride) !== JSON.stringify(brandedOptionsOverride)) {
+          setBrandedOptionsOverride(newBrandedOptionsOverride);
```
mrrsnn: This file is insanely long already, can we at the very least abstract most of this logic away to a util function in another file or something?

apps/patient/src/views/Pharmacy.tsx
jfab3: Could you explicitly specify the `dependencies` array for the `useEffect` to limit how often it runs?

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -193,10 +193,97 @@ export const Pharmacy = () => {
...
+
+          // If we have exactly the right number of medications
+          if (medications.length === patterns.length) {
+            // Check if we can match each pattern to a unique medication
+            const unmatchedMedications = [...medications];
+            const allPatternsMatch = patterns.every((pattern) => {
+              const matchIndex = unmatchedMedications.findIndex((med) => med.match(pattern));
+              if (matchIndex !== -1) {
+                // Remove the matched medication so it can't be matched again
+                unmatchedMedications.splice(matchIndex, 1);
```
jfab3: Where else is `unmatchedMedications` used?

apps/patient/src/assets/novo_circle.png
billoran8: super into that; i hate putting images in git repositories

apps/patient/src/components/PharmacyInfo.tsx
billoran8: yep, thought so to but then was like "ehhh... none of them feel like a great spot for it to be authortative" but that's not a probelm i'll just put it at the highest level

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -661,6 +765,9 @@ export const Pharmacy = () => {
             } else if (selectedId === process.env.REACT_APP_COSTCO_PHARMACY_ID) {
               type = 'MAIL_ORDER';
               selectedPharmacy = { id: selectedId, name: 'Costco Pharmacy' };
+            } else if (selectedId === process.env.REACT_APP_NOVOCARE_PHARMACY_ID) {
+              type = 'MAIL_ORDER';
```
jfab3, line 676: Feel like these should be enums instead of hard-coding the string everywhere

apps/patient/src/views/Pharmacy.tsx
billoran8: good idea, i'll make another file tos tore this logic. long term though, these shoudl be rewritten because everything's everywhere imo

apps/patient/src/views/Pharmacy.tsx
billoran8: yeah great point - i'll change that here and maybe we can spread it out more later. IMO these pages are a disaster so i don't want to spend too much time cleaning them up

billoran8: oh actually these are sort of enums, let me show u something

apps/patient/src/views/Pharmacy.tsx
billoran8: util function created

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -185,31 +187,35 @@ export const Pharmacy = () => {
...
-    async function fetchOffers() {
-      const offers = await getOffers(order.id);
-
-      const amazonOffers = offers
-        .filter((offer) => offer.supplier === 'AMAZON_PHARMACY')
-        .filter((offer) => offer.deliveryEstimate !== undefined);
-
-      if (amazonOffers.length > 0 && amazonOffers[0]?.deliveryEstimate?.deliveryPromise) {
-        setAmazonPharmacyOverride(amazonOffers[0]?.deliveryEstimate?.deliveryPromise);
+    const determineOverrides = async () => {
```
billoran8, line 177: larger functions all put into `pharmacy.util.tsx`

apps/patient/src/components/PharmacyInfo.tsx
billoran8: yep

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -185,31 +187,35 @@ export const Pharmacy = () => {
...
 
-    const validOrganizationIds = [
-      'org_KzSVZBQixLRkqj5d', // boson Test Organization 11
-      'org_kVS7AP4iuItESdMA', // neutron Photon Test Org
-      'org_wM4wI7rop0W1eNfM', // production found
-      'org_pcPnPx5PVamzjS2p' // production measured
-    ];
+      // these functions will be called and make a state change to brandedOptionsOverride using setBrandedOptionsOverride
+      // if there are branded option overrides
+      const amazonExperimentOverride = await fetchOffers(order);
```
billoran8, line 193: having ecerythign orcehstrated here beacuse we may only want one active at a time

apps/patient/src/components/BrandedPharmacyCard.tsx
```diff
@@ -6,15 +6,17 @@ import altoLogo from '../assets/alto_logo.svg';
 import costcoLogo from '../assets/costco_logo_small.png';
 import costPlusLogo from '../assets/costplus_logo_small_circle.png';
 import walmartLogo from '../assets/walmart_logo_small_circle.png';
+import novocareLogo from '../assets/novo_circle.png';
```
billoran8, line 9: i'll upload stuff to s3 later

mrrsnn (approved):

jfab3 (approved):

PR #1772, opened 2025-04-23
Remove settings

mrrsnn:


billoran8 (commented):
can we merge this in before things are copied over to the db-backed variants?

billoran8:
and/or how is that happening? are we telling customers to do it htemselves? are we helping them?

jknapp25 (approved):
Something like `providerMailOrderPharmacyOptions` would be a lil more clear. But probably won't be around for long anyway huh?

billoran8:
oh morrison already did it and the stuff is already backed up in the db

billoran8 (approved):

PR #1811, opened 2025-05-05
show prescription routing alerts for Zepbound

jfab3:


jfab3:
image

packages/components/src/index.ts
```diff
@@ -80,6 +81,7 @@ export {
   createQuery,
   formatDate,
   generateString,
+  getRoutingConstraint,
```
captDaylight: I don't think we need to be exporting this for the Elements to call, should probably be contained within the `components`, maybe in the `PrescribeProvider`

packages/elements/src/photon-multirx-form/photon-prescribe-workflow.tsx
```diff
@@ -263,6 +270,13 @@ export function PrescribeWorkflow(props: PrescribeProps) {
     setScreeningAlerts(data?.prescriptionScreen?.alerts ?? []);
   };
 
+  const routingConstraints = createMemo((): RoutingConstraint[] => {
+    const dedupedPrescriptions: Prescription[] = removeDuplicateTreatments(draftPrescriptions());
+    return dedupedPrescriptions.map((prescription: Prescription) =>
+      getRoutingConstraint(prescription)
+    );
+  });
+
```
jakefromphoton: Is RoutingConstraint logic mainly up here in `<photon-prescribe-workflow>` because of this dedupe logic from screenDraftPrescriptions? If so, I agree with @captDaylight - moving it down into `components/DraftPRescriptions` or `PrescribePRovider` would be better. 

I wouldn't mind duplicating the dedupe logic for now - or perhaps add a deduped memo property to `DraftPrescriptionsProvider` which would be available anywhere we have `useDraftPrescriptions`.

Screening logic needs to get refactored out of this file and into another spot -- we just haven't gotten to it yet.

jfab3: Removed the deduplication for routing constraints.

packages/components/src/index.ts
jfab3: Moved `getRoutingConstraint` call to `PrescribeProvider`

jakefromphoton (approved):

PR #1821, opened 2025-05-07
Update pharmacy selection for RTBC

jakefromphoton:
* Putting the selected coverage option's pharmacy into "Send to Patient" for now. Not ideal -- future work could move this to Local Pickup/Mail Order tabs. See [this notion ticket](https://www.notion.so/photons/RTBC-Pharmacy-Selection-1ec8bfbbf4ec8041bd06de2609c6819a?pvs=4).
* also began concept of a `orderFormData` store in the PrescribeProvider. This will be a type-safe replacement to the form store/actions. For now, it just holds the selected pharmacy ID.
* enabled sourcemaps for `components` package

Screenshot 2025-05-07 at 4 12 41 PM

https://github.com/user-attachments/assets/aefdd354-b0c0-4045-9571-274da0b9348f


packages/components/src/systems/PrescribeProvider/PrescribeProvider.tsx
```diff
@@ -111,6 +125,11 @@ export const PrescribeProvider = (props: PrescribeProviderProps) => {
   );
   const [didSelectOtherCoverageOption, setDidSelectOtherCoverageOption] =
     createSignal<boolean>(false);
+  const [selectedCoverageOption, setSelectedCoverageOption] = createSignal<
```
captDaylight, line 130: this logic might need to change if we ever allow there to be multiple checks, but for now it's fine

jakefromphoton: definitely could see it becoming a list of selections eventually. Then pharmacy selection gets even more fun :)

packages/components/src/systems/PrescribeProvider/PrescribeProvider.tsx
```diff
@@ -26,12 +26,22 @@ import {
 } from '../../fetch';
 import { triggerToast, useRecentOrders } from '../../index';
 import { useDraftPrescriptions } from '../DraftPrescriptions';
+import { createStore } from 'solid-js/store';
+
+// The order form data consists will consist of the list of the selected prescriptions and pharmacy.
```
captDaylight: consists will consist

captDaylight: also form will ultimately rely on the prescription ids, but maybe that nuance isn't necessary

packages/components/vite.config.ts
```diff
@@ -18,6 +18,7 @@ export default defineConfig({
       name: 'components',
       fileName: 'index'
     },
+    sourcemap: true,
```
captDaylight: hot

captDaylight (approved):

packages/components/src/systems/PrescribeProvider/PrescribeProvider.tsx
jakefromphoton: ya i agree let me update the comment

packages/components/vite.config.ts
jakefromphoton: reverting it for now because of the CI errors :/

PR #1828, opened 2025-05-08
Fix edit reselecting old treatment data

jakefromphoton:
- fix editing a draft prescription, and then clicking "Send Order" (or "Edit" draft again) showing a popup about "Lose Unsaved Prescription" improperly
- fix catalog.data null error
- bonus: this also seems to fix the ResizeObserverLoop error we keep seeing when interacting with the Med Search component

### DIFF
--- a/packages/elements/src/photon-dropdown/photon-dropdown.tsx
+++ b/packages/elements/src/photon-dropdown/photon-dropdown.tsx
@@ -166,7 +166,6 @@ export const PhotonDropdown = <T extends { id: string }>(props: {
     if (props.selectedData) {
       // @ts-ignore
       setSelected(props.selectedData);
-      dispatchSelect(props.selectedData);
     }
     if (props.selectedData === undefined) {
       setSelected(undefined);
--- a/packages/elements/src/photon-medication-dropdown/photon-medication-dropdown.tsx
+++ b/packages/elements/src/photon-medication-dropdown/photon-medication-dropdown.tsx
@@ -30,7 +30,7 @@ import tailwind from '../tailwind.css?inline';
 import styles from './style.css?inline';
 
 // Types
-import { Treatment, PrescriptionTemplate, TreatmentOption } from '@photonhealth/sdk/dist/types';
+import { PrescriptionTemplate, Treatment, TreatmentOption } from '@photonhealth/sdk/dist/types';
 
 //Virtual List
 import { createVirtualizer, VirtualItem } from '@tanstack/solid-virtual';
@@ -167,7 +167,6 @@ export const PhotonMedicationDropdown = <T extends { id: string }>(props: {
     if (props.selectedData) {
       // @ts-ignore
       setSelected(props.selectedData);
-      dispatchSelect(props.selectedData);
     }
     if (props.selectedData === undefined) {
       setSelected(undefined);
--- a/packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
+++ b/packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
@@ -292,9 +292,7 @@ const Component = (props: ComponentProps) => {
     <div
       ref={ref}
       on:photon-data-selected={(e: any) => {
-        if (store.catalog.data) {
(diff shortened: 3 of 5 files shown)


jknapp25 (approved):

packages/elements/src/photon-medication-dropdown/photon-medication-dropdown.tsx
```diff
@@ -167,7 +167,6 @@ export const PhotonMedicationDropdown = <T extends { id: string }>(props: {
     if (props.selectedData) {
       // @ts-ignore
       setSelected(props.selectedData);
-      dispatchSelect(props.selectedData);
     }
```
jakefromphoton, line 170: Removing `dispatchSelect` calls when the parent updates `props.selectedData` because it leads to reactivity cycles. We shouldn't be emitting "selection" events unless the user is selecting something new within the component itself.

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -174,6 +174,12 @@ export const AddPrescriptionCard = (props: {
     setSearchText('');
   };
 
+  createEffect(() => {
+    if (props.store.treatment?.value) {
+      setSearchText(props.store.treatment.value.name);
+    }
+  });
+
```
jakefromphoton, line 182: One side-effect of removing the `dispatchSelect` is that we need the parent to now directly perform some logic - such as updating the search text when the treatment changes.

packages/elements/src/photon-dropdown/photon-dropdown.tsx
```diff
@@ -166,7 +166,6 @@ export const PhotonDropdown = <T extends { id: string }>(props: {
     if (props.selectedData) {
       // @ts-ignore
       setSelected(props.selectedData);
-      dispatchSelect(props.selectedData);
     }
```
jakefromphoton, line 169: This `dispatchSelect` affected mobile's Med Search.

I manually tested the other 5 uses of Dropdown to make sure they work as expected:
* prescription dispense-units
* patient gender
* patient sex
* patient address state
* order patient select

packages/elements/src/photon-multirx-form/components/DraftPrescriptionCard.tsx
```diff
@@ -60,12 +60,9 @@ export const DraftPrescriptionCard = (props: {
         deletePrescription(formData.id);
       }
 
-      window.scrollTo({
+      props.prescriptionRef?.scrollIntoView({
         behavior: 'smooth',
```
jakefromphoton, line 64: `scrollIntoView` seems simpler (and actually works on my local machine)

packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
```diff
@@ -292,7 +292,7 @@ const Component = (props: ComponentProps) => {
     <div
       ref={ref}
       on:photon-data-selected={(e: any) => {
-        dispatchTreatmentSelected(ref, e.detail.data, store.catalog.data!.id || '');
+        dispatchTreatmentSelected(ref, e.detail.data, store.catalog.data?.id || '');
```
jakefromphoton, line 295: we were getting null reference errors in the console here

packages/elements/src/photon-dropdown/photon-dropdown.tsx
captDaylight: love these notes :)

packages/elements/src/photon-medication-dropdown/photon-medication-dropdown.tsx
captDaylight: could this be the bug when we open the dropdown?

captDaylight (approved):

packages/elements/src/photon-medication-dropdown/photon-medication-dropdown.tsx
jakefromphoton: Looks like it does fix that! woot

PR #1835, opened 2025-05-09
improve cancel order UI response time

jfab3:
- No longer manually updating the cache since we get the correct status from the `cancelOrder` mutation: https://github.com/Photon-Health/lambdas/pull/2021
- When `onQueryUpdated` is called for the `cancelOrder` mutation, do not execute the `GET_ORDER` query again. The query would run before `order.state` has been updated to `CANCELED`, so the order would appear as if it had not been canceled at all.

### DIFF
--- a/apps/app/src/views/routes/Order.tsx
+++ b/apps/app/src/views/routes/Order.tsx
@@ -243,7 +243,37 @@ export const Order = () => {
     cancelReasonRef.current = cancelReason;
   }, [cancelReason]);
 
-  const [cancelOrder] = useMutation(CANCEL_ORDER);
+  const [cancelOrder] = useMutation(CANCEL_ORDER, {
+    /*
+    optimisticResponse: {
+      cancelOrder: {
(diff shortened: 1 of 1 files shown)


apps/app/src/views/routes/Order.tsx
```diff
@@ -244,28 +245,12 @@ export const Order = () => {
...
-            order: {
-              ...existingOrder?.order,
-              state: types.OrderState.Canceled
-            }
-          },
-          variables: { id: id! }
-        });
+    onQueryUpdated: (observableQuery) => {
+      if (observableQuery.queryName !== GET_ORDER_QUERY_NAME) {
+        return client?.defaultOptions?.mutate?.onQueryUpdated;
```
jakefromphoton, line 250: What does this do? I'm new to ApolloClient and looking at docs, but it doesnt seem clear

jakefromphoton (approved):

PR #1840, opened 2025-05-13
enable auto-routing when constrained to a single pharmacy

jfab3:


packages/components/src/systems/RoutingConstraints/RoutingConstraint.test.tsx
```diff
@@ -0,0 +1,360 @@
+import {
+  identityConstraint,
+  combineAllRoutingConstraints,
+  combineRoutingConstraints,
+  RoutingConstraint
+} from './RoutingConstraint';
+
```
whittle-photon, line 7: Love the depth of testing here. This is great.

packages/components/src/systems/RoutingConstraints/RoutingConstraint.test.tsx
```diff
@@ -0,0 +1,360 @@
...
+        {
+          id: 'fake_id3',
+          name: 'Fake Pharmacy 3'
+        }
+      ]
+    };
+
+    const combinedRc = combineRoutingConstraints(rc1, rc2);
+    const expectedRc = {
+      prescription: { id: '', prescribable_name: '' },
```
whittle-photon: If you’re going to combine routing constraints into a single resultant constraint object, I think you really need to include all the prescriptions somehow. 

Alternately, roll up constraints into a different structure.

packages/components/src/systems/RoutingConstraints/RoutingConstraint.test.tsx
```diff
@@ -0,0 +1,360 @@
...
+          id: 'fake_id3',
+          name: 'Fake Pharmacy 3'
+        }
+      ]
+    };
+
+    expect(combinedRc).toStrictEqual(expectedRc);
+  });
+
+  it('combining an include list with the identiy constraint returns the input include list (regardless of input order)', () => {
```
whittle-photon: I think it’s a little confusing to refer to an identity constraint. I would label the special constraints as `no_route` and `no_advice`.

packages/components/src/systems/RoutingConstraints/RoutingConstraint.tsx
```diff
@@ -131,3 +131,137 @@ export function getPrescriptionRoutingConstraints(
...
+  if (rc.routing_constraint_type === 'no_advice') {
+    return {
+      prescription: rc.prescription,
+      constraint_pharmacies: [],
+      routing_constraint_type: 'exclude'
+    };
+  } else {
+    return rc;
+  }
+}
```
whittle-photon: These could be combined into a canonicalization operation.

packages/components/src/systems/RoutingConstraints/RoutingConstraint.tsx
```diff
@@ -131,3 +131,137 @@ export function getPrescriptionRoutingConstraints(
...
+
+  const getPharmaciesDifference = (idSet1: Set<string>, idSet2: Set<string>) => {
+    const pharmacies = [];
+    for (const id of Array.from(idSet1)) {
+      if (!idSet2.has(id)) {
+        pharmacies.push(pharmacyMap.get(id));
+      }
+    }
+    return pharmacies;
+  };
```
whittle-photon, line 232: Does the standard library really not have basic set operations?

whittle-photon (changes requested):
Love the testing! Just some thoughts.

packages/components/src/systems/RoutingConstraints/RoutingConstraint.tsx
jfab3: It does but only as of Node.js 22 (2024-04-24) 🙃

packages/components/src/systems/RoutingConstraints/RoutingConstraint.test.tsx
jfab3: Thanks!

packages/components/src/systems/RoutingConstraints/RoutingConstraint.test.tsx
jfab3: Makes sense–I am now combining the prescriptions for all constraints into an array.

Conceptually this probably makes more sense. Within a constraint, we have an array of prescriptions which can be routed to the valid pharmacies.

packages/components/src/systems/RoutingConstraints/RoutingConstraint.test.tsx
```diff
@@ -0,0 +1,386 @@
...
+          name: 'Fake Pharmacy 2'
+        }
+      ]
+    };
+
+    expect(combinedRc).toStrictEqual(expectedRc);
+  });
+
+  it('takes the set difference when the first routing constraint is type include and the second is type exclude', () => {
+    const rc1: RoutingConstraint = {
```
billoran8, line 63: these tests are incredible

billoran8: thorough, but simple, clearly labeled, simple to follow, one expect per, i love these

billoran8: great job!!!

billoran8: when these break it'll be so clear what the issue is

packages/components/src/systems/RoutingConstraints/RoutingConstraint.tsx
whittle-photon: I took until v.22 to add set operations. Amazing. Is there a polyfill we could use?

packages/components/src/systems/RoutingConstraints/PrescriptionRoutingAlert.tsx
```diff
@@ -57,6 +57,13 @@ export const PrescriptionRoutingAlert = (props: {
     if (!validTypes.includes(props.routingConstraint.routing_constraint_type)) {
       throw new Error('Invalid RoutingConstraintType for PrescriptionRoutingAlert');
     }
+
+    if (
+      props.routingConstraint.prescriptions.length !== 1 ||
+      props.routingConstraint.prescriptions[0].id !== props.prescription.id
```
whittle-photon: Could this test be abstracted out so that the same function is used here and when deciding if the alert is visible?

packages/components/src/systems/RoutingConstraints/RoutingConstraint.test.tsx
jfab3: thank you 🙏

packages/components/src/systems/RoutingConstraints/PrescriptionRoutingAlert.tsx
jfab3: done!

whittle-photon (approved):

PR #1845, opened 2025-05-13
Rate limit generateCoverageOptions to 5 calls

jakefromphoton:
* Up to 5 calls per patient session (instead of 1)
* Extra logic so that deleting a draft prescription does not trigger an unnecessary call
* reset call count on patient change or page reload
* this fixes a selected therapeutic alternative prescription not being linked to a coverage, and thus no price visible in Patient App (see this [mini RFC](https://www.notion.so/photons/Mini-RFC-RTBC-Link-Therapeutic-Alternative-Coverage-to-Prescription-1f28bfbbf4ec80b99a43f28369d47e0a?pvs=4)) (though the coverage would still be missing if the rate limit is reached)

### DIFF
--- a/packages/components/src/systems/PrescribeProvider/PrescribeProvider.tsx
+++ b/packages/components/src/systems/PrescribeProvider/PrescribeProvider.tsx
@@ -5,6 +5,7 @@ import {
   createMemo,
   createSignal,
   JSXElement,
+  untrack,
   useContext
 } from 'solid-js';
 import { format } from 'date-fns';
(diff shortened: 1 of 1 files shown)


packages/components/src/systems/PrescribeProvider/PrescribeProvider.tsx
```diff
@@ -186,16 +189,25 @@ export const PrescribeProvider = (props: PrescribeProviderProps) => {
   createEffect(() => {
     const pharmacyId = patientPreferredPharmacyId();
     const prescriptions = draftPrescriptions();
+
+    const GENERATE_COVERAGE_OPTIONS_RATE_LIMIT = 5;
+    const callCount = untrack(() => generateCoverageOptionCallCount());
+    const isBelowRateLimit = callCount < GENERATE_COVERAGE_OPTIONS_RATE_LIMIT;
+    const previousPrescriptionCount = untrack(() => previousPrescriptionIds().length);
+    const hasAddedPrescriptions = prescriptions.length > previousPrescriptionCount;
```
jfab3, line 197: Is checking prescription count robust enough? I guess we don't have any way to replace a prescription in a single operation, but I wonder if it's worth explicitly checking the prescription id's.

PR #1846, opened 2025-05-14
add disabled property to RadioGroupCards

jfab3:
Will be used to disable mail order pharmacies which cannot be routed to.

image


mrrsnn (approved):
LGTM

PR #1852, opened 2025-05-19
Disable un-routable mail order pharmacies 

jfab3:
image

### DIFF
--- a/packages/components/src/systems/PharmacySelect/PharmacySelect.tsx
+++ b/packages/components/src/systems/PharmacySelect/PharmacySelect.tsx
@@ -1,13 +1,11 @@
 import { types } from '@photonhealth/sdk';
 import { createEffect, createMemo, createSignal, For, onMount, Show, untrack } from 'solid-js';
-import RadioGroupCards, { RadioGroupCardsProps } from '../../particles/RadioGroupCards';
-import { createRadioGroupCardsController } from '../../particles/RadioGroupCards';
+import RadioGroupCards, { RadioGroupCardsContextValue } from '../../particles/RadioGroupCards';
 import Tabs from '../../particles/Tabs';
 import PharmacySearch from '../PharmacySearch';
 import { MailOrderPharmacy } from './MailOrderPharmacy';
 import { SendToPatient } from './SendToPatient';
 import { usePrescribe } from '../PrescribeProvider';
-import { Dynamic } from 'solid-js/web';
 
 enum SendToPatientEnum {
   sendToPatient = 'SEND_TO_PATIENT'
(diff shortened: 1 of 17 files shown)


packages/components/src/systems/PharmacySelect/PharmacySelect.tsx
```diff
@@ -156,17 +172,18 @@ export function PharmacySelect(props: PharmacySelectProps) {
           <div class={activeTab() !== TabNamesEnum.mailOrder ? 'hidden' : ''}>
             <RadioGroupCards
               label="Pharmacies"
-              initSelected={props?.mailOrderPharmacyIds?.[0]}
+              initSelected={initMailOrderPharmacyId()}
```
mrrsnn, line 175: What's the point of doing all this passing around and manipulation of RadioGroupCardsContext? Why can't we just change the prop and have an effect inside the component that updates the internal state when the initSelected changes? That way there's one linear flow of state change

jfab3: `initSelected` is only used to set the selected value when the `RadioGroupCardsRoot` mounts. 

Otherwise, anytime `initSelected` changes (which could happen when `unroutablePharmacyIds` changes), we'd set the selected card. This could override a manual selection a user made.

So I figured I could explicitly manipulate the RadioGroupContext to reset to `initSelected` **only** when the manually selected pharmacy is disabled.

packages/components/src/systems/PharmacySelect/PharmacySelect.tsx
```diff
@@ -156,17 +172,18 @@ export function PharmacySelect(props: PharmacySelectProps) {
...
                 if (activeTab() === TabNamesEnum.mailOrder) {
                   props.setPharmacyId(pharmacyId);
                 }
               }}
+              contextRef={(context) => (radioGroupContext = context)}
             >
               <For each={props?.mailOrderPharmacyIds || []}>
                 {(id) => (
-                  <RadioGroupCards.Option value={id}>
+                  <RadioGroupCards.Option value={id} disabled={unroutablePharmacyIds().has(id)}>
```
billoran8, line 186: nit: any thoughts on maybe describing why this is pharmacy is unroutable in the UI? i could imagine this not being clear to a presceriber

jfab3: Yes! That will be the sequel PR.

Going to have a message saying something like "One or more prescriptions in the order cannot be filled at this pharmacy" in the pharmacy card.

ritabulman (approved):

PR #1853, opened 2025-05-19
Use a local pharmacy for RTBC if the patient doesn't have a preferred pharmacy set

captDaylight:


packages/components/src/systems/PharmacySearch/index.tsx
```diff
@@ -111,6 +74,7 @@ interface PharmacyExtended extends Pharmacy {
 
 export default function PharmacySearch(props: PharmacySearchProps) {
   const client = usePhotonClient();
+  const { googleMapsServices } = usePrescribe();
```
jakefromphoton: this could break the PatientForm, which uses the `<PharmacySearch>` without a prescribe context

packages/components/src/systems/PrescribeProvider/PrescribeProvider.tsx
```diff
@@ -191,22 +229,74 @@ export const PrescribeProvider = (props: PrescribeProviderProps) => {
...
-      });
+      if (preferredId) {
+        // check coverage on the preferred pharmacy
+        generateCoverageOptions(prescriptions, preferredId).then((generatedCoverageOptions) => {
+          setCoverageOptions(generatedCoverageOptions);
+        });
+      } else {
+        // check coverage on a walgreens or cvs pharmacy near the patient
+        if (patientAddress()) {
+          fetchLocalPharmacies(patientAddress()!).then((pharmacies) => {
```
jakefromphoton: if we store `patientAddress()` in a variable at the top of the createEffect with the other signals, and use the variable after that, we won't have to do `patientAddress()!`

captDaylight: YUP good thinking

packages/components/src/systems/PharmacySearch/index.tsx
captDaylight: very big brained thought

jakefromphoton (approved):

PR #1871, opened 2025-05-21
Option for disabling toasts for prescribe flow element

jakefromphoton:
Making "Prescription Added" toasts optonal to disable. These were too chatty when a customer would add several prescriptions in sequence via `template-ids` prop. 

Toasts enable to ON by default. Error toasts will display even when toasts are disabled.

Example:
```
<photon-prescribe-workflow
    enable-toasts="false"
></photon-prescribe-workflow>
```

Part of this ticket: https://www.notion.so/photons/Elements-are-glitching-when-bumping-from-0-14-14-to-0-16-5-1f98bfbbf4ec80c7abdee941a8becf74?pvs=4


jakefromphoton:
Did this smaller fix instead: https://github.com/Photon-Health/client/pull/1878

PR #1881, opened 2025-05-22
Potential Auth0 Invitation Fix

captDaylight:
I believe the react provider was calling check session if there was in invitation id in the url and causing something to lose the invitation context along the way before the flow could complete.


ritabulman (approved):

jknapp25 (approved):

packages/react/src/provider.tsx
```diff
@@ -714,6 +714,18 @@ export const PhotonProvider = (opts: {
 
   useEffect(() => {
     const initialize = async () => {
+      // Check if we're in an invitation flow so we don't call check session and lose the invitation context
+      const searchParams = new URLSearchParams(window.location.search);
+      const invitation = searchParams.get('invitation');
+      const organization = searchParams.get('organization');
+      if (invitation && organization) {
+        console.log(
+          '*** invitation and organization found, Daoeusnth aosnetuh santheo uISPATCHING NULL USER'
```
jknapp25: good name for a child `Daoeusnth aosnetuh santheo`

packages/react/src/provider.tsx
```diff
@@ -714,6 +714,15 @@ export const PhotonProvider = (opts: {
...
   useEffect(() => {
     const initialize = async () => {
+      // Check if we're in an invitation flow so we don't call check session and lose the invitation context
+      const searchParams = new URLSearchParams(window.location.search);
+      const invitation = searchParams.get('invitation');
+      const organization = searchParams.get('organization');
+      if (invitation && organization) {
+        dispatch({ type: 'INITIALISED', user: null });
+        return;
+      }
```
jakefromphoton, line 724: curious how you learned about this fix!

PR #1901, opened 2025-05-28
enabling amazon pharmacy to be re-routable

billoran8:
note: i won't feel comfortable deploying this onto prod until i can verify some stuff on neutron


apps/patient/src/utils/general.ts
```diff
@@ -120,6 +120,14 @@ export function isDelivery({
   );
 }
 
+/**
+ * Currently we only allow re-routing to non-delivery pharmacies. This is changing and we'll need
+ * a more robust solution. This function is the first pipe.
+ */
+export function isRerouteablePharmacy({ pharmacy }: { pharmacy?: Pharmacy }): boolean {
```
jknapp25, line 127: need params in object? or okay to just do `pharmacy` since there's only one?

apps/patient/src/utils/general.ts
```diff
@@ -120,6 +120,14 @@ export function isDelivery({
...
+  return pharmacy?.id === 'phr_01GA9HPV5XYTC1NNX213VRRBZ3'; // amazon pharmacy
+}
```
mrrsnn, line 129: Naive question: Is there a reason we don't want to just add a column of `reroutable` on the pharmacy model? If it's just because we're unsure of what that criteria will be in the future, that's understandable.

apps/patient/src/views/Status.tsx
```diff
@@ -267,7 +275,11 @@ export const Status = () => {
                         !isDeliveryPharmacy &&
                         exception &&
                         callPharmacyButton(true)}
-                      {!isDeliveryPharmacy && displayPharmacy && canReroute && rerouteButton}
+                      {/* right now mail order (aka delivery pharmacies) are considered to not be re-routable but this will eventually change*/}
+                      {(!isDeliveryPharmacy || isRerouteable) &&
+                        displayPharmacy &&
+                        canReroute &&
```
jknapp25: `canReroute` sounds similar, is that related?

apps/patient/src/utils/general.ts
billoran8: not sure just yet! keeping my options open - maybe in the future this will be a property on the pharmacy? it could go lots of diff ways - just did what was easiest

mrrsnn: I like passing the object so that when you inevitably extend the number of required arguments, you don't have to adhere to annoying order of parameters that make no sense as it grows.

apps/patient/src/views/Status.tsx
billoran8: yep! it's whether ornot the order can reroute -i'll rename the variable to make it clearer

apps/patient/src/utils/general.ts
jknapp25: yeah same here @mrrsnn, wasn't sure if this list would grow tho

apps/patient/src/views/Status.tsx
```diff
@@ -260,14 +268,18 @@ export const Status = () => {
                       {displayPharmacy && !isDeliveryPharmacy && !exception && navigateButton}
                       {displayPharmacy &&
                         !isDeliveryPharmacy &&
-                        canReroute &&
+                        canOrderReroute &&
```
billoran8, line 271: there's a bunch of rules for whether or not an order can reroute independently of whether the current pharmacy allows it

apps/patient/src/utils/general.ts
billoran8: yep, not sure what the criteria will be and what other dials and knobs there'll be.

jknapp25 (approved):

apps/patient/src/utils/general.ts
billoran8: whether or not stuff is reroutable is only half the equation - what we'll also need to do is allow the flexibility for each pharmacy to handle reroutes in its own way as well - ie amazon right now will send faxes but eventually there'll be another endpoint we can ping to express it to amazon

PR #1912, opened 2025-05-29
Bug fix: Prescribers can click Save Prescriptions or Send Order while templated RXs are loading, results in error

mrrsnn:
Per Remedy Meds:https://photonhealth.slack.com/archives/C02V18YAEE9/p1748532411318539

Need to make the button component non-clickable while it's in a loading state. 

https://github.com/user-attachments/assets/ac0db9cf-d48b-4915-8600-86917e52b088

This, combined with the usage of `isLoadingPrefills` from the prescribe context should prevent fast-clicking providers from saving rxs or sending orders before the templated prescriptions are loaded into the form.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46676,7 +46676,7 @@
     },
     "packages/elements": {
       "name": "@photonhealth/elements",
-      "version": "0.14.14",
+      "version": "0.16.10",
       "license": "ISC",
       "dependencies": {
         "@photonhealth/components": "*",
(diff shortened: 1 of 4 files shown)


jknapp25 (approved):

jfab3 (approved):

mrrsnn:
Need to bump the version in package elements. Pls don't merge. Will be back shortly. Stuffing my face with ramen

PR #1920, opened 2025-06-09
Fix/infinite spinner external med history

jakefromphoton:
[notion ticket](https://www.notion.so/photons/External-Med-History-items-Infinite-Spinner-2038bfbbf4ec80169c9df82a615a596e?source=copy_link)

### DIFF
--- a/packages/components/src/systems/PatientMedHistory/PatientMedHistory.test.tsx
+++ b/packages/components/src/systems/PatientMedHistory/PatientMedHistory.test.tsx
@@ -22,21 +22,9 @@ vi.mock('../PrescribeProvider', () => {
   };
 });
 
+const createQueryMock = vi.fn();
 vi.mock('../../utils/createQuery', () => ({
-  createQuery: () => {
-    const mockResponse: GetPatientResponse = {
-      patient: {
-        id: 'test-patient-id',
-        treatmentHistory: [
-          {
-            treatment: { name: 'test-treatment-name', id: 'test-treatment-id', codes: {} },
-            prescription: createPrescription()
-          }
-        ]
-      }
-    };
-    return vi.fn().mockImplementation(() => mockResponse);
-  }
+  createQuery: () => createQueryMock()
 }));
 
 afterEach(() => {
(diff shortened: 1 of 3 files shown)


packages/components/src/systems/PatientMedHistory/PatientMedHistory.test.tsx
```diff
@@ -76,11 +82,50 @@ test('PatientMedHistory allows refills', async () => {
...
+          enableLinks={false}
+          enableRefillButton={true}
+        />
+      </MockPrescribeProvider>
+    </MockSDKProvider>
+  ));
+
+  const refillButton = screen.getByRole('button', { name: 'Refill' });
+  expect(refillButton).toBeDisabled();
+  expect(refillButton).not.toHaveTextContent('Loading...');
```
jakefromphoton, line 125: This works because IconButton Spinner component shows this accessible text

packages/components/src/systems/PatientMedHistory/PatientMedHistory.test.tsx
```diff
@@ -22,21 +22,9 @@ vi.mock('../PrescribeProvider', () => {
...
-            treatment: { name: 'test-treatment-name', id: 'test-treatment-id', codes: {} },
-            prescription: createPrescription()
-          }
-        ]
-      }
-    };
-    return vi.fn().mockImplementation(() => mockResponse);
-  }
+  createQuery: () => createQueryMock()
 }));
```
jakefromphoton, line 28: Refactored this so each test can setup different med history items

packages/components/src/systems/PatientMedHistory/PatientMedHistoryTable.tsx
```diff
@@ -132,7 +132,9 @@ export default function PatientMedHistoryTable(props: PatientMedHistoryTableProp
                       }
                     }}
                     disabled={!rowItem.prescription}
-                    loading={isCreatingPrescriptionId() === rowItem.prescription?.id}
+                    loading={
```
billoran8, line 135: what was the issue? that both could return undefined/null and that's a true conditional?

billoran8 (approved):

captDaylight (approved):

packages/components/src/systems/PatientMedHistory/PatientMedHistoryTable.tsx
jakefromphoton: it should only be loading if there is a prescription. External meds have a `null`/`undefined` prescription, so the original expression `isCreatingPrescriptionId() === rowItem.prescription.id` was truthy

PR #1932, opened 2025-06-16
adding amazon price

billoran8:


apps/patient/src/views/pharmacy.utils.tsx
```diff
@@ -1,18 +1,38 @@
...
+  | undefined
+> {
   const offers = await getOffers(order.id);
 
   const amazonOffers = offers
     .filter((offer) => offer.supplier === 'AMAZON_PHARMACY')
     .filter((offer) => offer.deliveryEstimate !== undefined);
 
+  const cashOffer = offers.find((offer) => offer.cost?.type == 'CASH');
+  const insuranceOffer = offers.find((offer) => offer.cost?.type == 'INSURANCE_ESTIMATE');
```
billoran8: everything is so cute up until here lol

apps/patient/src/components/PharmacyInfo.tsx
```diff
@@ -270,7 +285,8 @@ export const PharmacyInfo = ({
   const directionsUrl = `http://maps.google.com/?q=${pharmacy?.name}, ${pharmacyFormattedAddress}`;
 
   const isAmazonPharmacy = pharmacy.id === process.env.REACT_APP_AMAZON_PHARMACY_ID;
-  const showAmazonTagline = isAmazonPharmacy && brandedOptionOverride?.amazonPharmacyOverride;
+  const showAmazonTagline =
+    isAmazonPharmacy && brandedOptionOverride?.amazonPharmacyOverride?.cashOffer?.deliveryEstimate;
```
mrrsnn: don't show the tagline if it's an amazon offer but it's an insuranceOffer?

mrrsnn (approved):
lgtm overall. a little confused on the display logic, but I'm sure it's fine.

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -179,45 +182,65 @@ export const Pharmacy = () => {
 
   useEffect(() => {
     const determineOverrides = async () => {
-      let newBrandedOptionsOverride: BrandedOptionOverrides = {};
+      let fetchedOffers: AmazonOffer[] | undefined;
```
billoran8, line 184: starting here - start point is offers - the override that gets passed down to the brandedPharmacy is derived from looking at the offers. this makes things much easier to mock out for demo mode.

apps/patient/src/components/PharmacyInfo.tsx
```diff
@@ -270,7 +285,8 @@ export const PharmacyInfo = ({
   const directionsUrl = `http://maps.google.com/?q=${pharmacy?.name}, ${pharmacyFormattedAddress}`;
 
   const isAmazonPharmacy = pharmacy.id === process.env.REACT_APP_AMAZON_PHARMACY_ID;
-  const showAmazonTagline = isAmazonPharmacy && brandedOptionOverride?.amazonPharmacyOverride;
+  const showAmazonTagline =
+    isAmazonPharmacy && brandedOptionOverride?.amazonPharmacyOverride?.deliveryEstimate;
```
jakefromphoton, line 289: Would it make sense to use `brandedOptionOverride?.amazonPharmacyOverride?.costType` instead of `brandedOptionOverride?.amazonPharmacyOverride?.deliveryEstimate` here? Since we're using this to show a cost/price.

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -178,46 +181,66 @@ export const Pharmacy = () => {
   const heading = isReroute ? t.changePharmacy : t.selectAPharmacy;
 
   useEffect(() => {
-    const determineOverrides = async () => {
-      let newBrandedOptionsOverride: BrandedOptionOverrides = {};
+    const getOffers = async () => {
+      let fetchedOffers: AmazonOffer[] | undefined;
+
+      if (offers !== undefined) {
```
jakefromphoton: nit: I think `if(offers)` would suffice and be slightly more easy to read. This looks like double negative

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -178,46 +181,66 @@ export const Pharmacy = () => {
...
-
-          newBrandedOptionsOverride = {
-            ...novocareExperimentOverride
-          };
+          fetchedOffers = [
+            {
+              costType: 'NOVOCARE_OFFER',
+              deliveryEstimate: novocareExperimentSegment
+            }
+          ];
```
jakefromphoton: Could we eventually move this into the `fetchOffers` function, or an abstraction around that function? Would like to come up with patterns to start trimming down `Pharmacy.tsx`

jakefromphoton (approved):

apps/patient/src/views/Pharmacy.tsx
billoran8: yep grea pointo

apps/patient/src/components/PharmacyInfo.tsx
billoran8: imo no! we're also using it to display a delivery estimate which may or may not always have a price associated with it - there should always be the delivery estimate

apps/patient/src/views/Pharmacy.tsx
billoran8: agree - this file is HUGE. good point, i'll put it int here.

billoran8: look at how much simpler the logic is

billoran8:
NOTE: changed this so as not to actually show price because we're waiting on some additional experimentation. don't want this refactor to block anythign though.

apps/patient/src/components/BrandedOptions.tsx
```diff
@@ -11,8 +11,14 @@ interface Props {
   handleSelect: (id: string) => void;
 }
 
+export interface AmazonOffer {
```
jakefromphoton, line 14: Why prefix with `Amazon`? I figured the intent would be other offers from other organizations would share the offer data type/structure

apps/patient/src/components/PharmacyInfo.tsx
```diff
@@ -314,6 +330,13 @@ export const PharmacyInfo = ({
             ) : null}
           </VStack>
         ) : null}
+
+        {isAmazonPharmacy && showAmazonTagline ? (
```
jakefromphoton: could just be `showAmazonTagline` since isAmazonPharmacy defines that variable

apps/patient/src/components/PharmacyInfo.tsx
jakefromphoton: Ah ok I didn't see that `showAmazonTagline` also shows/hides the Delivery Estimate.

I figure someday we'll show offers with a price and no delivery estimate. So we'll probably want a pattern that is `showOfferPrice` and `showOfferDeliveryEstimate`. Not sure I like Tagline or have heard anyone use that language

apps/patient/src/components/BrandedOptions.tsx
billoran8: long way to go imo!  we still haven't decided on an exact schema to use for offers so i just wanna make this clear this is for amazon specifically

apps/patient/src/components/PharmacyInfo.tsx
billoran8: yep good point

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -621,13 +624,14 @@ export const Pharmacy = () => {
...
-                .replace(/_+/g, '_')
-                .replace(/^_+|_+$/g, '');
+              const slugifiedOverride =
+                brandedOptionsOverride?.amazonPharmacyOverride.deliveryEstimate
+                  ?.toLowerCase()
+                  .trim()
+                  .replace(/\s+/g, '_')
+                  .replace(/[^a-z0-9_]/g, '')
+                  .replace(/_+/g, '_')
+                  .replace(/^_+|_+$/g, '');
```
jakefromphoton, line 634: Will this be useful for future offers? If so, seems like a good candidate for a util function `getSlugifiedOverride(brandedOptionsOverride?.amazonPharmacyOverride.deliveryEstimate)`. That should help with the sloppy diff going forward too since this logic didnt really change

apps/patient/src/components/BrandedOptions.tsx
jakefromphoton: aim high! it's a nit though. We just want to make sure on next offer we dont establish this as a pattern

jakefromphoton:
Looks good to me. My comments are just style things.

apps/patient/src/views/pharmacy.utils.tsx
```diff
@@ -1,26 +1,51 @@
...
+  // measured will only want to show amazon offers if we do not have a novocare offer
+  if (order.organization.id === 'org_pcPnPx5PVamzjS2p') {
+    if (novocareOffers.length === 0) {
+      return amazonOffers;
+    } else {
+      return novocareOffers;
+    }
   }
+
+  return [...amazonOffers, ...novocareOffers];
```
jakefromphoton, line 44: Lovin this file

PR #1937, opened 2025-06-24
Setup patient app tests

jakefromphoton:
This PR's intent is to 

1. start a discussion around how we want to test the Patient App
2. provide at least 1 way in the short term for devs to be able to write automated tests, and facilitate a better dev workflow locally

This approach uses React Testing Library and Vitest mocks to isolate the functionality of pages/components in the Patient App. The assertions inspect what is rendered into the jsdom. My intent is to have tests that can verify the showing/hiding of different features (price toggle), not check styles like color or font size.

I used claude to generate most of this PR because setting up the initial test harness is usually annoying. But going forward, i don't expect Claude to be required to make more of these or maintain these.

References:
* https://vitest.dev/api/vi.html#mock-modules

### DIFF
--- a/apps/patient/src/setupTests.ts
+++ b/apps/patient/src/setupTests.ts
@@ -1,0 +1,20 @@
+// jest-dom adds custom jest matchers for asserting on DOM nodes.
+// allows you to do things like:
+// expect(element).toHaveTextContent(/react/i)
+// learn more: https://github.com/testing-library/jest-dom
+import '@testing-library/jest-dom';
+import { vi } from 'vitest';
+
+vi.mock('react-ga4', () => ({
+  event: vi.fn()
+}));
+
+vi.mock('@datadog/browser-rum', () => ({
+  datadogRum: {
+    addAction: vi.fn()
+  }
+}));
+
+vi.mock('@client/settings', () => ({
+  getOrgMailOrderPharms: vi.fn(() => ({ patient: [] }))
+}));
--- a/apps/patient/src/setupTests.ts
+++ b/apps/patient/src/setupTests.ts
@@ -1,7 +1,3 @@
-// jest-dom adds custom jest matchers for asserting on DOM nodes.
-// allows you to do things like:
-// expect(element).toHaveTextContent(/react/i)
-// learn more: https://github.com/testing-library/jest-dom
 import '@testing-library/jest-dom';
 import { vi } from 'vitest';
 
--- a/apps/patient/src/views/Canceled.test.tsx
+++ b/apps/patient/src/views/Canceled.test.tsx
@@ -1,0 +1,66 @@
+import { render, screen } from '@testing-library/react';
+import { ChakraProvider } from '@chakra-ui/react';
+import { Helmet } from 'react-helmet';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { Canceled } from './Canceled';
+import { useOrderContext } from './Main';
+
+vi.mock('./Main', () => ({
+  useOrderContext: vi.fn()
+}));
+
+vi.mock('../components', () => ({
+  PrescriptionsList: () => <div data-testid="prescriptions-list">Prescriptions List</div>
+}));
+
+const mockUseOrderContext = vi.mocked(useOrderContext);
+
+const mockOrder = {
+  patient: {
+    name: {
+      full: 'John Doe'
+    }
+  }
+};
+
+const renderCanceled = () => {
+  return render(
+    <ChakraProvider>
+      <Canceled />
+    </ChakraProvider>
+  );
+};
+
+describe('Canceled', () => {
+  beforeEach(() => {
+    mockUseOrderContext.mockReturnValue({
+      order: mockOrder
+    } as any);
+  });
+
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('sets the correct page title', () => {
+    renderCanceled();
+    const helmet = Helmet.peek();
+    expect(helmet.title).toBe('This order was canceled.');
+  });
+
+  it('renders the canceled order heading', () => {
+    renderCanceled();
+    expect(screen.getByRole('heading', { name: 'This order was canceled.' })).toBeInTheDocument();
+  });
+
+  it('displays the patient name', () => {
+    renderCanceled();
+    expect(screen.getByText('John Doe')).toBeInTheDocument();
+    expect(screen.getByText(/patient/i)).toBeInTheDocument();
+  });
+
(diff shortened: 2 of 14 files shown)


apps/patient/src/views/Canceled.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { Canceled } from './Canceled';
+import { useOrderContext } from './Main';
+
+vi.mock('./Main', () => ({
+  useOrderContext: vi.fn()
+}));
+
+vi.mock('../components', () => ({
+  PrescriptionsList: () => <div data-testid="prescriptions-list">Prescriptions List</div>
+}));
```
jakefromphoton: I normally wouldn't mock this, since I don't like mocking our own code. But mocking like this should be useful to help us get started chipping away at the patient app

jakefromphoton:
Mobbed with Bill and Nick on this for an hour. We discussed how we want to approach these tests going forward. Will need to get @captDaylight feedback too when he returns.

* We prefer not to mock things
* This PR mocks some things we'd prefer not to, such as `../components`, but we need more refactoring work to reduce the mocking of things we own
* We aren't sure if/how to document these "best practices". Can discuss at retro

jakefromphoton:
Closing to convert this into an RFC with alternative options for the team to discuss per @billoran8 recommendation

apps/patient/src/views/Canceled.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return render(
+    <ChakraProvider>
+      <Canceled />
+    </ChakraProvider>
+  );
+};
+
+describe('Canceled', () => {
+  beforeEach(() => {
+    mockUseOrderContext.mockReturnValue({
```
billoran8: didn't we come to the realization that we didn't need to mock the context at all?

apps/patient/src/views/Canceled.test.tsx
billoran8: fine with that, we're only mocking it because of poorly designed components that'll explode otherwise; in a perfect situation we wouldn't need to mock

apps/patient/src/views/Canceled.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+    } as any);
+  });
+
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('sets the correct page title', () => {
+    renderCanceled();
+    const helmet = Helmet.peek();
```
billoran8: lol what is helmet? for rendering html?

billoran8: react libs always have the cutest names

billoran8: tbh surprised helmet peek doesn't need to be awaited? wonder how itw orks under the hood

apps/patient/src/views/Canceled.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  it('renders the canceled order heading', () => {
+    renderCanceled();
+    expect(screen.getByRole('heading', { name: 'This order was canceled.' })).toBeInTheDocument();
+  });
+
+  it('displays the patient name', () => {
+    renderCanceled();
+    expect(screen.getByText('John Doe')).toBeInTheDocument();
+    expect(screen.getByText(/patient/i)).toBeInTheDocument();
```
billoran8: super nit: this shoudl be another test that is like "displays the patient title" or someshit

apps/patient/src/views/Canceled.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  it('renders the prescriptions list component', () => {
+    renderCanceled();
+    expect(screen.getByTestId('prescriptions-list')).toBeInTheDocument();
```
billoran8: is testId something that should only be used when making render wrappers in tests?

apps/patient/src/views/Pharmacy.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+  fills,
+  readyBy: 'Regular',
+  state: 'CANCELED',
+  isReroutable: false,
+  exceptions: [],
+  fulfillments: [],
+  discountCards: []
+});
+
+const createMockFill = (treatmentName: string) => ({
```
billoran8: nit: this isn' really a mock is it?

apps/patient/src/setupTests.ts
```diff
@@ -0,0 +1,20 @@
+// jest-dom adds custom jest matchers for asserting on DOM nodes.
+// allows you to do things like:
+// expect(element).toHaveTextContent(/react/i)
```
billoran8: this commenet can prob be removed right

apps/patient/src/views/Canceled.test.tsx
jakefromphoton: good catch. claude added this, but we can just use `getByText`. I only use testID as a last resort

apps/patient/src/views/Canceled.test.tsx
jakefromphoton: The "Patient" title-like text is hardcoded without any `if-else` conditional,  so i wouldn't actually want to test that part. Removing and just keeping the patient name part

apps/patient/src/views/Canceled.test.tsx
jakefromphoton: yea we cleaned it up for Pharmacy. Will update Canceled

apps/patient/src/views/Canceled.test.tsx
jakefromphoton: Helmet makes it easier in SPAs to update the html doc head. [See docs if you like](https://www.npmjs.com/package/react-helmet)

That said, I dont really care about this or want people to think it's something necesarily worth testing. Removing it.

apps/patient/src/views/Pharmacy.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+    });
+
+    describe('when order contains GLP-1 medication', () => {
+      beforeEach(async () => {
+        const glp1MedicationName = 'Semaglutide';
+        const glpFill = generateFlattenedFill({
+          treatment: generateTreatment({ name: glp1MedicationName })
+        });
+
+        await renderPharmacy({ flattenedFills: [glpFill] });
```
billoran8, line 81: with the way `renderPharmacy` and other render fucntions get called these tests prob can't be run parallel right? is that a problem?

billoran8 (approved):

PR #1941, opened 2025-06-26
Show coupons by default

jakefromphoton:
- Coupon Prices toggle on by default
  - added test for this feature
- Moved and renamed component so it could be unmocked
- moved formatter utils to separate file

[notion ticket](https://www.notion.so/photons/Make-Discount-Cards-Show-by-Default-for-All-Orders-2078bfbbf4ec8091ba7fe8d3dd81fa4e?source=copy_link)


apps/patient/src/views/Pharmacy.test.tsx
```diff
@@ -32,7 +33,6 @@ vi.mock('../components', () => ({
     <div data-testid="fixed-footer">{children}</div>
   ),
   LocationModal: () => <div data-testid="location-modal">Location Modal</div>,
-  PickupOptions: () => <div data-testid="pickup-options">Pickup Options</div>,
```
jakefromphoton, line 35: No longer need to mock this because it's in a separate export folder

apps/patient/src/utils/formatters.test.ts
```diff
@@ -0,0 +1,6 @@
+import { formatPrice } from './formatters';
+
+test('formatPrice', () => {
```
billoran8: nit can these expects be in individual tests describing the output adn why the output is correct?

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -841,7 +842,7 @@ export const Pharmacy = () => {
     (enableCourier || enableMailOrder || brandedOptionsOverride !== undefined) ?? false;
 
   const pickupPharmacyOptions = (location: string) => (
-    <PickupOptions
+    <PickupPharmacyCardList
```
billoran8, line 847: good job clarifying the name here

apps/patient/src/utils/formatters.test.ts
jakefromphoton: Honestly i dont like that and the more we do it the more setups are going to repeat and slow the test suite down.

jakefromphoton: I'm doing it for now but going to push in the future that we don't force people to do it that way. Also increases nesting and reduces readability IMO

jakefromphoton: I guess to clarify, i am doing it in the Pharmacy.test.tsx but i think the file has already become a bit of a mess

jakefromphoton: this number formatter is dead simple, so sure.

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -126,7 +126,7 @@ export const Pharmacy = () => {
     openNow !== null ? !!openNow : order?.readyBy === 'Urgent'
   );
   const [enable24Hr, setEnable24Hr] = useState(order?.readyBy === 'After hours');
-  const [enablePrice, setEnablePrice] = useState(false);
+  const [enablePrice, setEnablePrice] = useState(true);
```
jakefromphoton, line 129: main feature change...

captDaylight (approved):

PR #1948, opened 2025-07-02
fixing an issue that sometimes caused amazon to show up as an option when it shouldn't

billoran8:


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -206,10 +206,15 @@ export const Pharmacy = () => {
...
-    if (enablePrice) {
-      amazonPharmacyOverride = { ...cashOffer, costAmount: undefined };
-    } else {
-      amazonPharmacyOverride = { ...insuranceOffer, costAmount: undefined };
+
+    // we'll only want to set the override
+    // if we have at least one offer
+    if (insuranceOffer || cashOffer) {
+      if (enablePrice) {
+        amazonPharmacyOverride = { ...cashOffer, costAmount: undefined };
```
captDaylight, line 214: is it possible that `enablePrice` is true but `cashOffer` is `undefined`? (or vice versa with `insuranceOffer`)

Spreading undefined will cause an error

captDaylight (approved):

apps/patient/src/views/Pharmacy.tsx
billoran8: we just checked and it doesn't actually spread an error unless the value is undefined!

PR #1952, opened 2025-07-08
amazon offers will show price if they come back from the endpoint now

billoran8:


mrrsnn (approved):

billoran8:
note: this cannot be deployed until https://github.com/Photon-Health/services/pull/1090 goes out on the backend

PR #1957, opened 2025-07-10
branded options in the patient experience cannot be selected if they're already attached to the order

billoran8:


apps/patient/src/components/BrandedPharmacyCard.tsx
```diff
@@ -14,6 +14,7 @@ import { BrandedOptionOverrides } from './BrandedOptions';
 
 interface Props {
   pharmacyId: string;
+  isCurrentPharmacy: boolean;
```
mrrsnn: nit: the property name here is too business logic-y for a "view" component. This property could be called `disabled?: boolean`, or something.

apps/patient/src/components/BrandedOptions.tsx
```diff
@@ -7,6 +7,7 @@ interface Props {
   options: string[];
   location: string;
   selectedId: string;
+  currentPharmacyId?: string;
```
mrrsnn: another nit: `excludedPharmacyId?: string` could a better choice. I could see myself diving into the code at the level below this and be thinking, why are we disabling the current pharmacy. It's ambiguously similar to `selectedId`

mrrsnn (approved):

apps/patient/src/components/BrandedOptions.tsx
billoran8: yep great point; was just trying to copy the standard set in the localPickup version of the widget but i agree that this clarifies intent much clearer

apps/patient/src/components/BrandedPharmacyCard.tsx
billoran8: oh now THIS i have opinions about

apps/patient/src/components/BrandedOptions.tsx
billoran8: renamed

apps/patient/src/components/BrandedPharmacyCard.tsx
billoran8: i see now it's more about having a 'state'

PR #1962, opened 2025-07-29
Update copy for delayed provider-selected pharmacy orders

jakefromphoton:
Copy change to reduce confusion for Providers when we introduce the 15 minute delay to sending orders for provider-selected pharmacy orders.

~~We should deploy this after the Lambda PR is merged and deployed: https://github.com/Photon-Health/lambdas/pull/2145~~
Edit: We can release this whenever because the if predicates won't affect anything until the lambdas are released. The provider-selected pharmacies will be in PENDING until we release the lambda changes. For send-to-patient, the order will be in ROUTING but the pharmacyID will be `null`, so the old copy will continue to show.

[notion ticket](https://www.notion.so/photons/Add-delay-to-provider-selected-pharmacy-orders-2378bfbbf4ec800dbe58c8745ced3c0a?source=copy_link)

image

### DIFF
--- a/apps/app/src/views/routes/Order.tsx
+++ b/apps/app/src/views/routes/Order.tsx
@@ -205,7 +205,7 @@ Order:
 
 ----
 Patient:
-  ID: ${order?.patient?.id} 
+  ID: ${order?.patient?.id}
   Name: ${order?.patient?.name?.full}
 
 ----
@@ -222,8 +222,8 @@ ${fills.map(
 `
 )}
 
----- 
-Description: 
+----
+Description:
   `;
 }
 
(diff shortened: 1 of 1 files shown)


jakefromphoton:
Note: without this change, the provider would see this messaging:

image

apps/app/src/views/routes/Order.tsx
```diff
@@ -840,3 +842,24 @@ export const Order = () => {
     </>
   );
 };
+
+function getRoutingOrderStatusText(order: OrderType) {
```
billoran8, line 846: such cute snackable functions

billoran8 (approved):

apps/app/src/views/routes/Order.tsx
captDaylight: SO SNACKABLE

mrrsnn (approved):
beauty

PR #1969, opened 2025-07-30
Barebones Patient Analytics (via Rudderstack)

captDaylight:


apps/patient/src/configs/analytics.ts
```diff
@@ -0,0 +1,39 @@
+import { ApiObject, RudderAnalytics } from '@rudderstack/analytics-js';
+
+const RUDDERSTACK_WRITE_KEY = process.env.REACT_APP_RUDDERSTACK_WRITE_KEY;
+const RUDDERSTACK_DATA_PLANE_URL = process.env.REACT_APP_RUDDERSTACK_DATA_PLANE_URL;
+const ENVIRONMENT = process.env.REACT_APP_ENV_NAME || 'development';
+
+export class PatientAnalytics {
```
captDaylight, line 7: I'll be extending this more, but bill, is it cute enough?

apps/patient/src/configs/analytics.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (!RUDDERSTACK_WRITE_KEY || !RUDDERSTACK_DATA_PLANE_URL) {
+      throw new Error('RudderStack write key and data plane URL are required');
+    }
+
+    this.rudderanalytics = new RudderAnalytics();
+    this.rudderanalytics.load(RUDDERSTACK_WRITE_KEY || '', RUDDERSTACK_DATA_PLANE_URL || '');
+    this.environment = ENVIRONMENT;
+  }
+
+  page(category: string, name?: string, properties: ApiObject = {}) {
```
captDaylight, line 22: starting only with this method to verify pipes are working and keep this PR manageable

apps/patient/src/views/Canceled.tsx
```diff
@@ -4,11 +4,18 @@ import { Helmet } from 'react-helmet';
...
+import { patientAnalytics } from '../configs/analytics';
+import { useEffect } from 'react';
+import { useLocation } from 'react-router-dom';
 
 export const Canceled = () => {
   const {
     order: { patient }
   } = useOrderContext();
+  const location = useLocation();
+  useEffect(() => {
```
captDaylight, line 16: Declaring page analytics info in each file, could have done somethnig smart lke a custom hook but I think this is simpler and more straightforward

apps/patient/src/configs/analytics.ts
billoran8: v cute! imo it would be even cuter if all metrics sent to rudderstack were also sent to datadog (it might help us explore metrics in a way that rudderstack can't?)

apps/patient/src/configs/analytics.ts
billoran8: neat - we'll also need an `action` function yeah?

billoran8 (approved):

apps/patient/src/configs/analytics.ts
captDaylight: that would be a cute follow on for sure

apps/patient/src/configs/analytics.ts
captDaylight: yup!

PR #1975, opened 2025-08-04
Tracking patient interaction events from Patient App

captDaylight:
the thing I'm not sure about is having a naming convention for the events

### DIFF
--- a/apps/patient/src/components/order-details/OrderDetailsModal.tsx
+++ b/apps/patient/src/components/order-details/OrderDetailsModal.tsx
@@ -15,6 +15,7 @@ import {
 import dayjs from 'dayjs';
 import { ReactNode } from 'react';
 import { MdOutlineLocalPharmacy } from 'react-icons/md';
+import { patientAnalytics } from '../../configs/analytics';
 
 export interface PrescriptionData {
   rxName: string;
@@ -70,8 +71,16 @@ const defaultIcon = (
 );
 
 export const OrderDetailsModal = (props: OrderDetailsProps & OrderDetailsModalProps) => {
+  const handleClose = () => {
+    patientAnalytics.track('Close Order Details Modal', {
+      pharmacyName: props.pharmacyName,
+      prescriptionCount: props.prescriptions.length
+    });
+    props.onClose();
+  };
+
   return (
-    <Modal isOpen={props.isOpen} onClose={props.onClose} size="full">
+    <Modal isOpen={props.isOpen} onClose={handleClose} size="full">
       <ModalOverlay />
       <ModalContent backgroundColor="gray.100" alignItems="center" w="full">
         <ModalHeader>Order Details</ModalHeader>
--- a/apps/patient/src/configs/analytics.ts
+++ b/apps/patient/src/configs/analytics.ts
@@ -49,12 +49,13 @@ export class PatientAnalytics {
     this.rudderanalytics.track(eventName, trackProperties);
   }
 
-  identify(userId: string) {
+  identify({ userId, orgId, orgName }: { userId: string; orgId: string; orgName: string }) {
     if (!this.rudderanalytics) {
       return;
     }
 
     this.rudderanalytics.identify(userId);
+    this.rudderanalytics.group(orgId, { name: orgName });
   }
 }
 
--- a/apps/patient/src/configs/analytics.ts
+++ b/apps/patient/src/configs/analytics.ts
@@ -49,12 +49,22 @@ export class PatientAnalytics {
     this.rudderanalytics.track(eventName, trackProperties);
   }
 
-  identify({ userId, orgId, orgName }: { userId: string; orgId: string; orgName: string }) {
+  identify({
+    userId,
+    patientProperties,
+    orgId,
+    orgName
+  }: {
+    userId: string;
+    patientProperties: ApiObject;
+    orgId: string;
+    orgName: string;
+  }) {
     if (!this.rudderanalytics) {
       return;
     }
 
-    this.rudderanalytics.identify(userId);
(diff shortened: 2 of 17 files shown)


apps/patient/src/components/order-details/OrderDetailsModal.tsx
```diff
@@ -70,8 +71,16 @@ const defaultIcon = (
 );
 
 export const OrderDetailsModal = (props: OrderDetailsProps & OrderDetailsModalProps) => {
+  const handleClose = () => {
+    patientAnalytics.track('Close Order Details Modal', {
```
billoran8, line 75: i was thinking about this - does it make sense for it to be `identify` instead of `track`? `track` is mostly used for anonymous actions

apps/patient/src/components/order-details/OrderDetailsModal.tsx
```diff
@@ -70,8 +71,16 @@ const defaultIcon = (
...
+      pharmacyName: props.pharmacyName,
+      prescriptionCount: props.prescriptions.length
```
billoran8, line 77: can we get organization name and/or organization id in everything too?

billoran8: as well as medication name/medication id?

apps/patient/src/components/order-details/OrderDetailsModal.tsx
billoran8: also i'm wondering if we should have different categories of events too - closing a modal is sort of irrelevant for things but like, clicking on a pharmacy is a `Big Deal` - would be nice to have a way to easily distinguish between the two types

apps/patient/src/views/Main.tsx
```diff
@@ -57,10 +57,14 @@ export const Main = () => {
   const settings = order?.organization.settings;
 
   useEffect(() => {
-    if (order?.patient.id) {
-      patientAnalytics.identify(order.patient.id);
+    if (order?.patient.id && order?.organization.id && order?.organization.name) {
+      patientAnalytics.identify({
+        userId: order.patient.id,
```
billoran8, line 62: maybe more info about the patient - state? zip code? provider name?

billoran8 (approved):
some questions for thought but don't need to be answered now

apps/patient/src/components/order-details/OrderDetailsModal.tsx
captDaylight: 1. I like the terms `identify` and `track` as they are, the former is identifying the patient by their id and the latter is capturing an event to be tracked. Where is the context of `track` for anon actions coming from?
2. I'd assume that we emphasize importance of certain events downstream and not try and categorize them upstream. Focus more on capturing anything and everything and untangle later

apps/patient/src/views/Main.tsx
captDaylight: Yup we totally can, this was a question more I had about PHI... but since it's just passing through to our own service it should be fine. But does it mean that this new DB and s3 need to have hipaa compliance?

apps/patient/src/components/order-details/OrderDetailsModal.tsx
captDaylight: Yes, we can, what I'm unsure of I've added a `group` call in the identify method, which associates an org id and org name... I'd assume it would attach it with a session that we can track

apps/patient/src/components/order-details/OrderDetailsModal.tsx
billoran8: nevermind you're right i got some wires crossed. what about adding a `[group](https://www.rudderstack.com/docs/event-spec/standard-events/group/)` call for the organization?

billoran8: agree with capture everything and detangle later - but right now there's now way for us to detangle without using a hardcoded list of event names, which sounds weird!

apps/patient/src/views/Main.tsx
billoran8: IMO state, zipcode and provider name aren't personally identifiable infromation so we should be fine? there's no way for me to link anything to a specific human (email address, full name, etc?)

apps/patient/src/components/order-details/OrderDetailsModal.tsx
billoran8: yep i think it does that - we'll see exactly how things are linked together once dta starts flowin

apps/patient/src/components/order-details/OrderDetailsModal.tsx
captDaylight: Group is in there :) I lumped it in with the identify method since I'll have that info at the same time.

apps/patient/src/configs/analytics.ts
```diff
@@ -49,12 +49,13 @@ export class PatientAnalytics {
...
   }
 
-  identify(userId: string) {
+  identify({ userId, orgId, orgName }: { userId: string; orgId: string; orgName: string }) {
     if (!this.rudderanalytics) {
       return;
     }
 
     this.rudderanalytics.identify(userId);
+    this.rudderanalytics.group(orgId, { name: orgName });
```
billoran8, line 68: wondering how the `group` property manifests downstream.  wonder if we should have a `group` call for everything we might want to group by? date, week, month, state, etc?

apps/patient/src/components/order-details/OrderDetailsModal.tsx
billoran8: love group!

PR #1978, opened 2025-08-05
tracking offer impressions and selections

billoran8:


billoran8:
pharamcy select desperately eneds a refactor, let;'s do that later htough

apps/patient/src/utils/tracking/OfferImpressionTracker.tsx
```diff
@@ -0,0 +1,40 @@
...
+  isAlreadySelected
+}: {
+  children: React.ReactNode;
+  pharmacy: EnrichedPharmacy;
+  ordinalPosition: number;
+  isAlreadySelected: boolean;
+}) => {
+  const { ref } = useInView({
+    triggerOnce: true,
+    rootMargin: '-100px',
```
jakefromphoton, line 19: Did you already consider threshold instead of rootMargin? This pixel value makes me hesitate.

[options docs](https://github.com/thebuilder/react-intersection-observer/tree/1e8a3b7710f375b2d9cd321d50318450426d6c59?tab=readme-ov-file#options)

apps/patient/src/views/pharmacy.utils.tsx
```diff
@@ -108,3 +112,44 @@ export function determineNovocareExperimentSegment(order: Order): string | undef
...
 }
+
+export function getPharmacy(
+  allPharmacies: EnrichedPharmacy[],
+  selectedId: string
+): {
+  type: ExtendedFulfillmentType;
+  selectedPharmacy: { id: string; name: string } | PharmacyType | undefined;
+} {
+  // Fudge it so that we can show the pharmacy card on initial load of the
```
jakefromphoton, line 123: fudge it

jakefromphoton: Lol props for preserving this comment

apps/patient/src/utils/tracking/OfferImpressionTracker.tsx
jakefromphoton: If the rootMargin trigger works well locally then I would be cool with it.

billoran8: it does work locally - happy to change this but just didn't see any reason to! happy to demo as well

apps/patient/src/views/pharmacy.utils.tsx
billoran8: this entire file and experience neesd a significant refactoring - i'm not going to waste too much effort doing it now because we should strategize first

billoran8: just needed this funcion here because there's no other way to get the pharmacy name of a branded pharmacy (aka amazon)

apps/patient/src/utils/tracking/OfferImpressionTracker.tsx
```diff
@@ -0,0 +1,40 @@
...
+  pharmacy,
+  ordinalPosition,
+  isAlreadySelected
+}: {
+  children: React.ReactNode;
+  pharmacy: EnrichedPharmacy;
+  ordinalPosition: number;
+  isAlreadySelected: boolean;
+}) => {
+  const { ref } = useInView({
```
mrrsnn, line 17: I feel like the native `IntersectionObserver` api is pretty good! Do we really need to add another dependency package for this?

billoran8: this seems to be just a cute lil wrapper with a hook around it!

billoran8: (this def does use IntersectionObserver under the hood, just provides a more react-hooksy like way ot interact with it)

jakefromphoton (approved):

PR #1979, opened 2025-08-06
Fix multiRx orders showing prices

jakefromphoton:
[notion ticket](https://www.notion.so/photons/Discount-card-prices-showing-on-multi-med-orders-2468bfbbf4ec8078990ef06556c8a14f?source=copy_link)


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -127,7 +127,7 @@ export const Pharmacy = () => {
     openNow !== null ? !!openNow : order?.readyBy === 'Urgent'
   );
   const [enable24Hr, setEnable24Hr] = useState(order?.readyBy === 'After hours');
-  const [enablePrice, setEnablePrice] = useState(true);
+  const [enablePrice, setEnablePrice] = useState(showPriceToggle);
```
billoran8, line 130: oh wow oso that was the issue? did i do that?

jakefromphoton: No, i did that when i enabled it by default several weeks ago

billoran8 (approved):

PR #1987, opened 2025-08-06
Add more metadata to offer events

billoran8:


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -598,7 +598,7 @@ export const Pharmacy = () => {
       }
 
       if (shouldTrackOfferImpressionsAndSelections) {
-        patientAnalytics.track('Offer Selected', pharmacies[index]);
+        patientAnalytics.track('Offer Clicked', pharmacies[index]);
```
billoran8, line 601: renaming this event - i misunderstood what this function was. it gets called when someone clicks an offer/pharmacy but before they click submitted

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -742,7 +742,9 @@ export const Pharmacy = () => {
             if (shouldTrackOfferImpressionsAndSelections) {
               patientAnalytics.track('Offer Selected', {
                 ...selectedPharmacy,
-                ...extraOfferMetadata
+                ...extraOfferMetadata,
+                orderId: order.id,
```
billoran8, line 746: let's get the order and org id so we can aggregate

apps/patient/src/utils/tracking/OfferImpressionTracker.tsx
```diff
@@ -30,7 +33,9 @@ const OfferImpressionTracker = ({
           showReadyIn30Min: pharmacy.showReadyIn30Min,
           is24Hr: pharmacy.is24Hr,
           isClosingSoon: pharmacy.isClosingSoon,
-          isAlreadySelected: isAlreadySelected
+          isAlreadySelected: isAlreadySelected,
+          orderId: order.id,
```
billoran8, line 37: getting order and org so we can aggregate

jakefromphoton (approved):

PR #1990, opened 2025-08-06
adding more metadata

billoran8:


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -598,7 +598,12 @@ export const Pharmacy = () => {
       }
 
       if (shouldTrackOfferImpressionsAndSelections) {
-        patientAnalytics.track('Offer Clicked', pharmacies[index]);
+        patientAnalytics.track('Offer Clicked', {
+          ...pharmacies[index],
```
mrrsnn, line 602: does this not include the pharmacyId?

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -740,11 +745,22 @@ export const Pharmacy = () => {
...
 
             if (shouldTrackOfferImpressionsAndSelections) {
+              const brandedOptionObjects = brandedOptions.map((id) => ({
+                id
+              }));
+
               patientAnalytics.track('Offer Selected', {
                 ...selectedPharmacy,
                 ...extraOfferMetadata,
+                pharmacyId: selectedId,
```
mrrsnn, line 755: does `selectedPharmacy` not already have `pharmacyId`?

mrrsnn (approved):
just questions, nothing blocking.

apps/patient/src/views/Pharmacy.tsx
billoran8: it does, but it's stored as `id` and gets overwritten in rudderstack by its own internal id

apps/patient/src/views/Pharmacy.tsx
billoran8: it has `id` not `pharmacyId`, and when the `id` enters rudderstacks system it gets overwritten by the `event` id

PR #1994, opened 2025-08-11
Preparatory refactors for Patient App navigation rework

jakefromphoton:
* add navigation test for current patient app navigation (`review` -> `readyAt` -> `pharmacy[search]` -> `status`)
  * uses a journey testing pattern
* Extract routes to separate file so that a test can import them and use a MemoryRouter
* Move more pharmacy list components (Branded) out of `components` to enable more fine-grained vi.mocks

Preparing for [notion ticket](https://www.notion.so/Move-requested-pick-up-time-after-a-local-pharmacy-is-selected-2398bfbbf4ec809d87d7e846bbbccca3?source=copy_link)

### DIFF
--- a/apps/patient/src/Navigation.test.tsx
+++ b/apps/patient/src/Navigation.test.tsx
@@ -20,6 +20,7 @@ vi.mock('./api', () => ({
   }),
   getOrder: vi.fn(),
   getPharmacies: vi.fn().mockResolvedValue({ pharmaciesByLocation: [] }),
+  getOffers: vi.fn().mockResolvedValue([]),
   rerouteOrder: vi.fn(),
   setOrderPharmacy: vi.fn(),
   setPreferredPharmacy: vi.fn(),
@@ -30,8 +31,7 @@ vi.mock('./api', () => ({
 vi.mock('./configs/graphqlClient', () => ({
   setAuthHeader: vi.fn(),
   graphQLClient: {
-    request: vi.fn().mockResolvedValue({}),
-    GetOffersForOrder: vi.fn().mockResolvedValue({ offers: [] })
+    request: vi.fn().mockResolvedValue({})
   },
   gqlSerializer: {
     parse: vi.fn(),
--- a/apps/patient/src/Navigation.test.tsx
+++ b/apps/patient/src/Navigation.test.tsx
@@ -1,0 +1,122 @@
+import { render, screen, waitFor } from '@testing-library/react';
+import { afterEach, describe, expect, vi } from 'vitest';
+import { createMemoryRouter, createRoutesFromElements, RouterProvider } from 'react-router-dom';
+import { OrderContextType } from './views/Main';
+import {
+  generateFill,
+  generateFulfillment,
+  generateOrder,
+  generatePatient,
+  generatePharmacy
+} from './test-utils/generators';
+import userEvent from '@testing-library/user-event';
+import { routeElements } from './Routes';
+
+vi.mock('./api', () => ({
+  geocode: vi.fn().mockResolvedValue({
+    lat: 40.7128,
+    lng: -74.006,
+    address: '123 Main St, New York, NY 10001'
+  }),
+  getOrder: vi.fn(),
+  getPharmacies: vi.fn().mockResolvedValue({ pharmaciesByLocation: [] }),
+  rerouteOrder: vi.fn(),
+  setOrderPharmacy: vi.fn(),
+  setPreferredPharmacy: vi.fn(),
+  triggerDemoNotification: vi.fn(),
+  AUTH_HEADER_ERRORS: []
+}));
+
+vi.mock('./configs/graphqlClient', () => ({
+  setAuthHeader: vi.fn(),
+  graphQLClient: {
+    request: vi.fn().mockResolvedValue({}),
+    GetOffersForOrder: vi.fn().mockResolvedValue({ offers: [] })
+  },
+  gqlSerializer: {
+    parse: vi.fn(),
+    stringify: vi.fn()
+  }
+}));
+
+vi.mock('@datadog/browser-rum');
+vi.mock('./configs/analytics');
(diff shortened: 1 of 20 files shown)


apps/patient/src/Navigation.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+    patient: generatePatient({ name: { full: 'Jane Doe' } }),
+    fills: [generateFill('test-treatment')],
+    fulfillments: [generateFulfillment({ state: 'PROCESSING' })]
+  });
+
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  test('navigates from review page to readyBy to pharmacy to status', async () => {
```
jakefromphoton, line 74: I plan to use this test as a baseline to modify from, when we update the navigation to be conditional like in [the figjam](https://www.figma.com/board/zweQpWiOAJee9P3inYDheM/Patient-App---Pharmacy-Select-Logic-Flow?node-id=3-581&t=0Jf0JlXqGvYiDOm6-4) on the notion ticket:

image

apps/patient/src/Navigation.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+    await userEvent.click(screen.getByText('Select pharmacy'));
+    expect(setOrderPharmacyMock).toHaveBeenCalledWith(
+      'ord_testId777',
+      'phr_testId123',
+      'Urgent',
+      'Today',
+      expect.anything(),
+      true
+    );
+    await waitFor(() => screen.findByText('Preparing order...'), { timeout: 2500 });
```
jakefromphoton, line 111: Unfortunately, for now, we need a 2500 timeout because of the nested setTimeouts with 1000 delays in the Pharmacy component.

I tried use `vi.useFakeTimers()` but that was causing the whole test to blow up. We can look into a better setup for this later

package.json
```diff
@@ -174,5 +174,10 @@
   },
   "optionalDependencies": {
     "@rollup/rollup-linux-x64-gnu": "^4.37.0"
+  },
+  "overrides": {
+    "@chakra-ui/react": {
+      "@zag-js/focus-visible": "^1.18.0"
+    }
```
jakefromphoton, line 181: Updated this because the default version of Zag used by Chakra had errors that popped up during tests, such as `TypeError: Cannot set property  focus of #<HTMLElement> which has only a getter` where Zag was trying to set a readonly property

apps/patient/src/views/ReadyBy.tsx
```diff
@@ -107,7 +107,7 @@ export const ReadyBy = () => {
   useEffect(() => {
     if (selectedTime) {
       // Scroll to bottom to make sure selection isn't hidden by footer
-      window.scrollTo({ top: document.getElementById('root')!.scrollHeight, behavior: 'smooth' });
+      window.scrollTo({ top: document.getElementById('root')?.scrollHeight, behavior: 'smooth' });
```
jakefromphoton, line 110: This is safer way to handle missing root, particularly for tests

apps/patient/src/components/pharmacy-card-list/BrandedPharmacyCard.tsx
```diff
@@ -1,15 +1,15 @@
 import { Card, CardBody } from '@chakra-ui/react';
 
-import capsuleLogo from '../assets/capsule_logo_small_circle.png';
-import amazonPharmacyLogo from '../assets/amazon_pharmacy_logo_small_circle.png';
-import altoLogo from '../assets/alto_logo.svg';
-import costcoLogo from '../assets/costco_logo_small.png';
-import costPlusLogo from '../assets/costplus_logo_small_circle.png';
-import walmartLogo from '../assets/walmart_logo_small_circle.png';
-import novocareLogo from '../assets/novo_circle.png';
+import capsuleLogo from '../../assets/capsule_logo_small_circle.png';
```
billoran8, line 3: don't forget that there's a bunch of other files related to pharmacy card list elsewhere but they can be moved overwhen touched

apps/patient/src/Navigation.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+    renderApp({ order: testOrder });
+
+    expect(await screen.findByText('Review your prescription')).toBeInTheDocument();
+    await userEvent.click(screen.getByRole('button', { name: 'Search for a pharmacy' }));
+    expect(await screen.findByText('When do you need your order ready by?')).toBeInTheDocument();
+    await userEvent.click(screen.getByText('Urgent'));
+    await userEvent.click(screen.getByText('Next'));
+    expect(setOrderPharmacyMock).not.toHaveBeenCalled();
+    expect(await screen.findByText('Select a pharmacy')).toBeInTheDocument();
```
billoran8, line 100: nit: my pet peeve is having so many expects in a single test. i dislike that because the failure of one test prevents execution of any other test afterwards, which could be obscuring additional information about the failure. not a defect but is there any way to reframe this so that expects are in different tests and can highlight individual failures without impacting the ones that come after?

billoran8 (approved):

apps/patient/src/Navigation.test.tsx
jakefromphoton: I'm performing a page-level journey test here to document a user flow, so in my mind it's a different sort of test. I use expects to check-in at different points during the journey.

These are often seen more in Playwright/Cypress, but it seems cheaper for now to test and document the flow in this way.

billoran8: hmm good point

billoran8: my only thought is like, what if navigating to section A works, navigating to section B fails, but navigating to section C works? we won't know about section C until we fix section B

billoran8: it's a super minor quibble which is why its' not a defect

PR #1997, opened 2025-08-11
Refactor Patient Pharmacy page - extract methods

jakefromphoton:
Cleaning up the submit logic on the Pharmacy page to prepare for the navigation logic we'll need to implement for this [notion ticket](https://www.notion.so/photons/Move-requested-pick-up-time-after-a-local-pharmacy-is-selected-2398bfbbf4ec809d87d7e846bbbccca3?source=copy_link)

* Extract demo submit function
* Extract analytics function for successful submit

### DIFF
--- a/apps/patient/src/views/Pharmacy.tsx
+++ b/apps/patient/src/views/Pharmacy.tsx
@@ -39,7 +39,7 @@ import capsuleZipcodeLookup from '../data/capsuleZipcodes.json';
 import { demoPharmacies } from '../data/demoPharmacies';
 import { isGLP } from '../utils/isGLP';
 import { datadogRum } from '@datadog/browser-rum';
-import { GetPharmaciesByLocationQuery } from '../__generated__/graphql';
+import { GetPharmaciesByLocationQuery, Pharmacy as PharmacyType } from '../__generated__/graphql';
 import { getOrgMailOrderPharms } from '@client/settings';
 import { fetchOffers, getPharmacy } from './pharmacy.utils';
 import _ from 'lodash';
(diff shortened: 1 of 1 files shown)


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -671,95 +650,17 @@ export const Pharmacy = () => {
...
           );
 
+      const { type, selectedPharmacy } = getPharmacy(allPharmacies, selectedId);
+
       setTimeout(() => {
         if (result) {
+          // setSuccessfullySubmitted(true) shows a "Thank you" message in the button temporarily, during the setTimeout
+          // refactoring to an "await delay(1000)"-style timeout breaks this for some reason
           setSuccessfullySubmitted(true);
           setTimeout(async () => {
```
jakefromphoton, line 660: I really wanted to refactor the setTimeouts here but the button-text update broke, so I am leaving it here for now

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -843,6 +744,112 @@ export const Pharmacy = () => {
     }
   };
 
+  const handleDemoSubmit = () => {
```
billoran8, line 747: was gionna suggest putting these functions inot a new file but forgot how annoying it'd be to have to pass in all of the setState things too

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -631,28 +631,7 @@ export const Pharmacy = () => {
...
-            selectedPharmacy.name,
-            formatAddress(selectedPharmacy.address!)
-          );
-
-          navigate(`/status?demo=true&phone=${phone}`);
-        }, 1000);
-        setSubmitting(false);
-      }, 1000);
-
+      handleDemoSubmit();
```
billoran8, line 634: wonder if this should be awaited?

jakefromphoton: it uses setTimeouts so the await wouldnt do anything

apps/patient/src/views/Pharmacy.tsx
jakefromphoton: Yea, we can talk about the demo approach at the upcoming Patient App workshop. Just trying to make handleSubmit logic easier to follow for now.

mrrsnn (approved):

PR #2009, opened 2025-08-19
Optimistically update order context w new pharmacy and exceptions resolved on reroute

mrrsnn:
After rerouting, the status page was still showing the old pharmacy and its exceptions. we need to optimistically update the state locally while updates occur in the event system

This ticket: https://www.notion.so/photons/Status-page-needs-to-update-after-reroute-24c8bfbbf4ec80c18647ed89558dc29a?v=22b8bfbbf4ec80af8e2d000c73a7556b&source=copy_link

https://github.com/user-attachments/assets/69276560-ddb4-4202-8572-78556c41cb38

### DIFF
--- a/apps/patient/src/components/PharmacyInfo.tsx
+++ b/apps/patient/src/components/PharmacyInfo.tsx
@@ -295,7 +295,7 @@ export const PharmacyInfo = ({
   const taglineOverride = showNovocareTagline || showAmazonTagline;
 
   return (
-    <VStack align="start" w="full">
+    <VStack data-testid="pharmacy-info" align="start" w="full">
       <HStack w="full" justify="space-between">
         <VStack w="full">
           <HStack w="full">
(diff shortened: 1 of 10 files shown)


apps/patient/src/components/PharmacyInfo.tsx
```diff
@@ -295,7 +295,7 @@ export const PharmacyInfo = ({
   const taglineOverride = showNovocareTagline || showAmazonTagline;
 
   return (
-    <VStack align="start" w="full">
+    <VStack data-testid="pharmacy-info" align="start" w="full">
```
jakefromphoton, line 298: I don't love testIds if we can avoid them easily. Usually indicate accessibility issues. And the test would be a little less portable to a rewritten component/page.

That said, I'm OK with it for now but would prefer we don't make this a goto pattern on the team.

jakefromphoton (approved):

apps/patient/src/components/PharmacyInfo.tsx
mrrsnn: Yeah dude, there's been heaps of accessibility issue. This is the only tool to use at the moment

PR #2029, opened 2025-09-09
Fix login page auto-redirecting

jakefromphoton:
* Remove api call for `federated` org setting that checked token and re-directed the page
  * The `<Auth>` component is not reachable for situations where `federated` logout is relevant. Only for users authenticated with 0 organizations is it relevant.
  * This PR does not touch the Federated logout of our app's main logout button, in the `<Nav>` component
* Remove unreachable logout button. This was unreachable because the parent component `<Login>` does not render the `<Auth>` component when `isAuthenticated` is true.
* Fix e2e test due to recent Auth0 login page change 

[notion ticket](https://www.notion.so/photons/Fix-auto-redirect-from-login-pages-to-Auth0-portal-2698bfbbf4ec80e6b686c4f5a8e88044?source=copy_link)

### DIFF
--- a/apps/app/src/views/components/Auth.tsx
+++ b/apps/app/src/views/components/Auth.tsx
@@ -1,4 +1,4 @@
-import { Button, Stack } from '@chakra-ui/react';
+import { Button } from '@chakra-ui/react';
 import { usePhoton } from '@photonhealth/react';
 
 interface AuthProps {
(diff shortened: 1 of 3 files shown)


apps/app/src/views/components/Auth.tsx
```diff
@@ -49,5 +31,3 @@ export const Auth = (props: AuthProps) => {
     </Button>
   );
 };
-
-Auth.defaultProps = { returnTo: '/' };
```
jakefromphoton, line 53: defaultProps is deprecated. Moved this to props destructure w/ default.

captDaylight (approved):

PR #2034, opened 2025-09-10
Remove duplicative, erroneous handleRedirect

jakefromphoton:
Context: We have two calls to the auth0 SDK's `handleRedirect` logic that were always triggering in the `provider.tsx` file. The second call would always fail (because the `state` param was consumed by the first) and overwrite any error state detected in the first call with an `"Invalid state"` error. This `"Invalid state"` is specifically ignored in our app as a seemingly "noisy" error that we probably never understood.

More Context: A lot of this auth0 integration logic was added in February 2023 via two massive commits. My suspicion is this logic was ported from someone's previous app that had some confusing bugs, but mostly worked.

Auth0 docs clearly indicate calling the handleRedirect function twice would not be correct.

> [Relevant auth0 docs](https://auth0.com/docs/quickstart/spa/vanillajs/interactive) state:
> 
> If handling the callback was successful, the parameters **_should be removed from the URL_** so the callback handler will not be triggered the next time the page loads.

- [x] Delete the duplicative handleRedirect call in `provider.tsx` that has been causing `"Invalid state"` errors
- [x] Fix useful auth errors not showing due to `"Invalid state"` overwriting the real errors.
- [x] Change fading popup error alerts UX to a stationary alert
- [x] Delete the old, unnecessary Alert/alertStore system that isn't used and obfuscated how our alerts work 

This is a somewhat risky change, as there could be auth flows I'm unaware of. So it's worth making a list of all the flows that I am manually testing for this PR:

### Happy Path Flows
1. Standard login for existing user
2. Accepting an invite (single org user)
3. Accepting an invite (multi org user, 1 email)
4. Accepting an invite (multi org user, 1 email per org) <- effectively same as #2 single org user above
5. Switching organizations
6. Login with account that has 0 orgs -> see 0 orgs warning

### Error Flows
7. Click "Accept Invite" email link for expired invite -> Shows error popup "Invitation not found or already used"
image

9. Click "Accept Invite" email link for fresh invite -> login with wrong gmail account -> Shows error popup "The specified account is not allowed to accept the current invitation"
image

10. Login after seeing error -> see 0 orgs warning or login successfully with provisioned account
image

### Embed Note
This commit does not impact the embed's login code. Only the web app, particularly invite accepting flows but also post-invite login flows.


packages/react/src/provider.tsx
```diff
@@ -745,28 +744,6 @@ export const PhotonProvider = (opts: {
 
   /// Auth0
 
-  const handleRedirect = useCallback(
-    async (url?: string) => {
-      try {
-        await client.authentication.handleRedirect(url);
```
jakefromphoton, line 751: We already call handleRedirect in a more valid way in the useEffect above this deleted block of code.

apps/app/src/views/routes/Main.tsx
```diff
@@ -55,16 +55,11 @@ export const Main = () => {
       } else {
         addAlert({ message: error, type: 'error' });
         clearError();
+        navigate('/', { replace: true });
```
jakefromphoton: Adding this resets the queryParams in the URL for "specified account not allowed to accept" errors, allowing the next login button click to work properly without cached query params breaking the attempt.

apps/app/src/views/routes/Main.tsx
```diff
@@ -55,16 +55,11 @@ export const Main = () => {
...
       }
     }
   }, [isLoading, error]);
 
-  if (query.get('error_description') === 'invitation not found or already used') {
-    // https://www.notion.so/photons/Handle-auth0-Error-States-on-Login-3486ee6a8e5d4c60a2db091f8f0cbd78?pvs=4
-    // In this error state, instead of endless loop we can just kick the user back login with an error message
-    return <Navigate to="/login?orgs=0" replace />;
-  }
-
```
jakefromphoton, line 67: Deleting this hotfix to consolidate our error handling flows.

packages/react/src/provider.tsx
```diff
@@ -731,6 +730,7 @@ export const PhotonProvider = (opts: {
         } catch (e) {
           const message = (e as Error).message;
           dispatch({ type: 'ERROR', error: message });
+          return;
```
jakefromphoton, line 724: This prevents the next `dispatch` call in the function from clearing the error message that was just set. This fixes the alert popup not showing

apps/app/src/views/routes/Main.tsx
```diff
@@ -23,49 +22,23 @@ export const Main = () => {
...
-  if (isLoading || (previouslyAuthed && !isAuthenticated) || query.get('code')) {
+  if (isLoading || query.get('code')) {
```
jakefromphoton, line 41: I couldn't find a use for this, and it would trigger infinite spinner flows. I think we can remove this now that we're fixing the double `handleRedirect` calls. 

The `previouslyAuthed` value is no longer used after removing this, so it can be deleted entirely.

apps/app/src/views/routes/Login.tsx
```diff
@@ -48,7 +48,7 @@ export const Login = () => {
           {error && !isLoading && (
             <Alert status="error">
               <AlertIcon />
-              Access Denied
+              {error}
```
jakefromphoton, line 51: Actually show the errors happening in the auth flow from `provider.tsx` such as:

"invitation not found or already used"

"the specified account is not allowed to accept the current invitation"

apps/app/src/views/routes/Main.tsx
```diff
@@ -23,49 +22,23 @@ export const Main = () => {
...
-        clearError();
-      } else if (error.includes('invitation not found or already used')) {
-        addAlert({ message: error, type: 'error' });
-        clearError();
-        navigate('/', { replace: true });
-      } else {
-        addAlert({ message: error, type: 'error' });
-        clearError();
-      }
+      navigate('/', { replace: true });
```
jakefromphoton, line 37: This alert/error management isn't necessary now. Now we show the error in the Login page.

This logic was added in a massive commit in Feb 2023, likely ported from another app.

apps/app/src/views/routes/Main.tsx
jakefromphoton: Paul and I found this was included in the original commit of the `client` repo. My suspicion is an early engineer copied this to handle weird errors caused by the bad auth0 integration code.

captDaylight (approved):

PR #2040, opened 2025-09-17
Fix: price toggle should be enabled by default

jakefromphoton:
* Fix bug where we weren't defaulting the Coupon Price toggle to "on"
* Rewrite `Pharmacy.test.tsx` to be less brittle to refactors (such as moving state from Pharmacy.tsx up into Main.tsx)
  * major change: render entire `<Main>/<RouterProvider>` wrapper instead of merely `<Pharmacy>`. This way it doesn't matter where state lives in the app
  * minor change: use `test` block instead of many `it` blocks
  * setup 2 primary cases in `Pharmacy.test.tsx` and other cases in `shouldShowPriceToggle.test.tsx`
  
  
[notion ticket](https://www.notion.so/photons/Pricing-Toggle-Default-On-Plz-2728bfbbf4ec8033a373d7d9a5c4a117?source=copy_link)


apps/patient/src/views/Pharmacy.test.tsx
jakefromphoton, line 1: Rewrote Pharmacy.test.tsx to be a higher level page-pattern style test. It now renders the full app, rather than the Pharmacy.tsx component. This makes the test less brittle to moving state values like `enablePrice` or `showPriceToggle` around the app.

The diff won't make sense so recommend reading the file directly.

apps/patient/src/views/Main.tsx
```diff
@@ -127,7 +130,12 @@ export const Main = () => {
         pharmacy: currentPharmacy || newOrder?.pharmacy || order?.pharmacy
       });
 
-      setFlattenedFills(countFillsAndRemoveDuplicates(newOrder.fills));
+      const newFlattenedFills = countFillsAndRemoveDuplicates(newOrder.fills);
+      setFlattenedFills(newFlattenedFills);
+
+      const showPriceToggle = shouldShowPriceToggle(newFlattenedFills, newOrder);
+      setShowPriceToggle(showPriceToggle);
+      setEnablePrice(showPriceToggle);
```
jakefromphoton, line 138: this is the main fix. We need to turn on `enablePrice` if we're showing the toggle at all for an order.

apps/patient/src/Navigation.test.tsx
```diff
@@ -147,7 +147,7 @@ describe('App', () => {
       undefined,
       undefined,
       undefined,
-      false
+      true
```
jakefromphoton, line 150: Fixing assert for `patientSelectedPrice` parameter. We default to `true` because we default to showing the price toggle.

captDaylight (approved):

PR #2043, opened 2025-09-19
Search team members by user name

jakefromphoton:
* Add search by name input to Team settings for admin users to manage user list
  * Helps organizations with 1000+ members search their team. Auth0 can't return users outside of a 1000-count list, so a filter is recommended and essentially required for 1000+ user orgs
  * Helps small organizations search, too
  * Requires at least 3 characters due to a limitation in auth0 wildcard searching documented [here](https://auth0.com/docs/manage-users/user-search/user-search-query-syntax#wildcards)
* Jomi and I plan to iterate the design going forward. We'll treat this as a first iteration. 

[notion ticket](https://www.notion.so/photons/Fix-1000-user-pagination-by-adding-name-filter-2698bfbbf4ec801d937ed33540fc3a64?source=copy_link)

# Demo

https://github.com/user-attachments/assets/ad7ae4b2-ed19-4749-b020-f6df5d7ed130


billoran8 (approved):
oh cute

PR #2060, opened 2025-10-13
Add analytics to external URL click

jakefromphoton:
This will help us track analytics around who clicks on the new "Get coupon" external URL.

### DIFF
--- a/apps/patient/src/components/coupons/ExternalLinkCouponCard.tsx
+++ b/apps/patient/src/components/coupons/ExternalLinkCouponCard.tsx
@@ -3,9 +3,12 @@ import { DiscountCard } from '../../__generated__/graphql';
 import { CouponSourceLogo } from './CouponSourceLogo';
 import { formatPrice } from '../../utils/formatters';
 import { Card } from '../Card';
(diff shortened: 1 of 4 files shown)


apps/patient/src/components/coupons/ExternalLinkCouponCard.tsx
```diff
@@ -33,11 +48,8 @@ export const ExternalLinkCouponCard = (props: ExternalLinkCouponCardProps) => {
           </Text>
         ) : null}
         <Button
-          as="a"
+          onClick={handleGetCouponClick}
```
billoran8, line 51: do we only want to have an onClick event when there's an external URL? do we also have to do anything with like hoverstates or anything or is taht covered by as="a"

jakefromphoton: this component only renders when there is an externalURL, so it's unnecessary to check if it's there.
hoverState is covered by the `_hover` prop a few lines down

billoran8: oh cute

billoran8 (approved):

PR #2065, opened 2025-10-15
Block Medications in catalog for Form

captDaylight:
I ended up using treatment Ids because we don't get NDC on the lambdas model, but we do in Services. The great thing is we can extend this with `ndcs`, `categories`, `templates`

``` html
<photon-prescribe-workflow
  disable-list='[
    {
      "treatmentIds": ["med_01J6ZE5RSPDRXZRX7SEFHJSMDB"],
      "reason": "Not covered by patient insurance"
    }
  ]'
```

Passing a med id will block out templates, catalogs and meds in the general search that meet this criteria:

Screenshot 2025-10-15 at 3 36 58 PM

### DIFF
--- a/packages/elements/index.html
+++ b/packages/elements/index.html
@@ -46,7 +46,7 @@
           enable-coverage-check="true"
           enable-send-to-patient="true"
           enable-local-pickup="true"
-          disable-list='[{"ndcs": ["00169-4525-14"], "reason": "Not covered by patient insurance"}]'
+          disable-list='[{"treatmentIds": ["treatment_12345"], "reason": "Not covered by patient insurance"}]'
           id="one"
         ></photon-prescribe-workflow>
       </div>
--- a/packages/elements/index.html
+++ b/packages/elements/index.html
@@ -46,7 +46,7 @@
           enable-coverage-check="true"
           enable-send-to-patient="true"
           enable-local-pickup="true"
-          disable-list='[{"treatmentIds": ["med_01J6ZE5RSPDRXZRX7SEFHJSMDB"], "reason": "Not covered by patient insurance"}]'
+          disable-list='[{"treatmentIds": ["med_01J6ZE5RSPDRXZRX7SEFHJSMDB"], "reason": "Not covered by patient insurance"}, {"treatmentIds": ["med_01J6ZE5RTEB5FK0CES2DDQWYV2"], "reason": "Too expensive"}]'
           id="one"
         ></photon-prescribe-workflow>
       </div>
--- a/packages/elements/index.html
+++ b/packages/elements/index.html
@@ -46,7 +46,7 @@
           enable-coverage-check="true"
           enable-send-to-patient="true"
           enable-local-pickup="true"
-          disable-list='[{"treatmentIds": ["treatment_12345"], "reason": "Not covered by patient insurance"}]'
+          disable-list='[{"treatmentIds": ["med_01J6ZE5RSPDRXZRX7SEFHJSMDB"], "reason": "Not covered by patient insurance"}]'
           id="one"
         ></photon-prescribe-workflow>
       </div>
--- a/packages/elements/index.html
+++ b/packages/elements/index.html
@@ -46,6 +46,7 @@
           enable-coverage-check="true"
           enable-send-to-patient="true"
           enable-local-pickup="true"
+          disable-list='[{"ndcs": ["00169-4525-14"], "reason": "Not covered by patient insurance"}]'
           id="one"
         ></photon-prescribe-workflow>
       </div>
--- a/packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
+++ b/packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
@@ -192,22 +192,29 @@ function displayTreatment(
 function displayPrescriptionTemplate(
   t: PrescriptionTemplate,
   showFormattedMedicationName: boolean,
-  searchText: string
+  searchText: string,
+  disableList?: DisableList
 ) {
+  const treatmentId = t.treatment.id;
+  const { disabled, reason } = isTreatmentDisabled(treatmentId, disableList);
+
   if (showFormattedMedicationName) {
     const refills = t.fillsAllowed ? t.fillsAllowed - 1 : 0;
 
     return (
-      <div class="my-1">
-        <div class="text-sm whitespace-normal font-medium mb-1">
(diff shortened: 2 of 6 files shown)


packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -210,6 +212,7 @@ export const AddPrescriptionCard = (props: {
             help-text={props.store.treatment?.error}
             off-catalog-option={offCatalog()}
             search-text={searchText()}
+            disable-list={props.disableList}
```
billoran8, line 215: isn't the disable list going to be differnet per person/plan? shouldn't this be configured at the esrver level?

billoran8: maybe a brief RFC on how this will work but also not lock us through one way doors?

mrrsnn: It's going to be fed in as a prop to the element. It's a view control provided by the user.

mrrsnn: We don't reason about the disable list per person/plan, Form does. And we just take what they've figured and block. That's my understanding from the meeting yesterday at least.

billoran8: totally understand the concept of just being told specifically what to block; i just feel like this is building the feature on shaky ground - for consumers of this feature to get bugfixes/improvements they'll have to update their stuff. i get that this has to be done fast, i just feel like we're carrying on a pattern we all hate and will have to clean up later?

packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
```diff
@@ -16,12 +16,34 @@ import {
...
+
+  for (const item of disableList) {
+    if (item.treatmentIds) {
+      for (const disabledTreatmentId of item.treatmentIds) {
+        if (treatmentId === disabledTreatmentId) {
+          return { disabled: true, reason: item.reason };
+        }
+      }
+    }
+  }
```
mrrsnn: gonna nitpick the nested loop here. I can see a future way down the road where users abuse this with a long laundry list of blocking reasons, even against our own recommendation of usage. In this case the render time is `O(n*m*l)` where `n` is the number of reasons, `m` is the number of disabled treatments per reason, and `l` is the length of the medication list that we have to run this check against.

Try something like this: 
```suggestion
  const disableListsByReason = disableList.reduce((acc, cur) => {
    if (!cur.reason || !cur.treatmentIds) return acc;

    acc[cur.reason] = new Set(cur.treatmentIds);
    return acc;
  }, {} as Record<string, Set<string>>);

  for (const [reason, disabledTreatments] of Object.entries(disableListsByReason)) {
    if (disabledTreatments.has(treatmentId)) return { disabled: true, reason };
  }

```

Which would be `O(n*l)` at least. 

And if we instead move this check above the `displayTreatment` function so we only run the `reduce` and for loop once, then it's `O(n + l)`! Which, realistically will just be `O(l)` since the length of the medication list is likely longer than the length of reasons.

packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
```diff
@@ -147,34 +169,52 @@ export const boldSubstring = (inputString: string, substring: string) => {
 function displayTreatment(
   t: Treatment | TreatmentOption,
   showFormattedMedicationName: boolean,
-  searchText: string
+  searchText: string,
+  disableList?: DisableList
```
mrrsnn: I would change this to be `disabled?: boolean` and `disableReason?: string` so we can move that mapping that I proposed above.

mrrsnn: more optional than the former though

packages/elements/index.html
```diff
@@ -46,6 +46,7 @@
           enable-coverage-check="true"
           enable-send-to-patient="true"
           enable-local-pickup="true"
+          disable-list='[{"treatmentIds": ["med_01J6ZE5RSPDRXZRX7SEFHJSMDB"], "reason": "Not covered by patient insurance"}]'
```
mrrsnn: Just want to verify, did we test the scenario where this changes dynamically? Want to verify that the UI will be updated when they update this.

captDaylight: For our elements, we don't expect them to update dynamically. [Similar to the catalog id situation we spoke about before](https://photonhealth.slack.com/archives/C08HK30B86Q/p1758638248030839?thread_ts=1758635546.232059&cid=C08HK30B86Q),  our properties on elements are more a configuration for the initial state but not expected to reset on update. The would need to remount the element.

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
captDaylight: It will be different per person and plan, the thinking is that Form will configure the element's guardrails on demand based on their business logic rather than persisted in our system. 

I actually think this is an appropriate boundary of responsibility between our companies. We are agnostic to how they set this up but we aren't required to keep track of their business logic.

they are making this determination on the fly based on real time information, I don't know why we would need to involve our server

packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
captDaylight: yup that works for me

captDaylight:
Now If you pass in :
``` js
[
  {"treatmentIds": ["med_01J6ZE5RSPDRXZRX7SEFHJSMDB"], "reason": "Not covered by patient insurance"}, 
  {"treatmentIds": ["med_01J6ZE5RTEB5FK0CES2DDQWYV2"], "reason": "Too expensive"}, 
  {"treatmentIds": ["med_01J4565RSPDRXZRX7SEFHJSNIX"]}
]
```

it will create a memoized list like :

``` js
{
  "med_01J6ZE5RSPDRXZRX7SEFHJSMDB": "Not covered by patient insurance",
  "med_01J6ZE5RTEB5FK0CES2DDQWYV2": "Too expensive", 
  "med_01J4565RSPDRXZRX7SEFHJSNIX": undefined
}
``` 

much more efficient

mrrsnn (approved):

PR #2069, opened 2025-10-23
Patient Mail Order Selection Modal

mrrsnn:
per notion ticket: https://www.notion.so/photons/Add-Patient-Mail-Order-Pharmacy-Selection-2638bfbbf4ec8038870bf08ee42453f8?v=22b8bfbbf4ec80af8e2d000c73a7556b&source=copy_link

Designs: https://www.figma.com/design/WciGPaXDzLvPRUdqmk0qVq/Pharmacy-Search-MVP?node-id=10-4&p=f&t=lYbXYTH6v8ibMArC-11

https://github.com/user-attachments/assets/c2436dd1-b7cf-4139-8a7f-da988cf15039

- **Add patient api getPharmacies, rename old one to getPharmaciesByLocation**
- **Show different copy in the status page header for mail orders or integrated orders**
- **mail order select component**
- **add mail modal, attempt to refactor submit handler to support it**


apps/patient/src/views/Pharmacy.tsx
jakefromphoton, line 736: I could see why you might not want to touch this part, but you might be able to delete this line now that `selectedPharmacy` is passed in.

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -654,55 +655,69 @@ export const Pharmacy = () => {
...
+        // TODO: Remove this once we've got all pharmacies marked correctly in the db
+        // this historically was overriding pharmaicy type and presentation due to an inept datamodel
+        const override = getPharmacy(allPharmacies, selectedPharmacy.id);
+        const overridePharmacy = override.selectedPharmacy ?? selectedPharmacy;
+        const overrideType = override.selectedPharmacy
+          ? override.type
+          : selectedPharmacy.fulfillmentTypes?.[0];
+
+        await new Promise<void>((resolve) =>
+          setTimeout(() => {
```
jakefromphoton, line 683: Curious why you added the Promise wrapper.

Also, maybe if the Promise function is `async`, we could replace the `setTimeout` calls now with `await delay(1000)` calls without breaking the Submit button animation. I tried to do that before and it broke the animation, but maybe this Promise wrapper's context would allow for that.

Also OK with that being a separate refactor PR for another time.

apps/patient/src/views/Pharmacy.test.tsx
```diff
@@ -485,13 +489,118 @@ describe('Pharmacy page', () => {
...
-    const callArgs = getPharmaciesMock.mock.calls.map((call) => call[0].includePrice);
+    const callArgs = getPharmaciesByLocationMock.mock.calls.map((call) => call[0].includePrice);
     expect(callArgs).not.toContain(true);
 
     // ensure price UI is not showing in the cards
     expect(screen.queryByText('Coupon Price')).not.toBeInTheDocument();
     expect(screen.queryByText('Retail')).not.toBeInTheDocument();
   }, 10_000);
+
+  test('shows mail order select modal and submits order to mail order pharmacy', async () => {
```
jakefromphoton, line 500: Digging this test

apps/patient/src/views/Pharmacy.test.tsx
```diff
@@ -485,13 +489,118 @@ describe('Pharmacy page', () => {
...
+      patient: generatePatient(),
+      fills: [generateFill('test-treatment')],
+      address: {
+        street1: '123 Main St',
+        street2: undefined,
+        city: 'New York',
+        state: 'NY',
+        postalCode: '10001',
+        country: 'US'
+      }
```
jakefromphoton: nit: you could probably delete the address: {} here, because it looks like the default in `generateOrder`

apps/patient/src/views/Pharmacy.test.tsx
```diff
@@ -485,13 +489,118 @@ describe('Pharmacy page', () => {
...
     expect(callArgs).not.toContain(true);
 
     // ensure price UI is not showing in the cards
     expect(screen.queryByText('Coupon Price')).not.toBeInTheDocument();
     expect(screen.queryByText('Retail')).not.toBeInTheDocument();
   }, 10_000);
+
+  test('shows mail order select modal and submits order to mail order pharmacy', async () => {
+    const testOrder = generateOrder({
+      id: 'ord_testId666',
```
jakefromphoton, line 502: Bold move using that number

apps/patient/src/views/Pharmacy.test.tsx
```diff
@@ -485,13 +489,118 @@ describe('Pharmacy page', () => {
...
+        name: 'Capsule Pharmacy',
+        logo: 'https://logos.boson.health/pharmacies/capsule-logo.png',
+        fulfillmentTypes: ['MAIL_ORDER'] as FulfillmentType[]
+      },
+      {
+        id: 'phr_01GB9IPTA45KPVHG1H871Z7PZS',
+        name: 'CenterWell Pharmacy',
+        logo: 'https://logos.boson.health/pharmacies/centerwell-logo.png',
+        fulfillmentTypes: ['MAIL_ORDER'] as FulfillmentType[]
+      },
```
jakefromphoton: Could we just do 2 array items and scrub the real company names, use e.g. test names instead? Or do we need 4 array items for this test?

apps/patient/src/views/Pharmacy.test.tsx
```diff
@@ -485,13 +489,118 @@ describe('Pharmacy page', () => {
...
+
+    const modalOpenButton = await screen.findByText('See all mail orders');
+    expect(modalOpenButton).toBeInTheDocument();
+
+    await userEvent.click(modalOpenButton);
+
+    // find the modal Header
+    await waitFor(() => screen.findByText(/Mail Order Pharmacies/i));
+
+    const mailOrderOption = await screen.findByText(mailOrderPharmacyData.at(-1)?.name ?? '');
```
jakefromphoton: More nit but I would prefer

`const mailOrderOption = await screen.findByText(mailOrderPharmacyData.at(1)!.name);` 

or even 

`const mailOrderOption = await screen.findByText('test-mail-order-2');`

jakefromphoton (approved):

apps/patient/src/views/Pharmacy.tsx
mrrsnn: I would LOVE to. But unfortunately some of the hardcoded IDs in that `getPharmacy` check won't be changed with the correct fulfillment types until after this goes out.

apps/patient/src/views/Pharmacy.tsx
mrrsnn: I wrapped this in a promise because my mail order select modal sets a loading state and waits for the `onConfirm` function to resolve. This `handleSubmit` function is being used in the `onConfirm` prop, and it resolves way too soon because of the settimeout processing, then there's 2+ seconds of waiting with no loading state (and no guard on the submit button) before it's finished. 

I think the `await delay()` instead of setTimeouts would be much better. And in that scenario, we don't even need the wrapping `new Promise`. But I'm curious why that broke for you. I might give it a try right now.

mrrsnn: I would hate for a bug in this to require me to rollback the feature though

apps/patient/src/views/Pharmacy.test.tsx
mrrsnn: 😈

apps/patient/src/views/Pharmacy.test.tsx
mrrsnn: Don't need 4 at all, I just copy/pasted the response from my network tab 🤷

mrrsnn: Can totally remove the real company names, but I'm curious why?

apps/patient/src/views/Pharmacy.tsx
mrrsnn: Hate this function, `getPharmacy` doesn't really do what you think it would do by the name. It more so returns an override of name and type based on environment variable pharmacy ids.

PR #2075, opened 2025-10-27
Fix the optimistic update before status page to include the proper fulfillment type

mrrsnn:
https://github.com/user-attachments/assets/074a165d-92ce-489c-a94f-527a0f5271f6

tadaaa

### DIFF
--- a/apps/patient/src/views/Pharmacy.tsx
+++ b/apps/patient/src/views/Pharmacy.tsx
@@ -40,7 +40,11 @@ import capsuleZipcodeLookup from '../data/capsuleZipcodes.json';
 import { demoPharmacies } from '../data/demoPharmacies';
 import { isGLP } from '../utils/isGLP';
 import { datadogRum } from '@datadog/browser-rum';
-import { GetPharmaciesByLocationQuery, Pharmacy as PharmacyType } from '../__generated__/graphql';
+import {
+  FulfillmentType,
+  GetPharmaciesByLocationQuery,
+  Pharmacy as PharmacyType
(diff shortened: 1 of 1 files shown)


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -692,6 +696,11 @@ export const Pharmacy = () => {
                 const updatedOrder = await fetchOrder(overridePharmacy);
 
                 if (updatedOrder) {
+                  updatedOrder.fulfillment = {
```
billoran8, line 699: do we want to do this? or should we instead wait for the event to actually create the apprope entities?

captDaylight (approved):

apps/patient/src/views/Pharmacy.tsx
captDaylight: I think this works, especially for a snappy response for the user. The likelihood of the update failing after a successful request feels slim, but we technically could add a poll (after we display success) just to check if the value gets updated in the off chance it fails... but for now imo get it out

billoran8: fine with doing this to get it out now but we need to remember that a lot of things aren't truly "done" until an apprope entrypointi

billoran8 (approved):

apps/patient/src/views/Pharmacy.tsx
mrrsnn: yeah that could take up to 15 seconds because of our kinesis handlers. Can't make the patient wait that long

PR #2097, opened 2025-11-05
Self signup Validates input for name, email, npi, address

jakefromphoton:
* Name, email, NPI, address inputs are sent
  * they aren't saved yet but next ticket will remove the hardcoded NPI/address in Auth0 Action script: https://github.com/Photon-Health/terraform/pull/2944
* form validation using Formik

[LOOM](https://www.loom.com/share/43b09a7201eb4f349307207f135b979d)


captDaylight (approved):

captDaylight:
Creating an element implies a few things, so a few questions:

Will ToS be a general purpose thing that other customers will use? This will make it available to all customers.

Is this early over optimizing? We could easily rewrite this in the future.

Shouldn't ToS always exist inside of our site?

PR #2110, opened 2025-11-11
Prescribe Flow Mail Order Pharmacy Search

mrrsnn:
## Forward
In order to mark these mail order pharmacies as `MAIL_ORDER` instead of `PICK_UP`, we need a way to expose them on the clinical app and elements. Implementing a combobox input that hits our clinical-api `pharmacies()` resolver with the non-integrated mail order params

### Vid

https://github.com/user-attachments/assets/23eb9fb5-6f4d-4f65-90ef-1218738b3b3a

## Dev Notes
I had to refactor the `ComboBox` component to run `props.setSelected` from the context, in order to update the `props.value` provided directly. Then separately listen for `props.value` to change in an effect hook and update the internal state. 

The idea: make it feel and work more like a regular form component, rather than having an isolated state value inside that's uncontrollable from outside of the combobox context. 

This, combined with the update to `RadioGroupCards` which listens for `props.value` to change and updates its internal state as well, made it possible to control the same signal value between the combobox and radio cards on the mail order tab. 

### Location Select and Pickup Pharmacy Select

https://github.com/user-attachments/assets/6e82997b-00c4-41ae-8788-f524c7ce4370

### Dosage Calculator Unit Select

https://github.com/user-attachments/assets/13f519ef-fb07-45ab-9ebf-b47011c89f13


jakefromphoton (approved):

jakefromphoton:
Digging the cleanup of input components. New mail order picker looks great. Good demo videos too

PR #2171, opened 2025-12-10
Switch npm to trusted publishing

captDaylight:
https://github.blog/changelog/2025-07-31-npm-trusted-publishing-with-oidc-is-generally-available/
https://docs.npmjs.com/trusted-publishers


myang5 (approved):
ty!

PR #2190, opened 2025-12-18
use nx affected instead of hard-coded projects list in dev:app command

myang5:


apps/app/project.json
```diff
@@ -19,7 +19,7 @@
         "cwd": "apps/app",
         "commands": [
           "env-cmd -f .env.boson react-app-rewired start",
-          "nx watch --projects=app --includeDependentProjects -- nx run-many -t build --projects=elements,react,sdk"
+          "nx watch --projects=app --includeDependentProjects -- nx affected -t build --uncommitted=true --exclude=app"
```
mrrsnn, line 22: can you just briefly explain the nx flags here and why

myang5: `nx watch --projects=app --includeDependentProjects -- <command that runs on file change>`
- `--projects=app --includeDependentProjects` watch for changes in clinical app project including in dependencies projects (elements, sdk) (couldn't find a way to only watch for changes in dependencies)

`nx affected -t build --uncommitted=true --exclude=app"`
- `nx affected` can run commands on projects filtered by specific files instead of project names (which is what `nx run` and `nx run-many` do)
- `--uncommitted=true` run commands on projects containing uncommitted files (improvement over always rebuilding all dependencies)
- `--exclude=app` rebuilding the clinical app on change is already handled by webpack, no need to do extra work

mrrsnn: love it, thank you

mrrsnn (approved):

PR #2212, opened 2026-01-09
Address optional prescribe flow

captDaylight:
Adds new prop `optional-patient-address`, which defaults to false, that either hides or shows the blocking requirement of a patient's address.

- prescribe workflow
- new / update patient form

Screenshot 2026-01-16 at 2 57 08 PM
^ If you try and select local pickup or mail order without a patient address

https://linear.app/photon-health/issue/PHO-15/as-a-provider-i-am-not-required-to-input-patient-address-unless-i


packages/elements/src/photon-patient-form/photon-patient-form-component.tsx
```diff
@@ -303,13 +320,18 @@ const PatientForm = (props: { patientId: string }) => {
...
+                    <span class="text-gray-500 text-sm font-normal"> (optional)</span>
+                  </Show>
+                </p>
                 <photon-text-input
                   debounce-time="0"
                   invalid={store['address_street1']?.error}
                   help-text={store['address_street1']?.error}
                   label="Street 1"
-                  required="true"
+                  required={!props.optionalPatientAddress || hasAnyAddressField()}
```
myang5, line 330: Ah `hasAnyAddressField()` to account for existing patients before this feature is deployed

myang5 (approved):
Nice!

myang5 (approved):

packages/elements/src/photon-patient-form/photon-patient-form-component.tsx
captDaylight: it's to account for a partially filled out address field upon submit, in that case we should error... either fill it all out or keep it all empty

PR #2236, opened 2026-01-21
fix rudderstack error on sso page

jakefromphoton:
Wasn't this annoying? Using lazy load on `track` so SSO page doesn't attempt to load Rudderstack.

image


myang5 (approved):
dont have much context but lgtm

PR #2248, opened 2026-01-26
fixin form-data dependency

billoran8:


billoran8:
links to PHO-93

package.json
```diff
@@ -182,6 +182,8 @@
   "overrides": {
     "@chakra-ui/react": {
       "@zag-js/focus-visible": "^1.18.0"
-    }
+    },
+    "graphql-request": { "form-data": "3.0.4" },
```
billoran8, line 186: these two things are using impacted versions 3.0.3 and it's pined in place by the lock file. deleting and regenerating the entire lockfile causes build problems os i'm just going to include the upgrades here, we can get rid of these lines later.

jakefromphoton (approved):

PR #2250, opened 2026-01-26
Add "Sponsored" text to OfferInfo

myang5:
Part of PHO-137

Screenshot 2026-01-26 at 2 49 11 PM

### DIFF
--- a/apps/patient/src/components/offers/OfferCard.tsx
+++ b/apps/patient/src/components/offers/OfferCard.tsx
@@ -1,11 +1,10 @@
 import { Card, CardBody } from '@chakra-ui/react';
 
-import { Offer, PHARMACY_BRANDING } from '../pharmacy-card-list';
+import { Offer } from '../pharmacy-card-list';
 import { OfferInfo } from './OfferInfo';
 
 interface Props {
   offer: Offer;
-  pharmacyId: string;
   isPharmacyFulfillingCurrentOrder: boolean;
   selected: boolean;
   isPreferred: boolean;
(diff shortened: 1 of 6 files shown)


apps/patient/src/components/offers/OfferInfo.tsx
```diff
@@ -110,16 +113,14 @@ export const OfferInfo = ({ pharmacy, offer, isCurrentPharmacy, isPreferred }: O
...
 
-      <VStack direction={'column-reverse'} w="full" alignItems={'start'}>
-        <Text fontSize="sm" color="gray.500">
+      <VStack w="full" alignItems="start">
+        <Text fontSize="sm" fontWeight="semibold">
           {offer.deliveryEstimate}
         </Text>
+        <Text fontSize="sm" color="gray.500">
+          Sponsored
+        </Text>
```
mrrsnn: As it stands, I suppose Amazon is one of the only offers that we currently show. However, I'm not sure if we want to have the `Sponsored` tag for every and any offer going forward in the future. Deferring to @billoran8 on this one. Might need a way to differentiate regular offers from sponsored ones? Or can we safely assume that offers are always considered sponsored?

myang5: Confirmed with Entela and Anna that we should only show this for Amazon

apps/patient/src/components/offers/OfferCard.tsx
```diff
@@ -1,32 +1,22 @@
...
   isPreferred
 }: Props) => {
-  const brand = PHARMACY_BRANDING[pharmacyId];
-
-  const pharmacy = {
-    id: pharmacyId,
-    name: offer.pharmacy.name,
-    logo: brand?.logo ? brand.logo : ''
-  };
-
```
myang5, line 29: The `offer` object already contains all this info in this format including `logo`, not sure why we were re-deriving this object

apps/patient/src/components/offers/OfferInfo.tsx
```diff
@@ -34,14 +32,16 @@ const CurrentPharmacyTag = () => {
 };
 
 interface OfferInfoProps {
-  pharmacy?: EnrichedPharmacy;
+  pharmacy?: Pick<Offer['pharmacy'], 'id' | 'name' | 'logo'>;
```
myang5, line 35: `EnrichedPharmacy` type had a bunch of extra optional fields that were not representative of the data actually being passed down to the component

mrrsnn (approved):

PR #2266, opened 2026-01-27
Refactor how we dispatch photon-draft-prescription-created event to 

myang5:
The idea behind PrescribeEventDispatchProvider is to be able pass the HTML `dispatchEvent` method without tying it to the HTML of a component further down the tree.

PrescribeEventDispatchProvider itself owns rendering a `<div>` element and passing a ref to it, which allows us to define methods that call `dispatchEvent` and pass them to any of the context's children. This allows us to fix a bug where:
- `photon-draft-prescription-created` event is only dispatched in the child AddPrescriptionCard component
- the event is not dispatched for draft prescriptions created by passing `templateIds` or `prescriptionIds` to `<photon-prescribe-workflow>`

Was able to test that the element now fires two HTML events when two template IDs are passed to `<photon-prescribe-workflow>`:
Screenshot 2026-01-27 at 2 30 43 PM

If this pattern is valid I'll migrate more `dispatchEvent` calls in `<photon-prescribe-workflow>` into the provider to improve code organization.

Part of PHO-156


packages/components/src/systems/PatientMedHistory/PatientMedHistoryTable.tsx
```diff
@@ -46,12 +46,8 @@ export default function PatientMedHistoryTable(props: PatientMedHistoryTableProp
     if (isCreatingPrescriptionId() === undefined) {
       setIsCreatingPrescriptionId(prescription.id);
 
-      if (!prescribeContext) {
-        throw new Error('Refill requires <PrescribeProvider>');
-      }
-
```
myang5, line 52: I'm not sure why we were using this `usePrescribeOptional` when we check for the existence of `prescribeContext` anyway

packages/components/src/systems/PrescribeProvider/PrescribeProvider.tsx
```diff
@@ -445,6 +441,7 @@ export const PrescribeProvider = (props: PrescribeProviderProps) => {
       const created = res.data.createPrescription as Prescription;
       createdPrescription = created;
       setDraftPrescriptions((prev) => [...prev, created]);
+      dispatchDraftPrescriptionCreated(created);
```
myang5, line 444: Now possible to call `dispatchDraftPrescriptionCreated` in PrescribeProvider after refactor

packages/sdk/src/types.ts
```diff
@@ -612,30 +612,21 @@ export type PrescriptionFilter = {
 };
 
 export type PrescriptionInput = {
-  /** Number of days a single fill lasts */
```
myang5, line 615: Updating type to reflect lambdas schema

packages/elements/src/photon-multirx-form/components/PrescribeWorkflow.tsx
myang5, line 1: Wanted to make this into a pure solidjs component and have photon-prescribe-workflow-component.tsx handle any web components concerns. Was able to move all the shoelace code into photon-prescribe-workflow-component.tsx and rename this file

packages/elements/src/photon-multirx-form/components/PrescribeWorkflow.tsx
```diff
@@ -121,6 +108,7 @@ export type PrescribeProps = {
   allowOffCatalogSearch?: boolean;
   disableList?: DisableList;
   groupId?: string;
+  initialShowForm: boolean;
 };
```
myang5, line 112: There's already another signal called `prescriptionIds` so it was confusing to also pass a `prescriptionIds` props

packages/components/src/systems/PatientMedHistory/PatientMedHistoryTable.tsx
jakefromphoton: I believe this was here to prevent errors for customers who use the `<photon-patient-med-history>` component in isolate. The provider won't exist in that context. So we'll need to keep this, maybe add a comment to clarify why it's here

packages/components/src/systems/PrescribeProvider/PrescribeProvider.tsx
jakefromphoton: Awesome

packages/components/src/systems/PatientMedHistory/PatientMedHistoryTable.tsx
```diff
@@ -22,6 +22,8 @@ export type PatientMedHistoryTableProps = {
 };
 
 export default function PatientMedHistoryTable(props: PatientMedHistoryTableProps) {
+  // Component can be used as a standalone element
+  // so PrescribeProvider is not guaranteed to be rendered
```
jakefromphoton, line 26: :bufo-happy:

jakefromphoton:
Looks great. Will need to bump package version for Elements NPM release

PR #2270, opened 2026-01-27
fixing anoher form-data vuln

billoran8:


package.json
```diff
@@ -201,6 +201,7 @@
     "jsdom": { "form-data": "4.0.4" },
     "nx": { "form-data": "4.0.4" },
     "react-scripts": { "form-data": "3.0.4" },
-    "storybook": { "form-data": "4.0.4" }
+    "storybook": { "form-data": "4.0.4" },
+    "cypress": { "form-data": "2.5.4" }
```
mrrsnn: do we even use cypress anymore?

billoran8: probably not! i just want to resolve the vunlerability then we can clean up these deps. a lot of these desperately need to be updated.

mrrsnn: let's just remove cypress then? Why pin one if its dependencies if we're not using it and don't plan on using it. Especially since it's a dev dep

billoran8: hmm good point

jakefromphoton (approved):

PR #2271, opened 2026-01-27
move more HTML events into PrescribeEventDispatchProvider

myang5:
Since moving one event in this PR https://github.com/Photon-Health/client/pull/2266 worked, we can move the rest of the events dispatched in `<photon-prescribe-workflow-component>` into the provider

Part of PHO-156


packages/components/src/systems/SignatureAttestation/SignatureAttestationModal.tsx
```diff
@@ -201,7 +185,7 @@ export const SignatureAttestationModal = (props: SignatureAttestationModalProps)
   return (
     <>
       {/* This span is needed so we can easily dispatch events */}
-      <span ref={ref} hidden />
+      <span hidden />
```
jakefromphoton: nit: could delete the comment and span

jakefromphoton (approved):

PR #2290, opened 2026-02-05
fix form-data vulnerability again

myang5:


.prettierignore
```diff
@@ -30,7 +32,6 @@ __generated__
 *.webp
 
 # Ignore specific file types
-*.json
 *.txt
```
myang5, line 35: Mainly wanted to remove `*.json` from being ignored, rest of diff is consolidating specific paths into more flexible glob patterns

package.json
```diff
@@ -129,7 +126,7 @@
     "git-revision-webpack-plugin": "^5.0.0",
     "graphiql": "^2.4.1",
     "graphql": ">=16.0.0 <16.7.0",
-    "graphql-request": "^5.2.0",
+    "graphql-request": "^6.1.0",
```
myang5, line 129: After generating a new package-lock.json, this older version was causing type errors in codegen files (codegen uses 6.2.0 of graphql-request)

myang5:
Too much difficulty resolving errors, don't think generating a new package-lock.json was the right approach. Closing in favor of https://github.com/Photon-Health/client/pull/2311

PR #2291, opened 2026-02-05
Signup page improvements

jakefromphoton:
* Copy changes to make form less intimidating
* Remove fax
  * This way there's less optional data fields that could push providers away from finishing Signup

Part of PHO-181


myang5 (approved):
hopefully helps 🤞

PR #2294, opened 2026-02-05
fix types for updateProviderProfile schema changes

myang5:
Part of PHO-119


jakefromphoton:
Is this mainly two things, a (1) type rename and (2) building of gql types that never got built/refreshed properly?

jakefromphoton (approved):

PR #2305, opened 2026-02-06
Don't override org in ddrum context

jakefromphoton:
I noticed it's possible for a multi-org/power user (e.g. mainly us devs) to overwrite the `org` param in DDRum context when switching from non-selfsignup user to a selfsignup user session. I don't want to ever accidentally overwrite this.


apps/app/src/instrumentation/setInstrumentationUserContext.tsx
jakefromphoton, line 12: this is what i dont want to overwrite

billoran8 (approved):

PR #2317, opened 2026-02-11
Display Patients' Preferred Pharmacies

captDaylight:
Screenshot 2026-02-11 at 2 00 33 PM


apps/patient/src/graphql/queries.ts
```diff
@@ -190,6 +189,12 @@ export const GET_ORDER = gql`
           postalCode
           country
         }
+        preferredPharmacies {
```
captDaylight, line 192: this was just added to the patient schema

mrrsnn (approved):

PR #2329, opened 2026-02-12
Signup address autocompletes via google places API

jakefromphoton:
- Uses Google API for address prefill on Signup
- uses chakra for dropdown list

 ⚠️ hide whitespace when reviewing the diff

Part of PHO-205


apps/app/src/views/routes/SelfSignup/components/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,144 @@
...
+  const geocoder = useRef<google.maps.Geocoder | null>(null);
+  const onSelectRef = useRef(onSelect);
+  onSelectRef.current = onSelect;
+
+  const getService = () => {
+    if (!autocompleteService.current) {
+      autocompleteService.current = new google.maps.places.AutocompleteService();
+    }
+    return autocompleteService.current;
+  };
```
jakefromphoton, line 32: This google maps integration uses the script downloading in our app `index.html` file

mrrsnn: is there not an npm package?

jakefromphoton: looks like there is: https://www.npmjs.com/package/@googlemaps/js-api-loader

would rather migrate to that in a separate PR, though. It would involve refactoring our ~5 uses of of the maps API

jakefromphoton: will fix in ticket PHO-221

apps/app/src/views/routes/SelfSignup/components/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,144 @@
...
+    if (!geocoder.current) {
+      geocoder.current = new google.maps.Geocoder();
+    }
+    return geocoder.current;
+  };
+
+  const fetchSuggestions = useDebouncedCallback(async (input: string) => {
+    if (input.length < 3) {
+      setSuggestions([]);
+      return;
```
myang5: Add comment with reason for condition here? I can guess what this optimization is for (no need to send prediction request for such a short input) but I always wonder about hardcoded values

apps/app/src/views/routes/SelfSignup/components/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,144 @@
...
+  const closeSuggestions = useCallback(() => {
+    fetchSuggestions.cancel();
+    setSuggestions([]);
+  }, [fetchSuggestions]);
+
+  const openSuggestions = useCallback(() => {
+    if (cachedSuggestions.current.length > 0) {
+      setSuggestions(cachedSuggestions.current);
+    }
+  }, []);
```
myang5: Interesting use of refs!

mrrsnn (approved):
Actually really surprised with how much I like that hook it created.

PR #2342, opened 2026-02-17
fixin issue with tracking events; adding logs for nonprod environemnt events

billoran8:


apps/patient/src/configs/analytics.ts
```diff
@@ -240,6 +240,17 @@ export class PatientAnalytics {
       ...properties
     };
 
+    const isNonProductionEnvironment =
```
billoran8, line 243: sonia wants to understand events better and suggested logging out when specific events were happening so she could better undersand them by looking at the console. don't hate the idea! could be cute for everyone imo

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -753,23 +763,13 @@ export const Pharmacy = () => {
               patientSelectedPrice
             );
 
-        // TODO: Remove this once we've got all pharmacies marked correctly in the db
-        // this historically was overriding pharmaicy type and presentation due to an inept datamodel
```
billoran8, line 757: if you scroll up a bit this event was only being triggered on reroutes and for mailOrders.  i moved it up so it was triggered appropriately.

myang5 (approved):

mrrsnn (approved):

PR #2345, opened 2026-02-17
Improve StP Copy

captDaylight:
https://linear.app/photon-health/issue/PHO-236/better-explain-preferred-pharmacy-in-provider-ui

Screenshot 2026-02-17 at 3 36 49 PM

### DIFF
--- a/packages/components/src/systems/PharmacySelect/SendToPatient.tsx
+++ b/packages/components/src/systems/PharmacySelect/SendToPatient.tsx
@@ -37,7 +37,7 @@ const stpStates: {
   preferredPharmacy: {
     badgeColor: 'blue',
     badgeText: 'Preferred Pharmacy',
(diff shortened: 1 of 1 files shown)


packages/components/src/systems/PharmacySelect/SendToPatient.tsx
```diff
@@ -37,7 +37,7 @@ const stpStates: {
   preferredPharmacy: {
     badgeColor: 'blue',
     badgeText: 'Preferred Pharmacy',
-    text: 'This patient selected this as their preferred pharmacy.'
+    text: "The patient can update their preferred pharmacy selection for 15 minutes. Otherwise, we'll send it to their preferred pharmacy on file."
```
billoran8, line 40: wondering if it might be smart to extract the 15 mins into a global var somewhere for when/if we do wind up cahnging it

billoran8 (approved):

PR #2359, opened 2026-02-19
Clinical app analytics for Patient Form and Signature Attestation

jakefromphoton:
* add clinical app analytics for dropoff analysis (Photon web app only)
  *  patient form interactions for form interactions
  * signature attestation viewed / agreed to / canceled
* the forms were written in Elements and render within a shadow DOM on the Clinical App, so we have to manually emit analytics events from the Element and the Clinical App listens to those and makes the http call to the backend with the data 

Part of PHO-227


apps/app/src/views/routes/NewPatient/PatientForm.tsx
```diff
@@ -53,6 +54,14 @@ export const PatientForm = () => {
       ref.current.addEventListener('photon-patient-closed', () => {
         navigate(`/patients`);
       });
+      ref.current.addEventListener(
+        'photon-analytics-event',
+        (e: CustomEvent<FormAnalyticsEventDetail>) => {
+          const { event, formName, properties } = e.detail;
+          // remove "test-" after confirming this looks right on backend
+          track(`test-${event}`, { formName, ...properties });
```
jakefromphoton: Going to remove the `test-` before merging this PR. Will confirm data looks good on backend

jakefromphoton: todo: investigate why this track is called twice

apps/app/src/views/routes/NewPatient/PatientForm.tsx
```diff
@@ -49,11 +57,25 @@ export const PatientForm = () => {
...
+      'photon-analytics-event',
+      (e: CustomEvent<FormAnalyticsEventDetail>) => {
+        const { event, formName, properties } = e.detail;
+        // remove "test-" after confirming this looks right on backend
+        track(`test-${event}`, { formName, ...properties });
+      },
+      listenerOptions
+    );
+
+    return () => abortController.abort();
```
jakefromphoton, line 81: Aborting the signal fixes `track` being called twice -- the first track instance will be deallocated and the eventListener aborted.

apps/app/src/views/routes/NewPatient/PatientForm.tsx
```diff
@@ -36,11 +38,18 @@ export const PatientForm = () => {
...
 
   useEffect(() => {
-    if (ref.current) {
-      ref.current.open = true;
-      track('patient_form_opened');
+    if (!ref.current) return;
 
-      ref.current.addEventListener('photon-patient-created', (e: any) => {
+    const abortController = new AbortController();
+    const { signal: abortControllerSignal } = abortController;
```
jakefromphoton, line 44: Abort controller signals let the original `eventListeners` be removed nicely when the effect re-runs (particularly on `track` changing as analytics context is loaded in)

(Without the abort signal, we were getting double events invoked because of stale listeners)

apps/app/src/hooks/useProviderAnalytics.tsx
```diff
@@ -48,11 +48,18 @@ export interface ProviderContextData {
   customerName?: string;
 }
 
+// these values map to table names, so please use clinicalapp as a prefix
+type ClinicalAppTrackEventName =
+  | 'test_clinicalapp_patient_form_track_events'
+  | 'test_clinicalapp_signature_attestation_form_track_events'
+  | 'prescription_form_opened'
+  | 'prescription_form_created';
+
```
jakefromphoton, line 57: these will map to table names. next PR, when I work on Rx form analytics, I will replace these 2 prescription_form events. Making these type-safe makes it so devs have to be intentional about picking a name, and will look at these other names here and ideally be consistent with the formatting.

apps/app/src/views/routes/PrescriptionForm.tsx
```diff
@@ -64,11 +66,48 @@ export const PrescriptionForm = () => {
...
 
+  const openTracked = useRef(false);
   useEffect(() => {
-    if (ref.current) {
+    if (isReady && !openTracked.current) {
+      openTracked.current = true;
       track('prescription_form_opened', { patientId: patientId || undefined });
+    }
+  }, [isReady, track, patientId]);
+
```
jakefromphoton: move this to a separate event because it was triggering twice

apps/app/src/views/routes/PrescriptionForm.tsx
```diff
@@ -87,56 +126,49 @@ export const PrescriptionForm = () => {
...
+    );
+    return () => abortController.abort();
   }, [navigate, track, patientId, user, onClose]);
 
-  useEffect(() => {
-    if (patientId && ref.current) {
-      ref.current.patientId = patientId;
-    }
-  }, [patientId]);
-
```
jakefromphoton, line 139: moved this up into the above useEffect so that the analytics event can be guaranteed to be added before the patientId is set on the underlying component

apps/app/src/hooks/useProviderAnalytics.tsx
mrrsnn: an object `as const` type of enum with a derived type could be handy. That way users can call `track(ClinicalEvents.PrescriptionFormOpened)`

CLAUDE.md
```diff
@@ -101,6 +101,8 @@ Pure **Solid.js + TypeScript** component library (private, not published to npm)
 
 When building new UI for Photon Elements, add Solid.js components here and compose them in `packages/elements`.
 
+**Never import from `@photonhealth/components` (or any path into `packages/components/src`) inside `apps/app` or `apps/patient`.** Both apps use React's JSX transform (`"jsx": "react-jsx"`), while `packages/components` uses Solid's (`"jsxImportSource": "solid-js"`). Importing across that boundary causes TypeScript to type-check Solid JSX as React JSX, producing spurious errors (`class` vs `className`, `Show`/`Match` not valid JSX components, etc.). Any types that need to cross this boundary belong in `packages/sdk` — see **Shared types** in Code Conventions.
+
```
jakefromphoton, line 105: Put this here because some code Claude generated tried to import from components (Solidjs) into the clinical app (reactjs), which is a no-no and causes weird typescript errors

apps/app/src/views/routes/PrescriptionForm.tsx
```diff
@@ -37,7 +39,7 @@ const orgSettingsQuery = graphql(/* GraphQL */ `
 export const PrescriptionForm = () => {
   const ref: MutableRefObject<any> = useRef();
   const { user, clinicalClient } = usePhoton();
-  const { track } = useProviderAnalytics();
+  const { track, isReady } = useProviderAnalytics();
```
mrrsnn: maybe don't destructure here, otherwise `isReady` is kinda ambiguous later in the file. 
```suggestion
  const providerAnalytics = useProviderAnalytics();
```

then call `providerAnalytics.track()` and use `providerAnalytics.isReady`

packages/sdk/src/clinicalAnalyticsTypes.ts
```diff
@@ -0,0 +1,27 @@
...
+  | 'signature_attestation_shown'
+  | 'signature_attestation_agreed'
+  | 'signature_attestation_canceled'
+  | 'field_interaction';
+
+export interface FieldCompletionSnapshot {
+  [fieldName: string]: { completed: boolean };
+}
+
+export interface PhotonEmbedAnalyticsEventDetail {
```
jakefromphoton: I put Embed in this interface name to help document that all of these events are emitted from the Embed element. The clinical app wrappers will be listening for these in order to submit them to our backend

apps/app/src/views/routes/NewPatient/PatientForm.tsx
```diff
@@ -36,24 +38,47 @@ export const PatientForm = () => {
...
         }
-      });
-      ref.current.addEventListener('photon-patient-closed', () => {
-        navigate(`/patients`);
-      });
-    }
+      },
+      listenerOptions
+    );
+    ref.current.addEventListener(
```
mrrsnn, line 75: so weird to me that these weren't originally just event handler functions as properties. But I see that it was already this way.

packages/components/src/systems/PrescribeEventDispatchProvider.tsx
```diff
@@ -165,6 +172,14 @@ export const PrescribeEventDispatchProvider = (props: DraftPrescriptionProviderP
     ref?.dispatchEvent(event);
   };
 
+  const dispatchAnalytics = (
+    trackEventType: ClinicalAppTrackEventType,
+    fields?: FieldCompletionSnapshot,
+    properties?: Record<string, unknown>
+  ) => {
+    dispatchAnalyticsTrackEvent({ trackEventType, fields, properties }, ref);
+  };
```
mrrsnn: will we ever want to `dispatchAnalytics` without the context of a form? `fields` kinda binds this to form usage. any reason they can't just be properties? I might be missing something crucial. maybe this is just a naming thing.

packages/components/src/systems/SignatureAttestation/SignatureAttestationModal.tsx
```diff
@@ -154,6 +157,15 @@ export const SignatureAttestationModal = (props: SignatureAttestationModalProps)
     }
   });
 
+  createEffect(() => {
+    const curr = status();
+    if (curr.status === 'NEEDS ATTESTATION') {
+      dispatchAnalytics('signature_attestation_shown', undefined, {
+        attestationVersion: curr.version
+      });
+    }
```
mrrsnn: Ahh yep good example. kinda weird to have to pass `undefined` as a second unnamed arg.

apps/app/src/views/routes/UpdatePatientForm.tsx
```diff
@@ -36,21 +40,48 @@ export const UpdatePatientForm = () => {
...
+      'photon-analytics-track-event',
+      (e: CustomEvent<PhotonEmbedAnalyticsEventDetail>) => {
+        track(
+          'clinicalapp_patient_form_track_events',
+          buildPatientFormInteractionPayload(e.detail)
+        );
+      },
+      listenerOptions
+    );
+
```
mrrsnn, line 59: where does `trackEventType` fit into this after we've bubbled the event up? Is it used at all in the actual analytics call?

mrrsnn (commented):
I'm kinda struggling to follow along at some parts. Might need to just chat async to make sure i'm understanding the different event calls.

apps/app/src/views/routes/NewPatient/PatientForm.tsx
jakefromphoton: it's because of WebComponents not allowing functions as properties. JS events are the only way to communicate up to the parent

packages/components/src/systems/PrescribeEventDispatchProvider.tsx
jakefromphoton: Eh it's a good callout IMO. There are so many naming decisions on this PR and I didn't think too much about this one. I'll look into it

apps/app/src/views/routes/UpdatePatientForm.tsx
jakefromphoton: good Q here is the network call request body where you can see it in the properties:

```json
{
  "properties": {
    "environment": "boson",
    "providerId": "usr_01KJ65MPB4HPBS7MXXPBHH9X7W",
    "providerEmail": "test@photon.health",
    "providerName": "Jake Test",
    "providerNameFirst": "Jake",
    "providerNameLast": "Test",
    "orgId": "org_testId123",
    "orgName": "Jake Org",
    "customerId": "cst_10101010",
    "customerName": "test customer name",
    "trackEventType": "patient_created",
    "patientId": "pat_01KJ8P4R2H4CQFC0CZA67J4YCZ",
    "snap_first_name": true,
    "snap_last_name": true,
    "snap_date_of_birth": true,
    "snap_phone": true,
    "snap_sex": true,
    "snap_gender": true,
    "snap_email": true,
    "snap_address_street1": true,
    "snap_address_street2": false,
    "snap_address_city": true,
    "snap_address_state": true,
    "snap_address_zip": true,
    "snap_preferred_pharmacy": false
  },
  "event": "test_clinicalapp_patient_form_track_events",
  "type": "track",
  "channel": "web",
  "context": {
    "sessionId": 1771965643620,
    "page": {
      "url": "https://app-clinical_analytics.boson.health/prescriptions/new?patientId=pat_01KJ8P4R2H4CQFC0CZA67J4YCZ",
    },
    "timezone": "GMT-0500"
  },
  "anonymousId": "dc6aac00-88bb-4cf7-bb29-fe03c2622e94",
  "sentAt": "2026-02-24T20:41:09.641Z"
}
```

jakefromphoton: the `snap_` fields are attached to non-fieldInteraction trackEventTypes to log what fields were filled out when that trackEvent occurred. So we can see patients created with and without address, for example

packages/components/src/systems/PrescribeEventDispatchProvider.tsx
jakefromphoton: I think this comment is pushing me over the edge to setup unique Types for each event type. And put the fields into the `properties`. So the `properties` for Patient Form will be type safe and different from the Signature Attestation and Prescription form event properties.

mrrsnn (approved):

PR #2362, opened 2026-02-20
Settings Page Tab Updates

captDaylight:
1. Rearrange Catalog <> Template ordering, 248
2. Hide Organization tab for non-administrators, 247

### DIFF
--- a/apps/app/src/views/routes/Settings/index.tsx
+++ b/apps/app/src/views/routes/Settings/index.tsx
@@ -46,8 +46,8 @@ const tabs = [
   '/settings/user',
   '/settings/organization',
   '/settings/developers',
-  '/settings/templates',
-  '/settings/catalog'
+  '/settings/catalog',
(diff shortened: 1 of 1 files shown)


apps/app/src/views/routes/Settings/index.tsx
```diff
@@ -81,7 +89,7 @@ export const Settings = () => {
 
   const hasDeveloper = usePermissions(['read:client']);
   const hasTeam = usePermissions(['read:profile']);
-  const hasOrg = usePermissions(['read:organization']);
+  const canManageOrganization = usePermissions(['manage:organization']);
```
captDaylight, line 92: using management permission to determine whether to show the tab

myang5: Agree with this approach over looking for a Prescriber role

myang5 (approved):

PR #2369, opened 2026-02-23
Try auto advance on date select on ios

jakefromphoton:


jakefromphoton:
eh will come back to this later

PR #2373, opened 2026-02-25
add bookmarklet for form filling patient form quickly

jakefromphoton:
* added bookmarklets you can save to your browser to auto fill form events
* saves a phone number via prompt to localstorage
  * second bookmarklet will reset your localstorage if you want to change it


billoran8 (approved):
super cute

myang5 (approved):
Nice! I use Arc browser so there's no bookmarks lmao but pasting the code into the browser console also worked

PR #2380, opened 2026-02-26
Add analytics to prescription form

jakefromphoton:
* Add analytics to prescription form
  * continues pattern from #2359
* e2e test validates analytics are sent via request capture

Part of PHO-271


apps/app/e2e/utils/analytics_expect.ts
```diff
@@ -0,0 +1,58 @@
+import { expect, Page } from '@playwright/test';
+import { getCapturedAnalytics } from './analytics_intercept';
+
+/**
+ * Asserts that a specific field has the expected number of prescription_field_interaction events.
+ */
+export async function expectFieldInteraction(
```
mrrsnn:
```suggestion
export async function expectTrackFieldInteraction(
```

apps/app/e2e/utils/analytics_intercept.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    // Block RudderStack network requests
+    const originalFetch = window.fetch.bind(window);
+    window.fetch = (input: RequestInfo | URL, init?: RequestInit) => {
+      const url = typeof input === 'string' ? input : input instanceof URL ? input.href : input.url;
+      if (url.includes('rudderstack')) {
+        return Promise.resolve(new Response('{}', { status: 200 }));
+      }
+      return originalFetch(input, init);
+    };
```
mrrsnn: Does playwright not have the concept of "spies", where the original functionality still happens but you can call inspect previous method calls?

apps/app/e2e/utils/analytics_expect.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (expectedProperties) {
+    expect(events[index].properties).toEqual(expect.objectContaining(expectedProperties));
+  }
+  return events[index];
+}
+
+export async function findByTrackEventType(page: Page, eventName: string, trackEventType: string) {
+  const all = await getCapturedAnalytics(page);
+  return all.filter((e) => e.event === eventName && e.properties.trackEventType === trackEventType);
+}
```
mrrsnn, line 58: if we were using spies, then this would simply be a search on the spy's recent calls

apps/app/src/instrumentation/analyticsTrackEventListenerUtils.ts
```diff
@@ -44,3 +45,35 @@ export function buildSignatureAttestationFormInteractionPayload(
         : null
   };
 }
+
+export function buildPrescriptionFormInteractionPayload(detail: PrescriptionFormAnalyticsEvent) {
+  const { trackEventType } = detail;
+  // Use a plain object to safely extract optional properties without TS narrowing issues
```
jakefromphoton: Note to self Gonna review this

apps/app/src/views/routes/PrescriptionForm.tsx
```diff
@@ -66,20 +73,64 @@ export const PrescriptionForm = () => {
...
+      'draft_prescription_edited',
+      'order_created',
+      'combine_orders_viewed',
+      'draft_prescriptions_activated',
+      'screening_alert_acknowledged',
+      'screening_alert_canceled',
+      'prescription_form_closed',
+      'prescription_field_interaction',
+      'pharmacy_interaction'
+    ]);
```
mrrsnn: couldn't this be a gloabl? feels weird to initialize it in an effect hook

packages/elements/src/photon-multirx-form-wrapper/photon-multirx-form-wrapper-component.tsx
mrrsnn, line 127: ahhh so we're dispatching analytics in elements directly now? And also bubbling up the event for the app to listen to? does the app also dispatch an event on this bubble up?

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -37,7 +37,7 @@ const validators = {
     'Please select a dispensing unit'
   ),
   daysSupply: message(min(number(), 0), 'Days Supply must be at least 0'),
-  refillsInput: message(between(0, 11), 'Refills must be 0 to 11'),
+  refills: message(between(0, 11), 'Refills must be 0 to 11'),
```
mrrsnn, line 40: bold key change

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -174,6 +174,10 @@ export const AddPrescriptionCard = (props: {
             key: 'treatment',
             value: e.detail.medication
           });
+          dispatchAnalytics({
+            trackEventType: 'prescription_field_interaction',
+            properties: { fieldName: 'treatment', hasValue: true }
+          });
```
mrrsnn: These two events fire at different times. Referring to this one and the one below for the medication search input

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
```diff
@@ -261,6 +273,15 @@ export const AddPrescriptionCard = (props: {
...
                   }}
+                  onBlur={(e) => {
+                    dispatchAnalytics({
+                      trackEventType: 'prescription_field_interaction',
+                      properties: {
+                        fieldName: 'dispenseQuantity',
+                        hasValue: Boolean(e.currentTarget.value)
+                      }
+                    });
+                  }}
```
mrrsnn, line 269: all these events everywhere kinda makes me wonder if they should all just be firing in the form state manager or something. So it's in one spot. Although it is very clear on each element that an event fires.

mrrsnn (approved):
left comments, approving tho. use best judgement

packages/elements/src/photon-multirx-form-wrapper/photon-multirx-form-wrapper-component.tsx
jakefromphoton: Elements will emit the JS event, but not the HTTP request. The wrapper app is responsible for making HTTP requests

apps/app/src/views/routes/PrescriptionForm.tsx
jakefromphoton: I did try that, and then tried to extract the types from the const array and ran into some issues. But I do think this could be exported from the types file 👍

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
jakefromphoton: fixed. one is for add med history item

jakefromphoton: actually this was dead code. removing it. adding the med history item analytic to the patient Card

apps/app/e2e/utils/analytics_intercept.ts
jakefromphoton: switched to `.route()`

PR #2381, opened 2026-02-27
Add initial codegen for clinical-api in elements

myang5:
The clinical app and patient app already have codegen set up, but we don't have codegen for elements. This is probably due to lack of initiative in handling both queries to the lambdas API and services API.

In elements, the lambdas API is wrapped in the SDK, which is then manually typed in TS. We also cannot currently point codegen at lambdas schema due to the AppSync setup. However, we should be able to point codegen at the services API schema similar to what we do in the clinical app.

I've moved a few queries into the sdk/graphql folder to be auto-typed with codegen. Ideally, we would eventually move all queries shared by elements and the clinical app into this folder.

The main difference is the fact that customers may have different versions of elements installed, whereas the clinical app always interacts with the latest version of the code. This shouldn't be an issue though bc we manually copy the `graphql` folder into `dist` in the `build` command, similar to the existing `types` file.


packages/sdk/project.json
```diff
@@ -13,11 +13,12 @@
           "rimraf dist",
           "tsc --emitDeclarationOnly --outDir dist",
           "vite build",
-          "cp src/types.ts dist/types.ts"
+          "cp src/types.ts dist/types.ts",
+          "cp src/graphql/clinical-api/gql/* dist/graphql/clinical-api/gql"
         ],
         "parallel": false
       },
-      "outputs": ["{workspaceRoot}/dist/packages/sdk"]
```
myang5, line 20: This option seems outdated, the `vite build` output goes into `client/packages/sdk/dist` and not `client/dist/packages/sdk`

CLAUDE.md
```diff
@@ -384,4 +393,3 @@ All new utility functions, views, and components must include tests. Match the t
 - **Test files** are excluded from linting (configured in `.eslintrc.json` ignorePatterns)
 - **TypeScript strict mode** enabled; type check with `npx nx run <project>:tsc:boson`
 - **Shared types** belong in `packages/sdk/src/types.ts` since all packages already depend on `@photonhealth/sdk`. For GraphQL response shapes, always use generated types from codegen (`gql/` or `graphql/` directories) rather than defining them manually.
-- **Import types** from `@photonhealth/sdk/dist/types` when only type information is needed (e.g. `import { types } from '@photonhealth/react'` or `import { Env } from '@photonhealth/sdk'`)
```
myang5, line 389: I'm not actually sure what this line is referencing

packages/elements/src/photon-add-medication-history-dialog/photon-add-medication-history-dialog-component.tsx
```diff
@@ -24,15 +21,8 @@ import shoelaceDarkStyles from '@shoelace-style/shoelace/dist/themes/dark.css?in
...
-
-const SEARCH_TREATMENTS = gql`
-  query SearchTreatments($filter: TreatmentFilter!) {
-    treatments(filter: $filter) {
-      id
-      name
-    }
-  }
-`;
+import { SearchTreatmentsQuery } from '@photonhealth/sdk';
```
myang5, line 24: Moved these queries in particular to the `graphql` folder to resolve a codegen error with queries having the same name. We'll need to resolve these kinds of conflicts as we add more queries to codegen.

packages/sdk/project.json
```diff
@@ -13,11 +13,12 @@
           "rimraf dist",
           "tsc --emitDeclarationOnly --outDir dist",
           "vite build",
-          "cp src/types.ts dist/types.ts"
+          "cp src/types.ts dist/types.ts",
+          "cp src/graphql/clinical-api/gql/* dist/graphql/clinical-api/gql"
```
myang5, line 24: Ensures that types are packaged on `npm publish`

packages/sdk/src/graphql/clinical-api/query.ts
```diff
@@ -0,0 +1,20 @@
...
+`);
+
+export const SearchTreatmentOptionsQuery = graphql(`
+  query SearchTreatmentOptionsQuery($filter: TreatmentFilter!) {
+    treatments(filter: $filter) {
+      __typename
+      id
+      name
+    }
+  }
```
mrrsnn, line 19: while we're here, should we just converge these to the same query? the only real difference seems to be __typename

myang5: Thinking about the promise of GraphQL to be able to only fetch the fields you need... these two queries only differ by one field so it's not a big deal, but what is the argument against just fetching all fields on one entity then?

mrrsnn (approved):

billoran8 (approved):

PR #2382, opened 2026-02-28
adding therapeutic classifcations and brand/generic names to analytic…

billoran8:
…s metrics

### DIFF
--- a/apps/patient/src/configs/analytics.ts
+++ b/apps/patient/src/configs/analytics.ts
@@ -88,14 +88,28 @@ interface ContextData {
 }
 
 function mapOrderToContextData(order: Order): ContextData {
-  const medications: ContextDataMedication[] = order.fills.map((fill) => ({
-    id: fill.treatment.id,
-    name: fill.treatment.name,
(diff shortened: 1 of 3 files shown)


apps/patient/src/configs/analytics.ts
```diff
@@ -88,14 +88,28 @@ interface ContextData {
...
+        : undefined;
+
+    return {
+      id: fill.treatment.id,
+      name: fill.treatment.name,
+      strength:
+        'strength' in fill.treatment && fill.treatment.strength ? fill.treatment.strength : '',
+      quantity: fill.prescription?.dispenseQuantity || 0,
+      unit: fill.prescription?.dispenseUnit || '',
+      names,
```
mrrsnn, line 109: this is specifically where "names" breaks down. gonna be confusing in the analytics table.

mrrsnn: not blocking. use your best judgement.

mrrsnn (approved):

PR #2386, opened 2026-03-02
toggle buttons when creating patient based on patientCount

myang5:
Part of PHO-256


myang5:
Initially wanted to pull the patient count from services, but decided to point at lambdas instead. The lambdas `patients` query is already paginated and filtered by organization ID, so it accomplishes my goal of calculating whether the prescriber already has patients while pulling minimal data

mrrsnn (approved):

PR #2399, opened 2026-03-03
Setup boson-targeting e2e test

jakefromphoton:
* e2e tests can be run against `boson` and `neutron` by setting `PLAYWRIGHT_BASE_URL` env var in the `.env.local` file
* I setup two environments on the `client` repo [github settings](https://github.com/Photon-Health/client/settings/environments): boson and neutron.
  * these enable each environment to have different values for env vars when running `smoke_tests` repository_dispatch job :
    * `PLAYWRIGHT_BASE_URL`
    * `PLAYWRIGHT_E2E_ACCOUNT_USERNAME`
    * `PLAYWRIGHT_E2E_ACCOUNT_PASSWORD` 
* Going to add Post-merge `repository_dispatch` calls to run Boson e2e tests after lambdas merge, services-infra boson merge, and potentially dependabot auto-merges some day

This sort of job step in a Github Action will invoke the `smoke-tests`:
```
  - name: Trigger e2e smoke tests
    run: |
      gh api repos/Photon-Health/client/dispatches \
        --field event_type=smoke-test \
        --field 'client_payload[environment]=boson'
    env:
      GH_TOKEN: ${{ secrets.CROSS_REPO_PAT }}
```

Part of PHO-290

### DIFF
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -1,31 +1,28 @@
 name: Playwright Smoke Tests
 on:
   repository_dispatch:
-    types:
-#      - codebuild-complete
-# todo: discuss with team when to run this workflow. (e.g. after client/backend deployments? before?)
+    types: [ smoke-test ]
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: 'Environment to test against'
+        type: choice
+        options:
+          - boson
+          - neutron
+        default: boson
 
 jobs:
   test:
     timeout-minutes: 10
     runs-on: ubuntu-latest
+    environment: ${{ github.event.client_payload.environment || inputs.environment || 'boson' }}
+    env:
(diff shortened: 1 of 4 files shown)


.github/workflows/smoke-tests.yml
```diff
@@ -1,31 +1,28 @@
 name: Playwright Smoke Tests
 on:
   repository_dispatch:
-    types:
-#      - codebuild-complete
-# todo: discuss with team when to run this workflow. (e.g. after client/backend deployments? before?)
+    types: [ smoke-test ]
+  workflow_dispatch:
```
myang5, line 5: Will this allow us to trigger the workflow anytime from the repo Actions tab?

myang5 (approved):
lgtm!

.github/workflows/smoke-tests.yml
jakefromphoton: i dont think so - instead, we'll trigger from Actions running in other repos

PR #2416, opened 2026-03-04
simplify clinical api codegen elements 

myang5:


packages/elements/src/photon-add-medication-history-dialog/photon-add-medication-history-dialog-component.tsx
```diff
@@ -22,7 +22,7 @@ import shoelaceLightStyles from '@shoelace-style/shoelace/dist/themes/light.css?
 import tailwind from '../tailwind.css?inline';
 import styles from './styles.css?inline';
 import { SearchTreatmentsQuery } from '@photonhealth/sdk';
-import { Treatment } from '@photonhealth/sdk/dist/graphql/clinical-api/gql/graphql';
+import { Treatment } from '@photonhealth/sdk/dist/clinical-api/types';
```
myang5, line 25: Simplified import path in sdk/package.json

packages/sdk/codegen.clinical-api.ts
```diff
@@ -1,13 +1,16 @@
...
   // Only include documents that use clinical-api
   documents: ['./src/graphql/clinical-api/*.ts'],
   ignoreNoDocuments: true, // for better experience with the watcher
   generates: {
     './src/graphql/clinical-api/gql/': {
-      preset: 'client'
+      preset: 'client',
+      presetConfig: {
+        fragmentMasking: false
+      }
```
myang5, line 13: [Fragment masking](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#fragment-masking) is a pattern of composing GraphQL fragments into one query to fetch all fields required to render a page, then passing the page's subcomponents only the fields they need to use.

This composition pattern is interesting but it relies on colocating the fragment definitions with the component code, which I want to avoid so we don't have to add each file that needs clinical-api codegen to the `documents` key. Otherwise fragment masking just adds a lot of TS type information that makes it difficult to type helper functions that consume the query results, so we can turn this feature off for now

jakefromphoton (approved):

PR #2418, opened 2026-03-04
new Collapsible particle, load tailwind styles into vitest

myang5:
The Collapsible component unit tests depend on the `.not.toBeVisible` method, which checks if an HTML element or its parents have the `visibility: hidden` property set. Collapsible uses the Tailwind `hidden` class to add this property.

Apparently this is our first time writing a Vitest unit test that depends on loading Tailwind styles into the test environment, so Vitest wasn't setup for it.

Also new Collapsible component is based on Paul's collapsible Patient Info card while preserving the UI there
Screenshot 2026-03-04 at 4 35 47 PM
Screenshot 2026-03-04 at 4 37 11 PM


packages/components/src/particles/Collapsible/index.tsx
myang5, line 1: Much of this functionality is used in the new Supervising Physician Card, so it's not over-engineered :)

packages/components/project.json
```diff
@@ -12,18 +12,16 @@
         "command": "vite build && npx tailwindcss -i ./src/index.css -o ./dist/style.css"
       }
     },
-    "test-watch": {
-      "executor": "nx:run-commands",
-      "options": {
-        "cwd": "packages/components",
-        "command": "vitest"
-      }
-    },
```
myang5, line 21: Not used anywhere

packages/components/project.json
```diff
@@ -12,18 +12,16 @@
...
       "executor": "nx:run-commands",
       "options": {
         "cwd": "packages/components",
-        "command": "vitest run"
+        "commands": [
+          "npx tailwindcss -i ./src/index.css -o ./src/style.css",
+          "vitest run",
+          "rimraf ./src/style.css"
+        ],
+        "parallel": false
```
myang5: I really couldn't get Vitest to import the `dist/style.css` file that is generated in the `build` target. Claude Code said it's because Vite doesn't allow imports from the build directory but I couldn't find any docs on this and also couldn't find a clean workaround.

The tests only worked when `style.css` was present in the `src` directory, so using this hacky workaround for now.

packages/components/vite.config.ts
```diff
@@ -25,6 +25,8 @@ export default defineConfig({
   },
   test: {
     environment: 'jsdom',
-    globals: true
+    globals: true,
+    css: true,
+    setupFiles: ['./src/test-setup.ts']
```
myang5, line 30: [`css: true`](https://vitest.dev/config/css.html) tells Vitest to process css files instead of ignoring them, [`setupFiles`](https://vitest.dev/config/setupfiles.html) indicates files that Vitest should run before each test file

jakefromphoton (approved):

packages/components/src/index.ts
```diff
@@ -1,3 +1,4 @@
+import './index.css';
```
myang5, line 1: This generates index.css in the dist folder

packages/components/project.json
myang5: @jakefromphoton found a cleaner solution 🚀

PR #2419, opened 2026-03-04
migrate from CRA to VITE

jakefromphoton:
* CRA is deprecated and no longer maintained, so we need to get off it eventually. It causes lots of issues managing security vulnerabilities, such as the minimatch issues [here](https://github.com/Photon-Health/client/security/dependabot/235) where react-scripts depends on a very old version.
  * We already use Vite in packages, so it makes sense to switch to Vite in the apps folder.
* Removed `typedocs` for sdk package to fix a CVE, as they were massively out of date and unusable and not worth regenerating until we have a good reason to.
* fixed `koa` critical vuln (used by nx-react to generate more apps in the monorepo)


apps/app/project.json
```diff
@@ -87,44 +87,44 @@
...
         "cwd": "apps/app",
-        "command": "env-cmd -f .env.boson react-app-rewired build",
-        "outputs": "dist/apps/app"
+        "command": "env-cmd -f .env.boson vite build"
       },
       "dependsOn": ["^build", "codegen"]
     },
     "build:neutron": {
       "executor": "nx:run-commands",
+      "outputs": ["{workspaceRoot}/dist/apps/app"],
```
myang5, line 107: Why workspaceRoot and not projectRoot?

packages/react/src/provider.tsx
```diff
@@ -635,7 +635,7 @@ const stub = (): never => {
 };
 
 const PhotonClientContext = createContext<PhotonClientContextInterface>({
-  env: process.env.REACT_APP_ENV_NAME as Env,
+  env: undefined as unknown as Env,
```
myang5, line 638: Why did this change from REACT_APP_ENV_NAME to undefined and not VITE_ENV_NAME?

jakefromphoton: The `env` is always passed into the PhotonProvider at runtime (it's a required param). So it can be safely undefined here. We dont maintain the `packages/react` npm library for external use.

Also, tsconfig does not have `types: "vite/client"` for this `packages/react` library, which would enable the `import.meta.env` and other vite features - but we want to treat this as a library, which shouldn't know its environment as build time.

apps/app/project.json
jakefromphoton: current build process expects the artifacts at the workspace root (e.g. `client/dist`, not `client/apps/app/dist`)

packages/react/src/provider.tsx
myang5: Ah I see, the default value can be `undefined` because in practice (in runtime) it'll always be defined

billoran8 (commented):
lots of stuff being deleted! woah

myang5 (approved):
Image

PR #2431, opened 2026-03-05
Fixing several vulns by updating packages

jakefromphoton:
Fixing some smaller dependabot CVEs by upgrading packages

Very import to pin `"component-register": "0.8.7"` because the patched version includes a breaking change in `solid-element` that breaks our entire Elements setup.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   "devDependencies": {
     "@babel/preset-react": "^7.14.5",
     "@datadog/datadog-ci": "^5.8.0",
+    "@eslint/js": "^9.28.0",
     "@graphql-codegen/cli": "^5.0.0",
     "@graphql-codegen/client-preset": "^4.1.0",
     "@graphql-codegen/typescript-graphql-request": "^6.2.0",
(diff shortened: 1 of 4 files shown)


packages/sdk/src/auth.ts
```diff
@@ -227,7 +226,7 @@ export class AuthManager {
    * @param url - The url which contains the code and state parameters (defaults to window.location.href)
    * @returns
    */
-  public async handleRedirect(url?: string): Promise<RedirectLoginResult<any> | undefined> {
+  public async handleRedirect(url?: string) {
```
jakefromphoton, line 229: fresh npm install caused an update where this type return was no longer valid. but we can use implicit return value, so i removed this

package.json
```diff
@@ -161,6 +161,12 @@
...
     "rollup": "4.59.0",
     "koa": "2.16.4",
-    "minimatch": "^10.2.3"
+    "minimatch": "^10.2.3",
+    "glob": "^10.4.6",
+    "svgo": "^3.3.3",
+    "immutable": "5.1.5",
+    "@microsoft/api-extractor": "^7.57.5",
+    "babel-plugin-jsx-dom-expressions": "0.39.8",
+    "component-register": "0.8.7"
```
billoran8, line 170: just curious in which way did not having this break solidjs? anything we can write a test for? or would it just fail to compile?

billoran8 (approved):

package.json
jakefromphoton: the e2e test fails without these last 2 pins. This issue seems relevant, and i added a comment: https://github.com/ryansolid/component-register/issues/38

jakefromphoton: it broke our solid-element usage, and photon-client/photon-prescribe-workflow just stopped loading entirely in the DOM

PR #2444, opened 2026-03-06
STP Demo reads from token instead of query params

jakefromphoton:
Part of PHO-265


jakefromphoton:
we dont need tokens, just refactored backend to use our shortlink domains

PR #2445, opened 2026-03-07
Add additional properties to Offer Impression + Selection events

mrrsnn:
Requirements here: https://linear.app/photon-health/issue/PHO-300/add-additional-properties-for-order-created-offer-impression-and-offer

- Deriving _offerType_ from a combination of the properties of the Offer and the properties of the Pharmacy
  - For context, goodrx and rxsense don't have formal offers(yet). So we're going to track it as a goodrx/rxsense offer impression/selection if there's a price on the pharmacy card coming from either source. 
- Adding details around med counts, button text, location of selection, etc.
- Trying not to negatively affect Amazon Pharmacy formats


apps/patient/src/utils/offers.ts
```diff
@@ -0,0 +1,21 @@
+import { OfferDetails } from './models';
+import { EnrichedPharmacy, OfferType, OfferTypes } from './models';
+
+export function getOfferType({
```
billoran8, line 4: nit: i like the name `source` more because amazon can have multiple different offer types (PRIME_RX and INSURANCE)

billoran8: though i guess those are cost types?

mrrsnn: Yeah I think Sonia specifically wants offer type to only be either RxSense, GoodRx, or Amazon

billoran8 (approved):

PR #2480, opened 2026-03-18
Multi meds pricing ap

scottdesilva-photon:
- setting up a parallel path to pull offer bundles for multi med orders. Not currently displaying the response because we would like to watch the data coming through in prod for a bit to build up some confidence before displaying. Set up a new tracking event, "Fetched Offer Bundles" for this purpose. 
- showing cash toggle for multi med orders now but hiding price on pharmacy cards via new prop, `showPrice` in PickupPharmacyCardList component
- new coupon tag to show promotions from amazon offers

### DIFF
--- a/apps/patient/src/api/internal.ts
+++ b/apps/patient/src/api/internal.ts
@@ -28,11 +28,7 @@ export const getOffers = async (orderId: string) => {
 
 export const getOfferBundles = async (orderId: string) => {
   const response = await graphQLClient.GetOfferBundlesForOrder({ orderId: orderId });
-  if (response.offerBundles) {
-    return response.offerBundles;
-  } else {
-    throw new Error('No offer bundles found');
-  }
+  return response.offerBundles;
 };
 
 export const getPharmaciesByLocation = async ({
--- a/apps/patient/src/api/internal.ts
+++ b/apps/patient/src/api/internal.ts
@@ -25,6 +25,16 @@ export const getOffers = async (orderId: string) => {
     throw new Error('No offers found');
   }
 };
+
+export const getOfferBundles = async (orderId: string) => {
+  const response = await graphQLClient.GetOfferBundlesForOrder({ orderId: orderId });
+  if (response.offerBundles) {
+    return response.offerBundles;
+  } else {
+    throw new Error('No offer bundles found');
+  }
+};
+
 export const getPharmaciesByLocation = async ({
   searchParams,
   limit,
--- a/apps/patient/src/components/offers/OfferInfo.tsx
+++ b/apps/patient/src/components/offers/OfferInfo.tsx
@@ -1,8 +1,8 @@
 import { Box, HStack, Image, Tag, TagLabel, TagLeftIcon, Text, VStack } from '@chakra-ui/react';
-import { FiInfo, FiStar } from 'react-icons/fi';
+import { FiInfo, FiStar, FiTag } from 'react-icons/fi';
 import { Tooltip } from './Tooltip';
 import { text as t } from '../../utils/text';
-import { OfferDetails } from '../../utils/models';
+import { OfferBundleDetails, OfferDetails, Promotion } from '../../utils/models';
 import { formatPrice } from '../../utils/formatters';
 
 const PreferredTag = () => {
@@ -31,9 +31,32 @@ const CurrentPharmacyTag = () => {
   );
 };
 
+const getLargestPromotionAmount = (promotions: Array<Promotion> | undefined): number =>
+  Math.max(...(promotions ?? []).map((p) => p.amountSaved ?? 0));
+
+const CouponTag = ({ promotions }: { promotions?: Array<Promotion> }) => {
+  const largestPromotion = getLargestPromotionAmount(promotions);
+  return (
+    promotions?.length && (
+      <HStack bg="orange.50" color="orange.500" borderRadius="md" px={1.5} py={0.5}>
+        <FiTag size={10} />
+        {largestPromotion > 0 ? (
+          <Text fontSize="xs">
+            <strong>Up to ${formatPrice(largestPromotion)}</strong> off with coupon if eligible
+          </Text>
+        ) : (
+          <Text fontSize="xs" fontWeight="bold">
+            with coupon if eligible
+          </Text>
+        )}
+      </HStack>
+    )
+  );
+};
+
 interface OfferInfoProps {
   pharmacy?: Pick<OfferDetails['pharmacy'], 'id' | 'name' | 'logo'>;
-  offer: OfferDetails;
+  offer: OfferDetails | OfferBundleDetails;
   isCurrentPharmacy?: boolean;
   isPreferred?: boolean;
 }
@@ -76,6 +99,8 @@ export const OfferInfo = ({ pharmacy, offer, isCurrentPharmacy, isPreferred }: O
 
   const isAmazonPharmacy = pharmacy.id === import.meta.env.VITE_AMAZON_PHARMACY_ID;
 
+  const isOfferBundle = 'medications' in offer; // we can remove this variable once we contract to only using OfferBundleDetails in this component
+
   return (
     <VStack data-testid="pharmacy-info" align="start" w="full">
       {offerTags.length > 0 ? (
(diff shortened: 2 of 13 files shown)


apps/patient/src/api/internal.ts
```diff
@@ -25,6 +25,16 @@ export const getOffers = async (orderId: string) => {
     throw new Error('No offers found');
   }
 };
+
+export const getOfferBundles = async (orderId: string) => {
+  const response = await graphQLClient.GetOfferBundlesForOrder({ orderId: orderId });
+  if (response.offerBundles) {
+    return response.offerBundles;
+  } else {
+    throw new Error('No offer bundles found');
```
billoran8: nit: is it truly an error if no offer bundles are found? doesn't it mean there's just no offers?

apps/patient/src/utils/shouldShowPriceToggle.test.tsx
billoran8, line 62: changing logic behind the pricing toggle might require a few tests to be updato!

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -239,11 +242,19 @@ export const Pharmacy = () => {
...
     const getOffers = async () => {
-      let fetchedOffers: OfferDetails[] | undefined;
+      let fetchedOffers: OfferDetails[] | OfferBundleDetails | undefined = [];
 
       // only fetch offers if we don't have any
       if (!offers) {
-        fetchedOffers = await fetchOffers(order);
+        if (orderIsMultiRx) {
+          // for multi-rx orders, we are only tracking the response right now
+          const offerBundles = await fetchOfferBundles(order);
```
billoran8, line 251: might need to mock out another request?

apps/patient/src/views/pharmacy.utils.tsx
```diff
@@ -63,6 +69,63 @@ export async function fetchOffers(order: Order): Promise<OfferDetails[] | undefi
...
+
+  const amazonOffers = bundles
+    .filter((bundle) => bundle.supplier === 'AMAZON_PHARMACY')
+    .map((bundle) => {
+      const deliveryPromises = bundle.medications
+        .map((m) => m.deliveryEstimate?.deliveryPromise)
+        .filter((promise) => promise !== undefined);
+      // pick the latest delivery promise across all medications by looking at the first character
+      // Priority: '2...' (2-3 day) > '1...' (1-2 day) > 'S...' (same day)
+      const latestDelivery =
```
billoran8: this will only work if there's only 2 days? i've seen upwards of like, 7-8 days lol. maybe we can use regex to extract out the numbers?

billoran8: also imo this specific "find the latest delivery promise" logic should be split into a helper function with very clear tests to verify logic is working as expecto

apps/patient/src/views/pharmacy.utils.tsx
```diff
@@ -63,6 +69,63 @@ export async function fetchOffers(order: Order): Promise<OfferDetails[] | undefi
...
+        deliveryPromises.find((promise) => promise[0] === '2') ??
+        deliveryPromises.find((promise) => promise[0] === '1') ??
+        deliveryPromises[0];
+
+      return {
+        deliveryEstimate: latestDelivery,
+        costType: bundle.pricingType,
+        costAmount: bundle.aggregateCost?.totalAmount,
+        costAmountTitle: bundle.medications[0]?.medicationPrice?.amountTitle, // should move amount titles to the top level instead of per medication in the API but for now we'll just use the first one since we expect them to be the same across medications
```
billoran8, line 91: yeah weird... let's think about doing this?

apps/patient/src/views/pharmacy.utils.tsx
```diff
@@ -63,6 +69,63 @@ export async function fetchOffers(order: Order): Promise<OfferDetails[] | undefi
...
+        retailAmount: bundle.aggregateCost?.totalRetailAmount,
+        retailAmountTitle: bundle.medications[0]?.medicationPrice?.retailAmountTitle, // same as costAmountTitle, should be moved to top level eventually
+        medications: bundle.medications.map((m) => ({
+          name: m.prescription?.treatment?.name,
+          amount: m.medicationPrice?.amount,
+          retailAmount: m.medicationPrice?.retailAmount,
+          promotions: m.medicationPrice?.promotions
+        })),
+        pharmacy: {
```
billoran8, line 100: also i hate this and think we should put pharmacy in tehg raphql as well (but don't love the log bit lol i think that's encoded png)

apps/patient/src/api/internal.ts
scottdesilva-photon: good point

apps/patient/src/utils/shouldShowPriceToggle.test.tsx
scottdesilva-photon: deferring the showPriceToggle logic for now so we can get the tracking going

apps/patient/src/views/pharmacy.utils.tsx
scottdesilva-photon: good call, they actually did change the expected strings since I've started working on this. The new helper function should handle this better and I've added some light tests for now but we can expand on them as necessary

apps/patient/src/views/pharmacy.utils.tsx
scottdesilva-photon: deferring this for right now

apps/patient/src/components/offers/OfferInfo.tsx
```diff
@@ -31,9 +31,32 @@ const CurrentPharmacyTag = () => {
   );
 };
 
+const getLargestPromotionAmount = (promotions: Array<Promotion> | undefined): number =>
```
billoran8, line 34: think we should explicitly make it so we only consider the PRIME_COUPON type or whatever?

apps/patient/src/utils/deliveryPromise.ts
```diff
@@ -0,0 +1,11 @@
+// Parses day ranges via regex (e.g. "Same day" → 0, "...2-3 days" → 3, "...1-4 days" → 4)
+export function getLatestDelivery(deliveryPromises: string[]): string | undefined {
```
billoran8, line 2: love the docu and the isolation, let's get a few tests for this?

apps/patient/src/utils/shouldShowPriceToggle.test.tsx
billoran8: smart move

apps/patient/src/utils/deliveryPromise.test.tsx
```diff
@@ -0,0 +1,27 @@
+import { expect } from 'vitest';
+import { getLatestDelivery } from '../utils/deliveryPromise';
+
+const SAME_DAY = 'Same day';
```
billoran8, line 4: imo this file should be renamed to `deliveryPromise.test.ts`; and the test shoudl always be adjacent to what's being tested (ove it to the utils folder)

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -239,11 +242,19 @@ export const Pharmacy = () => {
...
+          patientAnalytics.track('Fetched Offer Bundles', order, {
```
billoran8, line 252: perf - we will never show anythign and we'll send them to rudderstack

apps/patient/src/components/offers/OfferInfo.tsx
scottdesilva-photon: hmm what if there's another coupon type that we're not aware of yet? Should we monitor the analytics and then include in next batch of updates?

apps/patient/src/utils/deliveryPromise.test.tsx
scottdesilva-photon: yes! I moved the logic for getting latest delivery promise out of pharmacy utils and didn't rename the tests. thanks for pointing out

apps/patient/src/components/offers/OfferInfo.tsx
billoran8: yep good idea, let's monitor add something if we see something. i was thinking thsi just in case they add a new type without telling us in the future.

scottdesilva-photon: cool I'll add to my list of changes to include next because I agree

billoran8 (approved):

PR #2481, opened 2026-03-19
Update "Don't see Pharmacy" div to match pharmacies below

zac-photon:
Currently on the Pharmacy Selection Page there's a mismatch in the alignment of the "mail order options" card and the actual selectable Pharmacy Cards, see image below for an example:
image

The fix for this was to bring the mail orders callout into alignment with the PharmacyCard elements (which were set to be at -2 relational to the margin -- that is to say stretched just outside the parent div).

In order to do so I had to replicate the component structure of those cards, wrapping what was originally an HStack of text elements in a VStack and wrapping the whole thing in a Card with the margin set at -2. Eventually we should align to either remove this "outside the margin" behavior or move it into a reusable element between the components since this now brings these cards out of alignment with the blue Coupon Box above:

Screenshot 2026-03-20 at 11 10 42 AM

### DIFF
--- a/apps/patient/src/views/Pharmacy.tsx
+++ b/apps/patient/src/views/Pharmacy.tsx
@@ -1,6 +1,7 @@
 import {
   Box,
   Button,
+  Card,
   Center,
   CircularProgress,
   Container,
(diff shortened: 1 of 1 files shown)


jakefromphoton (approved):

jakefromphoton:
Woot!

PR #2483, opened 2026-03-19
adding page view events

billoran8:


apps/patient/src/hooks/usePageAnalytics.ts
```diff
@@ -23,6 +24,10 @@ export const usePageAnalytics = ({ pageName, properties }: UsePageAnalyticsProps
         organizationId: order.organization.id,
         organizationName: order.organization.name
       });
+
+      patientAnalytics.track('Page Opened', order, {
```
billoran8, line 28: making page opened events because they will be helpful in mixpanel

billoran8, line 28: there's now a page_opened event which will be accessable in mixpanel

billoran8: all routes have a pageName specified on pageload

myang5 (approved):

PR #2496, opened 2026-03-23
Change pharmacy reason feature flag

myang5:
Part of X-14

Also added command `npm run patient-tau-url <url>` to convert a URL from the shortlinks table to one you can run access locally

Screenshot 2026-03-31 at 1 22 13 PM
Screenshot 2026-03-31 at 1 22 21 PM
Screenshot 2026-03-31 at 1 21 46 PM
Screenshot 2026-03-31 at 1 21 58 PM


apps/patient/src/configs/analytics.ts
```diff
@@ -350,6 +392,25 @@ class RudderAndMixPanelPatientAnalytics implements PatientAnalytics {
       mixpanel.identify(userId);
     }
   }
+
+  async getFlagValue<K extends FlagNames>(
```
billoran8, line 396: cute n simple!

does mixpanel allow us to pass up additional information in the event that, like, we want all people with GLP1s to get variant A (configured on mixpanel's UI?)

apps/patient/src/graphql/mutations.ts
```diff
@@ -13,11 +13,17 @@ export const SET_PREFERRED_PHARMACY = gql`
...
+    $orderId: ID!
+    $pharmacyId: ID!
+    $patientSelectedPrice: Boolean
+    $reason: String
+  ) {
     rerouteOrder(
       orderId: $orderId
       pharmacyId: $pharmacyId
       patientSelectedPrice: $patientSelectedPrice
+      reason: $reason
```
billoran8, line 26: does this also require a graphql schema change?

apps/patient/tsconfig.base.json
```diff
@@ -1,21 +0,0 @@
-{
```
billoran8, line 1: wait help do we mean to delete this file?

tools/patient_tau_url.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
billoran8, line 1: oh what a cute lil script maybe describe what it is and what it's used for?

tools/patient_tau_url.sh
```diff
@@ -0,0 +1,18 @@
...
+
+order_id=$(echo "$url" | grep -oE 'orderId=[^&]+' | cut -d= -f2)
+
+if [ -z "$order_id" ]; then
+  echo "Error: could not extract orderId from URL"
+  exit 1
+fi
+
+echo -e "\nPATIENT URL:"
+echo "http://localhost:3001/?token=$order_id&orderId=$order_id"
```
billoran8: maybe also use the `open` command? it'll open in your preferred browser

apps/patient/src/configs/analytics.ts
myang5: Yes, but not in the `getFlagValue` method. We would configure the flag in Mixpanel's UI to expect certain `context` values, and then in our client-side code we send `context` either on init:
```js
mixpanel.init("YOUR_PROJECT_TOKEN", {
  debug: true,
  flags: {
    context: {
      company_id: "X",
      custom_properties: {
        platform: "web",
      },
    },
  },
});
```
or send Mixpanel the context whenever it's available:
```js
mixpanel.flags.update_context({
  company_id: "Y",
  custom_properties: {
    platform: "mobile",
  },
});
```
([docs](https://docs.mixpanel.com/docs/tracking-methods/sdks/javascript/javascript-flags))

apps/patient/src/graphql/mutations.ts
myang5: Yes, did that here: https://github.com/Photon-Health/services/pull/1393/changes#diff-e9e5478e854846559f3e63cc7e64e260e266bd8108998292b7026c472691468b
backwards-compatible `services` change has already been deployed to photon

apps/patient/tsconfig.base.json
myang5: Oops forgot to comment, but yea it's actually not referenced anywhere. Each app references the root-level `client/tsconfig.base.json`

billoran8 (approved):

PR #2498, opened 2026-03-24
patient app graphql query files can be .graphql instead of .ts

myang5:


myang5:
Closing, this is not worth bothering the team with but good to know all checks passed and patient app successfully builds with this approach

PR #2500, opened 2026-03-25
move out of client

myang5:
Part of PHO-372

Moving secrets into AWS affects a few of our key workflows:
- Codebuild
  - ✅ Moved all env vars into Secrets Manager and set up IAM policies to give Codebuild access https://github.com/Photon-Health/terraform/pull/3229
- Local development
  - ✅ Added Nx targets for populating a local .env file with boson env vars from AWS, assuming that engineers have AWS CLI configured with the right credentials.
  - .env.tau continues to live in the repo to provide local overrides for values in .env
- Github Action checks
  - Many checks run codegen, which relied on an env var to provide the GraphQL schema URL.
    - ✅ We can hardcode the URL in the codegen config file since it's not a sensitive value.
  - E2E tests need all env vars inside the workflow to spin up each app.
    - ✅ @jakefromphoton suggested running the checks against the deployed branch env so we don't need to provision credentials with Secrets Manager access to the GHA.
    -  Some distinct GH concepts I didn't understand before: _Github Actions_ that are triggered on pull requests create _workflows_, which create _status checks_ against the PR branch head (extremely helpful guide https://www.kenmuse.com/blog/creating-github-checks/)
    - Updated our setup so that Codebuild sends a workflow dispatch to the client repo on successful build > E2E test workflow initializes for the app that finished building > test workflow manually creates a status check for the PR branch head with the test success/failure. This check is present in the PR's Checks tab and the list of checks above the Merge button.


.github/workflows/pr-e2e-tests.yml
```diff
@@ -1,31 +1,85 @@
...
+            echo "e2e=npx nx run patient:e2e" >> $GITHUB_OUTPUT
+            echo "codegen=npx nx run patient:codegen" >> $GITHUB_OUTPUT
+            echo "base_url=https://orders-${{ inputs.branch }}.boson.health" >> $GITHUB_OUTPUT
+            echo "report_name=playwright-report-patient" >> $GITHUB_OUTPUT
+            echo "report_path=apps/patient/playwright-report/" >> $GITHUB_OUTPUT
+          else
+            echo "Error: inputs.app must be 'app' or 'patient', got '${{ inputs.app }}'"
+            exit 1
+          fi
+
```
myang5, line 46: Replace matrix with this job. We no longer want to trigger patient and clinical app tests at the same time, only when each app's respective Codebuild project builds successfully

.github/workflows/pr-e2e-tests.yml
```diff
@@ -1,31 +1,85 @@
...
+        name: Add in-progress check to PR
+        id: create-check
+        env:
+          APP_NAME: ${{ inputs.app }}
+          SHA: ${{ steps.get-sha.outputs.sha }}
+          DETAILS_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        with:
+          script: |
+            const scripts = require('./.github/scripts/add-e2e-test-checks.js')
+            return await scripts.createCheck({github, context, core})
```
myang5, line 82: Because this workflow is no longer triggered on pull request, we need to manually add the result of each workflow run to the pull request. We can do this by adding a check to the head SHA.

Before tests begin, create an in-progress check on the PR (displays a check with a yellow loading spinner)

.github/workflows/pr-e2e-tests.yml
```diff
@@ -37,25 +91,37 @@ jobs:
...
+      - uses: actions/github-script@v8
+        name: Update check with result
+        if: ${{ !cancelled() }}
+        env:
+          CHECK_RUN_ID: ${{ steps.create-check.outputs.result }}
+          CONCLUSION: ${{ steps.run-tests.outcome }}
+        with:
+          script: |
+            const scripts = require('./.github/scripts/add-e2e-test-checks.js')
+            await scripts.updateCheck({github, context, core})
```
myang5, line 127: Marks the check we created as complete with either a success/failure result

apps/app/.env.tau
myang5, line 1: tau mainly uses the same env vars as boson but we need to maintain these specific overrides. tau targets in project.json now consume both .env and .env.tau

apps/app/project.json
```diff
@@ -5,78 +5,77 @@
...
       "dependsOn": ["^build", "codegen:tau"]
     },
-    "start:neutron": {
-      "executor": "nx:run-commands",
-      "options": {
-        "cwd": "apps/app",
-        "command": "env-cmd -f .env.neutron vite"
-      },
-      "dependsOn": ["^build", "codegen:neutron"]
-    },
```
myang5, line 73: I assume no one's using this command

.github/workflows/pr-e2e-tests.yml
myang5, line 1: This workflow is no longer triggered by a new pull request, but by a request sent from Codebuild. More on that further down in the file

.github/workflows/pr-e2e-tests.yml
```diff
@@ -1,31 +1,85 @@
...
+      report_name: ${{ steps.set-config.outputs.report_name }}
+      report_path: ${{ steps.set-config.outputs.report_path }}
+    steps:
+      - name: Determine config for app
+        id: set-config
+        run: |
+          if [ "${{ inputs.app }}" = "app" ]; then
+            echo "e2e=npx nx run app:e2e" >> $GITHUB_OUTPUT
+            echo "codegen=npx nx run app:codegen" >> $GITHUB_OUTPUT
+            echo "base_url=https://app-${{ inputs.branch }}.boson.health" >> $GITHUB_OUTPUT
```
myang5, line 33: Instead of building locally by requesting secrets from AWS, run tests against the deployed Codebuild artifact, which requires the workflow to wait until Codebuild is finished

mrrsnn: Clever

apps/app/README.md
```diff
@@ -6,6 +6,12 @@ In local development, it is possible to run the client against your local "tau"
 
 ### Default setup
 
+First, download environment variables used at runtime from AWS:
+
+```
+npx nx run app:pullenv
+```
+
```
jakefromphoton: I had to run `AWS_PROFILE=boson aws sso login` before this command would work. Can we add that or an alternative you had in mind?

Would also be good to specify where this downloads to, e.g. 

"First, download environment variables used at runtime from AWS to the `.env` file via:"

apps/app/codegen.ts
```diff
@@ -1,7 +1,17 @@
...
 import { CodegenConfig } from '@graphql-codegen/cli';
 
+const schemaByEnv: Record<string, string> = {
+  boson: 'https://clinical-api.boson.health/graphql',
+  neutron: 'https://clinical-api.neutron.health/graphql',
+  photon: 'https://clinical-api.photon.health/graphql',
+  tau: 'http://clinical-api.tau.health:8080/graphql'
+};
+
+const env = process.env.VITE_ENV_NAME ?? 'boson';
```
jakefromphoton: I tend to prefer production by default. Devs can always fix their local env, but prod users can't

jakefromphoton (approved):
Looks great overall! Thanks for working on this

apps/app/README.md
myang5: Updated README and added the `login` command - I probably had credentials locally from connecting to DB and didn't get a credentials error

PR #2501, opened 2026-03-25
Update demo offerings with prices and move the PatientAnalysis behind a Provider

zac-photon:
This PR has 2 changes to it! The big one is putting all of our Patient Analytics calls behind a Provider that differentiates "demo" sessions (in which we don't capture any analytics data) and regular sessions (in which we do) Thanks @billoran8 for the ask, this was a fun refactor!

The smaller change is an update to the demo pharmacies, adding coupon prices, and a test for the experience.

Screenshot 2026-03-25 at 1 54 21 PM

### DIFF
--- a/apps/patient/src/Routes.tsx
+++ b/apps/patient/src/Routes.tsx
@@ -23,8 +23,6 @@ export const routeElements = (
       <Route path="/readyBy" element={<ReadyBy />} />
       <Route path="/pharmacy" element={<Pharmacy />} />
       <Route path="/status" element={<Status />} />
-      {/* Leaving this here in case we need to roll back */}
-      {/* <Route path="/status" element={<Status />} /> */}
       <Route path="/canceled" element={<Canceled />} />
     </Route>
     <Route
--- a/apps/patient/src/Routes.tsx
+++ b/apps/patient/src/Routes.tsx
@@ -7,10 +7,18 @@ import { Status } from './views/Status';
 import { Canceled } from './views/Canceled';
 import { NoMatch } from './views/NoMatch';
 import { InfoPage } from './views/Info';
+import { PatientAnalyticsProvider } from './hooks/usePatientAnalytics';
 
 export const routeElements = (
   <Route>
-    <Route path="/" element={<Main />}>
+    <Route
+      path="/"
+      element={
+        <PatientAnalyticsProvider>
+          <Main />
+        </PatientAnalyticsProvider>
+      }
+    >
       <Route path="/review" element={<Review />} />
       <Route path="/readyBy" element={<ReadyBy />} />
       <Route path="/pharmacy" element={<Pharmacy />} />
@@ -19,7 +27,21 @@ export const routeElements = (
       {/* <Route path="/status" element={<Status />} /> */}
       <Route path="/canceled" element={<Canceled />} />
     </Route>
-    <Route path="/info" element={<InfoPage />} />
-    <Route path="*" element={<NoMatch />} />
+    <Route
+      path="/info"
+      element={
+        <PatientAnalyticsProvider>
+          <InfoPage />
+        </PatientAnalyticsProvider>
+      }
+    />
+    <Route
+      path="*"
+      element={
+        <PatientAnalyticsProvider>
+          <NoMatch />
+        </PatientAnalyticsProvider>
+      }
+    />
   </Route>
 );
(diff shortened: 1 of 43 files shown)


captDaylight (approved):

billoran8:
can we make sure the demo doesn't trigger impressions?

apps/patient/src/configs/analytics.ts
```diff
@@ -216,7 +216,37 @@ function mapOrderToContextData(order: Order): ContextData {
...
+    userId?: string;
+    address?: IdentifyTraits['address'];
+    orgId?: string;
+    orgName?: string;
+  }): void {
+    return;
+  }
+}
+
+class RudderAndMixPanelPatientAnalytics implements PatientAnalytics {
```
zac-photon, line 249: This and the Provider itself are probably the biggest change here, the long and short of it is we now have some dependency injection in our patient app, with a Noop Analytics and a Rudder And Mix Panel Analytics and the fetcher can grab one or the other.

apps/patient/src/Routes.tsx
```diff
@@ -7,10 +7,18 @@ import { Status } from './views/Status';
 import { Canceled } from './views/Canceled';
 import { NoMatch } from './views/NoMatch';
 import { InfoPage } from './views/Info';
+import { PatientAnalyticsProvider } from './hooks/usePatientAnalytics';
 
 export const routeElements = (
   <Route>
-    <Route path="/" element={<Main />}>
+    <Route
```
zac-photon, line 14: This file is the only slightly smelly change, since the Provider needs access to the searchParams to dictate if we're "in a demo" or not, we need to attach it _within_ the route element instead of it sitting on top of all routes. Most biz logic is within Main here but unfortunately I need to repeat the call down below in info and *.

billoran8:
love this but want stout or paul to ✅ as well

billoran8 (approved):

apps/patient/src/SendToPatientDemo.test.tsx
```diff
@@ -87,6 +87,16 @@ describe('Send To Patient Demo', () => {
     await waitFor(() => screen.findByText('Order placed'), { timeout: 2500 });
     expect(await screen.findByText('Amazon Pharmacy')).toBeInTheDocument();
   }, 10_000);
+
+  test('displays coupon prices for non-offer pharmacies', async () => {
+    const expectedNumberOfCouponPrices = demoPharmacies.filter((p) => p.price).length;
```
jakefromphoton, line 92: noice!

apps/patient/src/views/ReadyBy.tsx
```diff
@@ -42,12 +42,13 @@ const checkDisabled = (option: string): boolean => {
...
 export const ReadyBy = () => {
   const { order, flattenedFills, enablePrice, fetchOrder } = useOrderContext();
-  usePageAnalytics({ pageName: 'Ready By Time' });
   const navigate = useNavigate();
 
   const [searchParams] = useSearchParams();
   const token = searchParams.get('token');
   const isDemo = searchParams.get('demo');
+  const patientAnalytics = usePatientAnalytics();
+  usePageAnalytics({ pageName: 'Ready By Time' });
```
jakefromphoton, line 51: Are you just keeping these analytics lines together, or was there a reason to move the Page analytics line around? Just curious~

apps/patient/src/setupTests.ts
```diff
@@ -10,6 +10,14 @@ vi.mock('@datadog/browser-rum', () => ({
...
   }
 }));
+const mockPatientAnalytics = {
+  page: vi.fn(),
+  identify: vi.fn(),
+  track: vi.fn()
+};
+vi.mock('./configs/analytics', () => ({
+  getPatientAnalytics: () => mockPatientAnalytics
+}));
```
myang5, line 20: very DRY 👍 I assume this file is configured to run before all tests?

apps/patient/src/configs/analytics.ts
myang5: small nit: less vendor-dependent name, maybe `LivePatientAnalytics` or smth

apps/patient/src/views/ReadyBy.tsx
zac-photon: Oh this is a leftover from an earlier approach where I passed the patientAnalytics down to usePageAnalytics, but I can change it up

apps/patient/src/setupTests.ts
zac-photon: yuuup

jakefromphoton (approved):

PR #2514, opened 2026-03-31
check if already logged in during pullenv command

myang5:
This just prevents the AWS login popup from opening if the script can verify you're already logged in. Stolen directly from @billoran8's bastion script 🙏


billoran8 (approved):
super cute!

PR #2516, opened 2026-03-31
add tracking in rudderstack for mail order modal open/close

scottdesilva-photon:
adding tracking for open close mail order modal


apps/patient/src/components/mail-order-select/MailOrderSelectModal.tsx
```diff
@@ -87,8 +87,10 @@ export function MailOrderSelectModal({
     // track the modal opening event
     if (modalProps.isOpen) {
       datadogRum.addAction('patient_mail_order_modal_opened');
```
mrrsnn: do you mind actually removing all the rum actions in here while we're at it? they're no longer used.

mrrsnn (approved):

PR #2517, opened 2026-03-31
remove price toggle logic and tests

scottdesilva-photon:
- RIP price toggle
- removing tests having to do with price toggle but lmk if you have opinions on adding any sort of tests
- keeping enablePrice/setEnablePrice state because it controls prices being displayed on cards


mrrsnn (approved):
good shit, great cleanup. Just make sure you're still able to see prices. Prescribe yourself some Lipitor or Mounjaro and verify you're seeing prices still

PR #2534, opened 2026-04-03
PrescribeWorkflow data flow refactor investigation

myang5:
Summary of changes:
- refactored `<photon-patient-select>` into `<PatientSelect>`, removing as a web component and export from elements
- removed deprecated `<photon-update-patient-dialog>` web component as an export from elements
- refactored singleton `PatientStore` to instead be initialized whenever `<PatientCard>` is mounted, removes need for manual teardown.


packages/components/src/particles/ComboBox/index.tsx
```diff
@@ -141,21 +143,35 @@ function ComboOptions(props: { children?: JSX.Element }) {
...
+        class="relative cursor-default select-none py-2 pl-3 pr-9 text-gray-400"
+        role="option"
+        aria-disabled="true"
+        tabindex="-1"
+      >
+        <span class="block truncate">{props.children}</span>
+      </li>
+    );
+  }
+
```
myang5, line 173: Ability to create a disabled `ComboOption` to show no results or loading state

packages/elements/src/photon-multirx-form-wrapper/photon-multirx-form-wrapper-component.tsx
```diff
@@ -245,7 +243,6 @@ const Component = (props: {
         closeBody="You will not be able to recover unsaved prescriptions."
         onClosed={() => {
           dispatchClosed();
-          patientActions.clearSelectedPatient();
```
myang5, line 236: `PatientStore` no longer exists as a singleton and is created and destroyed along with `<PatientCard>`, so parent components higher up in the tree no longer need to handle resetting its state

packages/components/src/systems/PatientInfo/index.tsx
```diff
@@ -62,42 +36,23 @@ export type Address = {
...
+  loading?: boolean;
   weight?: number;
   weightUnit?: string;
   editPatient?: () => void;
-  updatedAt?: number;
   address?: Address;
 };
 
 export default function PatientInfo(props: PatientInfoProps) {
-  const client = usePhotonClient();
```
myang5: This component is only used within `<PatientCard>`, so it can use the values in `patientStore.selectedPatient.data` instead of refetching the patient

packages/elements/src/photon-patient-dialog/photon-patient-dialog-component.tsx
myang5, line 1: `<photon-patient-dialog>` was referencing the singleton `PatientStore`. It's used both as a child of `<PatientCard>` and by itself on the clinical app `/patients/new` and `/patients/update routes`.

Without the singleton, it fetches the patient from the backend to account for usage on the routes. Didn't want to blow up the PR further by adding the ability to pass an existing patient object.

packages/elements/src/photon-patient-dialog/photon-patient-dialog-component.tsx
```diff
@@ -1,45 +1,91 @@
...
+`;
+
 const Component = (props: PatientDialogProps) => {
   let ref: any;
   const client = usePhoton();
   const [loading, setLoading] = createSignal(false);
   const [isCreatePrescription, setIsCreatePrescription] = createSignal<boolean>(false);
   const [formStore, setFormStore] = createSignal<any>(undefined);
-  const [selectedStore, setSelectedStore] = createSignal<any>(undefined);
   const [actions, setActions] = createSignal<any>(undefined);
```
myang5, line 71: The current data flow is that the child component `<photon-patient-form>` (only used in this component) owns creating a form store, and then updates this Solid signal on field change. This is goes against a more intuitive model > view data flow and ideally `<photon-patient-dialog>` would own the form store, but didn't want to increase the PR diff further.

packages/elements/src/photon-patient-form/photon-patient-form-component.tsx
myang5, line 1: Instead of using the singleton `PatientStore`, parent component `<photon-patient-dialog>` passes an initial patient object as a prop. `<photon-patient-form>` only owns rendering the form UI and passing changes back up to the parent.

packages/elements/src/photon-multirx-form/components/PatientCard.tsx
myang5, line 1: Instead of using the singleton `PatientStore`, `<PatientCard>` now owns initializing a `patientStore` and passing any data and CRUD methods down to its children.

`patientStore` mainly exists to store and forward patient data to the larger prescribe flow form store. Haven't removed this duplicated store yet, but refactoring <PatientCard> child components so that only <PatientCard> accesses `patientStore` and `patientActions for` allows for easy removal in the future.

packages/components/src/systems/PatientInfo/index.tsx
myang5, line 1: This component is only used within `<PatientCard>`, so it can ingest the values from `patientStore.selectedPatient.data` instead of refetching the patient.

packages/elements/src/stores/patient.ts
```diff
@@ -75,6 +75,14 @@ const createPatientStore = () => {
...
   });
 
+  const setSelectedPatient = (patient: Patient) => {
+    setStore('selectedPatient', {
+      ...store.selectedPatient,
+      isLoading: false,
+      data: patient
+    });
+  };
+
```
myang5, line 85: `setSelectedPatient` differs from `getSelectedPatient` in that it simply updates the store with data from the fetched array `patients.data` instead of initiating an HTTP request to fetch patient data. I couldn't think of a realistic situation where we would need to refetch a single patient instead of just populating data from `patients.data`.

We still need `getSelectedPatient` for cases when:
- the prescribe flow is initialized with a `patientId` > fetch the full patient object
- the prescriber updates patient info insided the prescribe flow > refetch the patient object to update `patientStore`

packages/components/src/particles/InputGroup/index.tsx
```diff
@@ -157,7 +154,7 @@ function InputGroupWrapper(props: InputGroupProps) {
         <Show when={props.error || props.helpText}>
           <p
             class={`text-sm ${props.error ? 'text-red-400' : 'text-gray-500'}`}
-            id={ariaDescribedBy}
+            id={ariaDescribedBy()}
```
captDaylight, line 157: was this not resolving before?

packages/elements/src/photon-patient-dialog/photon-patient-dialog-component.tsx
```diff
@@ -1,45 +1,91 @@
...
 
-const PATIENT_FIELDS = gql`
-  fragment PatientFields on Patient {
+const HAS_PATIENTS_FIELDS = gql`
+  fragment HasPatientsFields on Patient {
     id
   }
 `;
 
+const GET_PATIENT_FOR_FORM = gql`
```
captDaylight, line 29: not to blow up the scope, but when I see this I feel like business logic is moving into elements rather than into components, is there a way that elements can become dumber?

packages/components/src/systems/PatientInfo/index.tsx
```diff
@@ -1,38 +1,12 @@
...
 import { Patient } from '@photonhealth/sdk/dist/types';
-import { usePhotonClient } from '../SDKProvider';
 import Button from '../../particles/Button';
 import Text from '../../particles/Text';
 import Card from '../../particles/Card';
 import formatDate from '../../utils/formatDate';
 import Collapsible from '../../particles/Collapsible';
 
-const GET_PATIENT = gql`
-  query GetPatient($id: ID!) {
```
captDaylight, line 13: where did this go? is it not needed?

packages/elements/src/photon-multirx-form-wrapper/photon-multirx-form-wrapper-component.tsx
captDaylight: awesome, patient store sucked. but I think we need to test the shit out of this. there are all kind of weird places it's used

packages/elements/src/photon-patient-dialog/photon-patient-dialog-component.tsx
```diff
@@ -119,31 +164,32 @@ const Component = (props: PatientDialogProps) => {
...
-    if (
-      store['preferredPharmacy']!.value &&
-      pStore.selectedPatient.data?.preferredPharmacies &&
-      pStore.selectedPatient.data?.preferredPharmacies.length !== 0 &&
-      !pStore.selectedPatient.data?.preferredPharmacies
-        ?.map((x: any) => x?.id)
-        .filter((x: any) => x !== null)
-        .includes(store['preferredPharmacy']!.value)
-    ) {
+    const fetchedPatient = patientQuery()?.patient;
```
captDaylight, line 167: love naming these

packages/components/src/particles/InputGroup/index.tsx
myang5: It was resolving, but if `props.error` or `props.helpText` changed over time then `ariaDescribedBy` wouldn't have changed accordingly since it wasn't being defined as a derived signal.

I tested this by adding code that flips the value of `props.helpText` every 5 seconds, and the value of `ariaDescribedBy` stayed the same while my test `ariaDescribedByReactive` changed.

ESLint surfaces warnings for these kinds of lines, but we suppress warnings.
```
65:66  warning  The reactive variable 'props.patient' should be used within JSX, a tracked scope (like createEffect), or inside an event handler function, or else changes will be ignored    solid/reactivity
```

Kept this change in since I was able to validate the suggestion, and luckily this nuance doesn't affect business logic. But we should have some kind of KTLO task to figure out if there are bugs resulting from a lack of reactivity

packages/components/src/systems/PatientInfo/index.tsx
myang5: Not needed, it's replaced by passing in `props.patient`, which is defined as `patientStore.selectedPatient.data` in `<PatientCard`. The query in `client/packages/elements/src/stores/patient.ts` happens to also include these fields.

Ideally we should define a GQL fragment that contains all the fields to render this component, and use the fragment in a query that fetches patient data for the prescribe flow. But that would require a bigger refactor of how data is passed around.

packages/elements/src/photon-patient-dialog/photon-patient-dialog-component.tsx
myang5: I think the fix here would be a larger refactor of `<photon-patient-dialog>`. It's tricky since it's both used as a full-page component and a child of `<PatientCard>`.

By isolating all the logic the component needs into this file and removing its dependency on `patientStore`, we're setting up for a smoother refactor where the component becomes dumber and no longer owns the GQL request, so I think this is an okay middle step.

myang5:
Rabbit holed too much on this refactor. I'll be creating smaller PRs where I incrementally pull in changes from this PR

PR #2539, opened 2026-04-06
render DoB next to Patient name when searching for patients

zac-photon:
Some prescribers have reported issues of mis-identifying patients when using the drop-down patient search. This update adds patient date of birth to the dropdown:

Screenshot 2026-04-07 at 11 11 44 AM

### DIFF
--- a/packages/elements/src/photon-patient-select/photon-patient-select-component.tsx
+++ b/packages/elements/src/photon-patient-select/photon-patient-select-component.tsx
@@ -83,7 +83,7 @@ const Component = (props: {
           if (!p) return '';
           const element = (
             <div>
-              <span>{p.name?.full}</span> <span>{p.dateOfBirth}</span>
+              <span>{p.name?.full}</span> <span style="color:grey">{p.dateOfBirth}</span>
             </div>
           );
           return element;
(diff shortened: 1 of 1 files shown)


packages/elements/src/photon-patient-select/photon-patient-select-component.tsx
```diff
@@ -79,7 +79,15 @@ const Component = (props: {
         invalid={props.invalid}
         isLoading={store.patients.isLoading || store.selectedPatient.isLoading}
         hasMore={store.patients.data.length % 25 === 0 && !store.patients.finished}
-        displayAccessor={(p) => p?.name?.full || ''}
+        displayAccessor={(p) => {
+          if (!p) return '';
+          const element = (
+            <div>
+              <span>{p.name?.full}</span> <span>{p.dateOfBirth}</span>
```
mrrsnn: I think it would be good taste if we made the text styles different between each. Maybe a smaller text with italic or something for dob?

zac-photon: Changed the DOB to be greyed out!

mrrsnn (approved):

myang5:
Another nit but do we display dates in that format (1996-08-11) anywhere else in the prescribe flow? I think as a non-tech person I'd except to see smth like 08/11/1996. But also if we're not making that UX effort in the prescribe flow fine to keep as is

PR #2554, opened 2026-04-07
fix Add Prescription btn always hidden when creating a refill

myang5:
Part of KLU-120

Solution here is to replace the "Add Prescription" button at the bottom of the prescribe element with a button inside `DraftPrescriptionCard`. The button is no longer tied to the `hideSubmit` prop, which was causing the bug.

Screenshot 2026-04-08 at 11 16 43 AM

### DIFF
--- a/packages/elements/src/photon-multirx-form/components/PrescribeWorkflow.tsx
+++ b/packages/elements/src/photon-multirx-form/components/PrescribeWorkflow.tsx
@@ -692,7 +692,7 @@ export function PrescribeWorkflow(props: PrescribeProps) {
                   enableOrder={props.enableOrder}
                   // If rx form is hidden, we need a button to toggle it to visible
                   // If rx form is visible, we don't need the button
-                  onAddDraft={!showForm() ? () => setShowForm(true) : undefined}
+                  onAddAnotherClick={!showForm() ? () => setShowForm(true) : undefined}
                 />
                 <Show when={props.enableOrder && needsSupervisor()}>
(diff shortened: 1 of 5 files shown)


mrrsnn:
Why is this `Add+` button necessary? Seems kind confusing that it's plopped at the top and then appears below with the same title `Draft Prescriptions` at the bottom after clicking the button. Can't we just automatically add it to the drafts at the bottom when they click `Renew`?

packages/elements/src/photon-multirx-form/components/PrescribeWorkflow.tsx
```diff
@@ -688,13 +683,16 @@ export function PrescribeWorkflow(props: PrescribeProps) {
...
+                  setIsEditing={() => setShowForm(true)}
                   handleDraftPrescriptionsChange={function () {
                     screenDraftedPrescriptions();
                   }}
                   screeningAlerts={screeningAlerts()}
                   routingConstraints={pharmacySelectionContext.routingConstraints()}
                   enableOrder={props.enableOrder}
+                  // If rx form is hidden, we need a button to toggle it to visible
+                  // If rx form is visible, we don't need the button
+                  onAddDraft={!showForm() ? () => setShowForm(true) : undefined}
```
jakefromphoton: rename `onAddDraft` to something like `onAddAnotherDraftPrescriptionClick` / `onAddAnotherClick`

jakefromphoton (approved):

PR #2561, opened 2026-04-07
hide change pharmacy reasons for order error

scottdesilva-photon:
hide the change pharmacy reasons at the bottom of the page when asking patients to go into the app and choose a new pharmacy after hitting an order error


apps/patient/src/views/Status.tsx
```diff
@@ -156,7 +156,13 @@ export const Status = () => {
   };
 
   const handleReroute = () => {
-    const isEnabled = patientAnalytics.getFlagValueSync('change_pharmacy_reasons', false);
+    const hasUnresolvedOrderError = order.exceptions.some(
```
billoran8, line 161: perhaps a test?

scottdesilva-photon: great reminder, thank you

myang5 (approved):

apps/patient/src/views/Status.tsx
billoran8: do we also want to put the unresolved order rorr eason as the reason for the reroute?

apps/patient/src/views/Status.tsx
```diff
@@ -156,7 +156,13 @@ export const Status = () => {
...
+      (e) => e.exceptionType === 'ORDER_ERROR' && !e.resolvedAt
+    );
+    const isEnabled =
+      patientAnalytics.getFlagValueSync('change_pharmacy_reasons', false) &&
+      !hasUnresolvedOrderError;
```
mrrsnn: don't we also need to send the new overridden `reason` with the reroute request?

scottdesilva-photon: yup, just added

mrrsnn (approved):

apps/patient/src/views/Status.tsx
```diff
@@ -157,6 +157,14 @@ export const Status = () => {
 
   const handleReroute = () => {
     const isEnabled = patientAnalytics.getFlagValueSync('change_pharmacy_reasons', false);
+    const hasUnresolvedOrderError = order.exceptions.some(
+      (e) => e.exceptionType === 'ORDER_ERROR' && !e.resolvedAt
+    );
+
+    if (hasUnresolvedOrderError) {
+      handleSelectReason('Order Error');
```
billoran8: doing it this way will trigger the `pharmacy change reason seleced` event; do we want to do taht or should it be skipped?

additionally we're hardcoding the reason to be "order error"; is that what the value will always be? can it anything else? is there any more descriptive data in  order exceptions?

scottdesilva-photon: I'm under the assumption that we wanted it tracked but I'll check with sonia. I think it makes sense to have these metrics easily queryable in MixPanel but I wonder if it's making the experiment data fluffier.

I think hardcoding this reason makes sense here because (as of right now) there's no useful data to pass. The only additional info in the exceptions is orderId. If there are other scenarios for hiding the change pharmacy reasons in the future, we can build out additional checks and hopefully have more data in those exceptions to pass along as the reason

scottdesilva-photon: another good call!! thanks for pointing out. We're now skipping the `pharmacy change reason selected` event and including reason on the `reroute order` event, per sonia's request.

Also, I think down the road, if there are more order error states that we'd want to include here, we could check for other exception types.

billoran8 (approved):
yeah dude

PR #2566, opened 2026-04-09
Mixpanel replay masking

scottdesilva-photon:
- adding mixpanel to provider analytics so replays now work on both patient and provider flows
- defaulting mixpanel to mask all input fields but reveal all text fields...meaning we have to individually mask text. To start, I've masked allthe fields that were being masked in datadog rum sessions
- added more masking on provider side, for both datadog and mixpanel, to hide user and other provider names, emails, address, etc.

provider masking:
https://mixpanel.com/project/4007639/view/4503408/app/profile#distinct_id=usr_01KJR2QZY3V3AKJECZZGTH2GFA&replay_id=6be4fb13-c8e3-43e3-96c9-eed7f9a76116&replay_timestamp=9506.590600013733

### DIFF
--- a/apps/app/src/configs/providerAnalytics.ts
+++ b/apps/app/src/configs/providerAnalytics.ts
@@ -1,37 +1,61 @@
 import { ApiObject, IdentifyTraits, RudderAnalytics } from '@rudderstack/analytics-js';
+import mixpanel from 'mixpanel-browser';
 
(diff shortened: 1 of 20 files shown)


myang5 (approved):
Nice! I assume `mp-mask` is from the docs?

scottdesilva-photon:
> Nice! I assume `mp-mask` is from the docs?

Yup! Docs were pretty good imo and mixpanel provides a lot of customizability but I think this is a good starting point. https://docs.mixpanel.com/docs/session-replay/session-replay-privacy-controls

PR #2571, opened 2026-04-09
update buildspec for mixpanel on clinical app

scottdesilva-photon:
- updating buildspec to find mixpanel env var
- turning off autocapture in both apps because it's too noisy


billoran8 (approved):
yuh

PR #2573, opened 2026-04-09
Experiment -- Remove "Review Your Prescriptions" Page in Patient Flow

mrrsnn:
Per tickets:
https://linear.app/photon-health/issue/X-59/set-up-feature-flag-for-experiment
https://linear.app/photon-health/issue/X-58/build-ui-for-surfacing-prescription-at-the-top-of-marketplace

## Control

https://github.com/user-attachments/assets/269796e2-4e03-42ef-8070-548dbc62be6e

## Treatment

https://github.com/user-attachments/assets/4f60943a-24cb-48b9-a550-e1c1f1989142

### DIFF
--- a/apps/patient/src/configs/featureFlags.ts
+++ b/apps/patient/src/configs/featureFlags.ts
@@ -1,0 +1,22 @@
+export const FeatureFlags = {
+  ChangePharmacyReasons: 'change_pharmacy_reasons',
+  RemoveReviewYourRxPage: 'remove_review_your_rx_page'
+} as const;
+
+export type FlagKeys = (typeof FeatureFlags)[keyof typeof FeatureFlags];
+
+export type FlagValues = {
+  [FeatureFlags.ChangePharmacyReasons]: boolean;
+  [FeatureFlags.RemoveReviewYourRxPage]: {
+    skipReviewPage: boolean;
+    showRxSummaryOnPharmacyPage: boolean;
+  };
+};
+
+export const FEATURE_FLAG_DEFAULTS: FlagValues = {
+  [FeatureFlags.ChangePharmacyReasons]: false,
+  [FeatureFlags.RemoveReviewYourRxPage]: {
+    skipReviewPage: false,
+    showRxSummaryOnPharmacyPage: false
+  }
+};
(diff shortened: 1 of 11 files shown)


apps/patient/src/configs/featureFlags.ts
```diff
@@ -0,0 +1,22 @@
+export const FeatureFlags = {
```
myang5: I like the idea of a string literal type better so we don't have to import `FeatureFlags` all over the app and clean it up when we eventually remove the experiment:
```ts
type FeatureFlags = 'change_pharmacy_reasons' | 'remove_review_your_rx_page';
```
and the rest of the types are like:
```ts
  type FlagValueMap = {
    change_pharmacy_reasons: boolean;
    remove_review_your_rx_page: {
      skipReviewPage: boolean;
      showRxSummaryOnPharmacyPage: boolean;
    };
  };
  type FlagValues = { [K in FeatureFlags]: FlagValueMap[K] };

  const FEATURE_FLAG_DEFAULTS: FlagValues = {
    change_pharmacy_reasons: false,
    remove_review_your_rx_page: {
      skipReviewPage: false,
      showRxSummaryOnPharmacyPage: false,
    },
  };
```

mrrsnn: We kinda have to delete all these mixpanel flag fetching calls and clean it up after the experiment regardless.

mrrsnn: In general i like a const map as an enum to derive string literal union types from because it's just derivative downstream. If the string key changes at some point during development because we decide another name works better, you just have to change it in one spot.

mrrsnn: If there's a general consensus amongst others though, I'll change my ways! Not trying to just dismiss this.

myang5: True that the string could be changed arbitrarily - the current strings are pretty easy to search and replace but that may not always be true.
As long as adding + removal is ergonomic cool w your approach as well

mrrsnn: Update: Polled the office. Midori wins

scottdesilva-photon (approved):

zac-photon (approved):

PR #2604, opened 2026-04-17
fix source maps for elements/components/react packages

jakefromphoton:
* noticed our webapp was showing bad stack traces for errors caused in elements
* adding sourcemaps for our packages to fix this

# example stacktrace pre-fix
```
TypeError: Proxy handler's 'get' result of a non-configurable and non-writable property should be the same value as the target's property
  at gW @ https://app.photon.health/assets/index-ho5SpQR4.js:3426:119169
  at Xh @ https://app.photon.health/assets/index-ho5SpQR4.js:3426:120318
```

### DIFF
--- a/apps/app/vite.config.ts
+++ b/apps/app/vite.config.ts
@@ -2,12 +2,20 @@
 import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react';
 import tsconfigPaths from 'vite-tsconfig-paths';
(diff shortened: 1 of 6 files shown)


myang5:
Dym posting what this looks like postfix?

captDaylight (approved):

mrrsnn (approved):

PR #2605, opened 2026-04-17
Fix failling e2e test for patient

mrrsnn:
Tests became flakey when we introduced a feature flag experiment. Making review page actions optional in patient e2e


apps/patient/src/configs/analytics.ts
```diff
@@ -308,7 +309,7 @@ class RudderAndMixPanelPatientAnalytics implements PatientAnalytics {
       return;
     }
 
-    if (MIXPANEL_TOKEN) {
+    if (MIXPANEL_TOKEN && !IS_E2E_TEST) {
```
billoran8: what about also getting into the habit of having qa-test-user or whatever being in a contorl group and always getting `control`?

mrrsnn: patient is created on each test run. new ID every time, can't really add that as a qa tester every time.

jakefromphoton (approved):

PR #2619, opened 2026-04-20
unified tailwind build for elements + components

myang5:
### Problem

In our current setup, elements need to import two stylesheets:
```typescript
// Generated by components:build, needed in case the element imports anything from components
import photonStyles from '@photonhealth/components/dist/index.css?inline';

// Generated by elements:build
import tailwind from '../tailwind.css?inline';

// Later in the component....
  return (
    <>
      <style>{photonStyles}</style>
      <style>{tailwind}</style>
```

This creates unexpected behavior from these two stylesheets "cascading" into each other and the bottom stylesheet overriding the previous one, creating bugs like these:
```typescript
    return (
      <>
        {/*Using !mt-8 because of tailwind issue in shadowDom elements */}
        {/*when not using the !important modifier*/}
        <p class="font-sans text-lg mt-4 md:!mt-8" role="heading" aria-level="3">
```
### Solution
Since components is basically a helper library for elements and is never rendered on its own, we can resolve this issue by having elements own the Tailwind build process for both elements and components. Then, Tailwind is able to correctly resolve and order the utility classes into a single stylesheet, which we continue importing into each element.

### DIFF
--- a/packages/components/tailwind.base.cjs
+++ b/packages/components/tailwind.base.cjs
@@ -1,0 +1,71 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const defaultTheme = require('tailwindcss/defaultTheme');
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { createThemes } = require('tw-colors');
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const colors = require('tailwindcss/colors');
+
+/** Shared Tailwind base config for components and elements packages. No `content` array — add that per-package. */
+module.exports = {
+  mode: 'jit',
+  theme: {
+    screens: {
+      xs: '475px',
+      ...defaultTheme.screens
+    },
+    extend: {
+      colors: {
+        'photon-blue': '#3182ce',
+        'photon-blue-dark': '#2b6cb0',
+        'photon-light': '#F7F4F4',
+        blue: {
+          25: '#EFF8FF',
+          50: '#D1E9FF',
+          100: '#B2DDFF',
+          200: '#84CAFF',
+          300: '#53B1FD',
+          400: '#2E90FA',
+          500: '#1570EF',
+          600: '#175CD3',
+          700: '#1849A9',
+          800: '#194185',
+          900: '#102A56',
+          950: '#102A56'
+        }
+      },
+      boxShadow: {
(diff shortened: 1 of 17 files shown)


packages/components/tailwind.base.cjs
myang5, line 1: Base config file shared with elements and components. components still needs its own tailwind config for running tests

packages/elements/src/photon-add-medication-history-dialog/photon-add-medication-history-dialog-component.tsx
```diff
@@ -12,7 +12,6 @@ import '@shoelace-style/shoelace/dist/components/menu-item/menu-item';
 import '@shoelace-style/shoelace/dist/components/spinner/spinner';
 import { setBasePath } from '@shoelace-style/shoelace/dist/utilities/base-path.js';
 //Styles
-import photonStyles from '@photonhealth/components/dist/index.css?inline';
 import shoelaceDarkStyles from '@shoelace-style/shoelace/dist/themes/dark.css?inline';
 import shoelaceLightStyles from '@shoelace-style/shoelace/dist/themes/light.css?inline';
 import tailwind from '../tailwind.css?inline';
```
myang5, line 17: All elements continue to import elements/src/tailwind.css, which now includes styles that were previously defined in @photonhealth/components/dist/index.css. This makes @photonhealth/components/dist/index.css redundant.

packages/elements/src/photon-multirx-form/components/AddPrescriptionCard.tsx
myang5, line 1: `xs:flex-row` was never being applied due to unexpected cascading, so combining Tailwind into one stylesheet revealed an issue in the HTML markdown structure here.

packages/elements/src/PhotonFormWrapper/PhotonFormWrapper.tsx
myang5, line 1: photon-form-wrapper was already exporting a mostly Solid PhotonFormWrapper component, so I renamed the files to avoid confusion (human and AI - had to clarify to Claude that this was no longer a web component a few times). Can't move this into components yet b/c it still imports other web components

zac-photon (approved):
Looks great! Approving but just want to make sure we've got a couple screen shots of this not impacting our current flow.

PR #2622, opened 2026-04-21
Stop adding photonhealth/ packages as direct deps

jakefromphoton:
Components and SDK are built directly into the elements dist artifact; this will fix them being detected by consumers scanners (particularly components stuck at v0.0.0 as an unlicensed package)

Bumping version to 0.22.0 because it changes our dependencies significantly.

Tested the Element locally in system-tests successfully; these dependencies are redundant, as they are included in the Elements build into `dist`, and thus can be moved to dev deps (and thus excluded from npm deps)


myang5 (approved):
🚀

PR #2625, opened 2026-04-21
update tooltip on patient app

scottdesilva-photon:
- close on scroll and tap
- add tooltip to fillsummary

https://github.com/user-attachments/assets/07aee9c2-b9fd-4b26-aaed-e1f0e260702c


apps/patient/src/components/offers/OfferInfo.tsx
```diff
@@ -170,15 +170,7 @@ export const OfferInfo = ({ pharmacy, offer, isCurrentPharmacy, isPreferred }: O
           {offer.medications?.map((med) => (
             <HStack key={med.name} w="full" justify="space-between" align="start">
               <VStack align="flex-start">
-                <Tooltip
-                  label={med.name}
-                  placement="top-start"
-                  wrapperProps={{
-                    onClick: (e) => {
-                      e.stopPropagation();
```
mrrsnn, line 178: isn't this necessary to stop the tooltip click from bubbling up and automatically selecting the pharmacy card?

scottdesilva-photon: I added the wrapperProps back just to lean on the side of caution. I originally thought it was causing an issue with my changes when scrolling but it actually looks fine. New behavior works with or without the wrapper props but I feel like it's safer to just keep it cause there might be a scenario I'm failing to see

mrrsnn (approved):

PR #2629, opened 2026-04-22
Use credentials instead of title

jakefromphoton:
* After backfilling titles to credentials for relevant users, we can now use credentials for the supervisor check
* Only showing titles for expected honorific-style titles, since we haven't removed the `MD/DO/NP/PA`-like titles from the data (yet). Most users do not have a valid title set currently (outside users with credentials).

Part of KLU-202

### DIFF
--- a/packages/sdk/src/graphql/clinical-api/query.ts
+++ b/packages/sdk/src/graphql/clinical-api/query.ts
@@ -22,9 +22,7 @@ export const SearchTreatmentOptionsQuery = graphql(`
 export const MeUserQuery = graphql(`
   query MeUserQuery {
     me {
-      name {
-        title
-      }
+      credentials
       address {
         state
       }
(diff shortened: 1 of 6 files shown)


packages/sdk/src/graphql/clinical-api/query.ts
```diff
@@ -22,9 +22,7 @@ export const SearchTreatmentOptionsQuery = graphql(`
 export const MeUserQuery = graphql(`
   query MeUserQuery {
     me {
-      name {
```
billoran8, line 25: any thoughts on possibly putting the formatted name of the doctor in the `name` object? that way the logic can be shared in the backend?

myang5 (approved):
Supervisor change lgtm!

mrrsnn (approved):
nit: would love to be consistent in the verbiage of `title` vs `honorific` but otherwise lgtm

packages/sdk/src/graphql/clinical-api/query.ts
jakefromphoton: yea we can do that after i get the backend updated to fix the `full` name

PR #2634, opened 2026-04-22
use playwright docker image to avoid long runtime for installing deps

myang5:
### Problem
I noticed that the e2e-tests check would fail randomly due to `npx playwright install --with-deps` timing out, for ex:
Screenshot 2026-04-22 at 4 10 54 PM
Full list of canceled e2e jobs (usually due to timeout): https://github.com/Photon-Health/client/actions/workflows/pr-e2e-tests.yml?query=is:cancelled

### Solution
Playwright maintains a [Docker image](https://playwright.dev/docs/docker) with pre-installed dependencies that can be [used in Github Actions](https://playwright.dev/docs/ci#via-containers) to avoid running this step every time.


mrrsnn (approved):
sick, good idea

PR #2639, opened 2026-04-23
Fix redirect on login

zac-photon:
When we were checking if the org setting for web app access is loaded on the main app we were inappropriately redirecting to the root to "refresh" the page. Instead this just dumps the user to a loading screen without the redirect.


apps/app/e2e/sso_login.spec.ts
```diff
@@ -66,7 +66,7 @@ function expectSingleLogoutRedirect(navigationUrls: string[], connection: string
 }
 
 async function expectPhotonHomePageVisible(page: Page) {
```
jakefromphoton, line 68: rename to `expectLandingPageVisible`

PR #2646, opened 2026-04-24
Add Web App Prescribing Override

zac-photon:
We reverted these changes before because we were seeing issues logging in, specifically redirecting away from the login page to the SSO dropdown. This PR reimplements the changes in that PR in a way that does not impact the login flow.

Fortunately this implementation also allows "blocked" users to access the support and settings pages.

Screenshot 2026-04-24 at 2 09 25 PM


apps/app/src/App.tsx
```diff
@@ -63,21 +64,23 @@ export const App = () => {
...
-              <Route path="/prescriptions">
-                <Route path="/prescriptions" element={<Prescriptions />} />
-                <Route path="new" element={<PrescriptionForm />} />
-                <Route path=":prescriptionId" element={<Prescription />} />
-              </Route>
-              <Route path="/orders">
-                <Route path="/orders" element={<Orders />} />
-                <Route path="new" element={<NewOrder />} />
-                <Route path=":orderId" element={<Order />} />
+              <Route element={<AppOverride />}>
```
zac-photon, line 67: Ignore whitespace here, the only diff is wrapping all the inner routes in the override (except the settings/support)

jakefromphoton (approved):

apps/app/src/views/routes/SSOLogin.tsx
```diff
@@ -81,7 +81,7 @@ function isCurrentOriginAllowed(): boolean {
     const currentOrigin = window.location.origin;
     return (
       currentOrigin === 'http://localhost:3000' ||
-      /^https:\/\/app(-[a-z0-9-]+)?\.boson\.health$/.test(currentOrigin) ||
+      /^https:\/\/app(-.*)\.boson\.health$/.test(currentOrigin) ||
```
zac-photon, line 84: Had to change this because my local setup running E2E tests was https://app-zac_no-pho_fix-override.boson.health/

The _'s blocked me here so running the returnTo didn't work and caused the test to fail here but not locally.

apps/app/playwright.config.ts
```diff
@@ -6,6 +6,7 @@ const baseUrl = process.env.PLAYWRIGHT_BASE_URL ?? 'http://localhost:3000';
  * See https://playwright.dev/docs/test-configuration.
  */
 export default defineConfig({
+  timeout: 180_000,
```
zac-photon, line 9: The default timeout for Playwright is 30_000,, we were never setting it higher, with the new GQL call we were exceeding the call pretty consistently.

PR #2648, opened 2026-04-27
Fix bug for Form where additional-notes are ignored in templated rx creation

mrrsnn:
Per form bug

Screenshot 2026-04-24 at 3 19 37 PM
Screenshot 2026-04-24 at 3 19 51 PM


billoran8 (approved):
let's figure out how we can deprecate this feature asap it seems like a cudgel

PR #2649, opened 2026-04-27
Fix boson redirect

zac-photon:
The change I made to the redirect check on SSO login made `app.boson.health` not a viable base URL to redirect from, this fixes that by making the **-your-branch-name** section of the URL optional.

### DIFF
--- a/apps/app/src/views/routes/SSOLogin.spec.ts
+++ b/apps/app/src/views/routes/SSOLogin.spec.ts
@@ -1,0 +1,45 @@
+import { isCurrentOriginAllowed } from './SSOLogin';
+
+type TestCase = {
+  origin: string;
+  expectedAllowed: boolean;
+};
+const testCases = [
+  {
+    origin: 'app.boson.health',
+    expectedAllowed: true
+  },
+  {
+    origin: 'app.neutron.health',
+    expectedAllowed: true
+  },
+  {
+    origin: 'app.photon.health',
+    expectedAllowed: true
+  },
+  {
+    origin: 'app-test-branch-name.boson.health',
+    expectedAllowed: true
+  },
+  {
+    origin: 'app-1234.boson.health',
+    expectedAllowed: true
+  },
+  // Bad ones
+  {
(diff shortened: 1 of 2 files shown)


apps/app/src/views/routes/SSOLogin.tsx
```diff
@@ -81,7 +81,9 @@ function isCurrentOriginAllowed(): boolean {
     const currentOrigin = window.location.origin;
     return (
       currentOrigin === 'http://localhost:3000' ||
-      /^https:\/\/app(-.*)\.boson\.health$/.test(currentOrigin) ||
+      // Test for generated builds from PRs in the form of
```
billoran8, line 85: perhaps a test?

zac-photon: Good call!

apps/app/src/views/routes/SSOLogin.spec.ts
```diff
@@ -0,0 +1,51 @@
+import { isCurrentOriginAllowed } from './SSOLogin';
+
+type TestCase = {
+  origin: string;
+  expectedAllowed: boolean;
+};
+const testCases = [
```
billoran8, line 7: not a defect just a nit! sometiems tablef ormat is cuter (not sure if now is one of those times but eh)

```describe('sum', () => {
  test.each`
    inputA | inputB | expected
    ${1}   | ${1}   | ${2}
    ${2}   | ${3}   | ${5}
  `(
    'should return $expected if input is $inputA and $inputB',
    ({ inputA, inputB, expected }) => {
      expect(sum(inputA, inputB)).toBe(expected)
    }
  )
})```

billoran8 (approved):

apps/app/src/views/routes/SSOLogin.spec.ts
zac-photon: Oh I really prefer the typed object but maybe we can have an arch guild about testing structures

PR #2652, opened 2026-04-29
bugfix order date display in clinical app

myang5:
Part of KLU-253

In the clinical app and elements, we should be aiming to use the `formatDate` util to consistently display all dates in UTC.

Whether we should display in the user's local timezone is a different issue, this just fixes the discrepancy our customer reported.

### DIFF
--- a/apps/app/src/views/routes/Prescription.tsx
+++ b/apps/app/src/views/routes/Prescription.tsx
@@ -28,7 +28,6 @@ import {
 } from '@chakra-ui/react';
 import { FiChevronRight, FiPlus, FiRepeat } from 'react-icons/fi';
 import { GraphQLClient } from 'graphql-request';
-import dayjs from 'dayjs';
 
 import { formatAddress, FormatAddressProps, formatDate } from '../../utils';
 
@@ -485,7 +484,7 @@ export const Prescription = () => {
                             <Text fontSize="sm" color="gray.500">
                               {addressString}
                               {addressString ? <br /> : null}
-                              Created: {dayjs(fill?.order?.createdAt).format('MMM D, YYYY, h:mm A')}
+                              Created: {formatDate(fill?.order?.createdAt)}
                             </Text>
                           </VStack>
                           <Box alignItems="end">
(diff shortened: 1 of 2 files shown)


apps/app/src/views/routes/Prescription.tsx
```diff
@@ -485,7 +484,7 @@ export const Prescription = () => {
                             <Text fontSize="sm" color="gray.500">
                               {addressString}
                               {addressString ? <br /> : null}
-                              Created: {dayjs(fill?.order?.createdAt).format('MMM D, YYYY, h:mm A')}
+                              Created: {formatDate(fill?.order?.createdAt)}
```
myang5, line 487: dayjs is used throughout the patient app so it was allowed to be imported here. No other place in the clinical app uses dayjs

packages/components/src/systems/RecentOrders/RecentOrdersDuplicateDialog.tsx
```diff
@@ -57,7 +58,7 @@ export default function RecentOrdersDuplicateDialog() {
           </Text>
           <Text size="sm" color="gray">
             Written by {state?.duplicateFill?.prescription?.prescriber?.name?.full} on{' '}
-            {new Date(state?.duplicateFill?.prescription?.writtenAt).toLocaleDateString()}
+            {formatDate(state?.duplicateFill?.prescription?.writtenAt)}
```
myang5, line 61: toLocaleDateString with no argument's would format the date in the browser's timezone - use formatDate util instead

mrrsnn: This change will make all the dates appear as 4/23 when they were expecting to see 4/22 in all spots, right?

myang5: My take on the issue is that it's more about confusion around seeing different dates, not expecting to see a specific date

jakefromphoton (approved):

PR #2654, opened 2026-05-01
format order and prescription related dates in clinical app in local time

myang5:
Part of KLU-253

Fixes the fact that we were displaying order and prescription related dates in UTC. This may have caused confusion for orders created after 9pm EST because they would display as the next calendar date in UTC (e.g. 4/23 instead of 4/22).

We need to maintain a separate util `formatDateUTC` for formatting date of birth, which should not shift based on the user's timezone.

### DIFF
--- a/packages/elements/src/utils.ts
+++ b/packages/elements/src/utils.ts
@@ -33,13 +33,6 @@ export const isZip = (zip: string) => {
   return /^\d{5}(-\d{4})?$/.test(zip);
 };
 
-// Takes a date string in the format 'YYYY-MM-DD'
-// and returns it in the format 'MM-DD-YYYY'.
-export const formatDate = (dateString: string) => {
-  const [year, month, day] = dateString.split('-');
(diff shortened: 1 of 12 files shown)


packages/elements/src/utils.ts
```diff
@@ -33,13 +33,6 @@ export const isZip = (zip: string) => {
   return /^\d{5}(-\d{4})?$/.test(zip);
 };
 
-// Takes a date string in the format 'YYYY-MM-DD'
-// and returns it in the format 'MM-DD-YYYY'.
-export const formatDate = (dateString: string) => {
-  const [year, month, day] = dateString.split('-');
-  return `${month}-${day}-${year}`;
-};
-
```
myang5, line 42: Not used anywhere

jakefromphoton (approved):
Maybe worth noting to CLAUDE.md how we want to handle dates going forward. otherwise LGTM!

PR #2655, opened 2026-05-01
pass null to resolver to remove logo

scottdesilva-photon:
brandLogo wasn't getting sent to resolver when set to undefined so it wasn't getting removed


jakefromphoton (approved):

zac-photon (approved):
This is good for now but I'm not in love with the pattern of "null means remove". Not urgent but maybe in the future we add another possible value to the input like 'remove logo'

scottdesilva-photon:
> This is good for now but I'm not in love with the pattern of "null means remove". Not urgent but maybe in the future we add another possible value to the input like 'remove logo'

agreed. I'll create a ticket in linear to track this

PR #2664, opened 2026-05-05
move faqs to backend

scottdesilva-photon:
getting rid of static FAQs and hittinh GET_FAQS_FOR_ORDER instead

### DIFF
--- a/apps/patient/src/components/FAQ.tsx
+++ b/apps/patient/src/components/FAQ.tsx
@@ -9,67 +9,38 @@ import {
   Link,
   Text
 } from '@chakra-ui/react';
+import { ReactNode } from 'react';
 
-const questions = [
-  {
-    question: 'Can I re-route my prescription or do I need to contact my doctor?',
-    answer: (
-      <>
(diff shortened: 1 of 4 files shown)


apps/patient/src/components/FAQ.tsx
```diff
@@ -9,67 +9,38 @@ import {
...
+export type FAQEntry = {
+  question: string;
+  answer: string;
+};
+
+// Renders the answer text with inline markdown-style links: [label](url).
+// Anything outside the link syntax is rendered as plain text.
+// If we need more complex markdown rendering, we can consider adding react-markdown.
+const LINK_PATTERN = /\[([^\]]+)\]\(([^)]+)\)/g;
+const renderAnswersWithLinks = (answer: string): ReactNode[] => {
```
scottdesilva-photon, line 23: lil bit ugly but gets links/phone-numbers rendered without needing to add a new package and enlarge our bundle. Open to discuss though if we feel like there's a better route

myang5 (approved):

PR #2665, opened 2026-05-06
check patient's address before skipping review page

scottdesilva-photon:


apps/patient/src/views/Main.tsx
```diff
@@ -188,7 +188,10 @@ export const Main = () => {
         const removeReviewStepExperiment = await patientAnalytics.getFlagValue(
           'remove_review_your_rx_page'
         );
-        redirectPath = removeReviewStepExperiment.skipReviewPage ? '/pharmacy' : '/review';
+        redirectPath =
+          removeReviewStepExperiment.skipReviewPage && newOrder.patient.address
```
mrrsnn: this might pollute the experiment results. Didn't @billoran8 say we can do this with a cohort in mixpanel?

scottdesilva-photon: I think that was the ideal solution but ended up being a headache to try to set up. Also I think this is not really an experiment anymore but I'll verify with Sonia and remove the flag altogether

scottdesilva-photon: great catch, sonia wants to keep running the experiment so I'm bypassing patients without addresses and not making the call to grab feature flag

mrrsnn (approved):

PR #2676, opened 2026-05-13
next steps copy for AP orders

scottdesilva-photon:
Screenshot 2026-05-13 at 1 06 15 PM

Adding a "next steps" copy just for amazon pharmacy orders. For the sake of not over-engineering, I'm keeping the copy on the frontend as opposed to pulling from a new endpoint that reads from custom-copies collection for the following reasons:
1. custom copy config could make more sense if we had a copy for everyone but needed overrides for specific orgs/pharmacies
2. we'd be making an api call and reading from dynamo every time someone hits order details when only AP orders would be the only one responding with data. According to MP, this button was hit 12.13K in the last week.
3. this text shouldn't change much

if down the road, “Next steps” copy becomes common across all pharmacies, we can revisit and use the custom-copies table and it'll be minimal effort to transition

### DIFF
--- a/apps/patient/src/views/Status.tsx
+++ b/apps/patient/src/views/Status.tsx
@@ -272,8 +272,9 @@ export const Status = () => {
       <OrderDetailsModal
         isOpen={orderDetailsIsOpen}
         onClose={() => setOrderDetailsIsOpen(false)}
-        pharmacyName={order.pharmacy?.name ?? 'My Pharmacy'}
-        pharmacyId={order.pharmacy?.id}
+        pharmacyName={displayPharmacy?.name ?? 'My Pharmacy'}
+        pharmacyId={displayPharmacy?.id}
+        pharmacyLogo={displayPharmacy?.logo}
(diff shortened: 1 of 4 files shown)


apps/patient/src/views/Status.tsx
```diff
@@ -272,7 +272,9 @@ export const Status = () => {
       <OrderDetailsModal
         isOpen={orderDetailsIsOpen}
         onClose={() => setOrderDetailsIsOpen(false)}
-        pharmacyName={order.pharmacy?.name ?? 'My Pharmacy'}
+        pharmacyName={displayPharmacy?.name ?? 'My Pharmacy'}
+        pharmacyId={displayPharmacy?.id}
+        pharmacyLogo={displayPharmacy?.logo}
```
scottdesilva-photon, line 277: wiring up pharmacy logo so we don't always see the blue default icon in the order details page

zac-photon (approved):

PR #2678, opened 2026-05-15
only show "next steps" for AP

scottdesilva-photon:


apps/patient/src/views/Status.tsx
```diff
@@ -276,6 +276,7 @@ export const Status = () => {
         pharmacyId={displayPharmacy?.id}
         pharmacyLogo={displayPharmacy?.logo}
         prescriptions={prescriptions}
+        status={orderState}
```
mrrsnn: In know the variable was already called `orderState` but since the property type is basically fulfillment states, should we rename this to `fulfillmentState`?

scottdesilva-photon: good point, that's a lot more intuitive

apps/patient/src/components/order-details/OrderDetailsModal.tsx
```diff
@@ -31,6 +32,7 @@ export interface OrderDetailsProps {
   pharmacyId?: string;
   pharmacyLogo?: string;
   prescriptions: PrescriptionData[];
+  status?: PrescriptionFulfillment['state'] | 'DELAYED' | 'FILLING' | 'SHIPPED';
```
mrrsnn: could also be `fulfillmentStatus`

mrrsnn (approved):

PR #2680, opened 2026-05-15
Provider side reroutes

mrrsnn:
Fulfilling all the needs of this ticket: https://linear.app/photon-health/issue/X-109/build-ui-for-provider-rerouting
# Technical Notes:
- had to expose a web component wrapper specifically for the order pharmacy select portion
- Had to create a Dialog component because Chakra's modal component portals out and prevents the pharmacy select web component events from bubbling up and being caught in the reroute order modal 
- Had to use a Portal and fixed position for `Combobox` options so that they didn't get clipped by any modal UI and still respected position flipping when the screen changes. 
- A lot of the lift for this is already done on the backend. 
- Tested edge cases for patient with preferred pharmacies, recent orders, rerouting to mail order, rerouting to pick up, rerouting as STP and selecting via patient UI
  - In order to do a full component test for the reroute button, i needed to add a test harness for the clinical-app component tests. 
    - added `app/src/test-utils` and had claude follow similar test harness patterns established elsewhere in the project. 
    - fixtures should grow and be less painful with time
    - harness is mutable for per-test flexibility
    - each component test runs in a separate node process via vite, so each test has its own msw server and should prevent brittle tests
- Was getting frustrated with all the inline graphql queries. Moved them to specific folders `src/(queries|mutations)/(lambda-api|clinical-api)`
  - Can move more there in a follow-up refactor pr

# Functionality: 

### Reroutes to another pharmacy

https://github.com/user-attachments/assets/bd13cae3-c808-4e23-be2f-405b62c8e34c

### Displays an error message when max reroutes is hit
https://github.com/user-attachments/assets/a325b05f-4851-4059-aa14-cd2acff3c8a9

### Shows the order as back in pending selection

https://github.com/user-attachments/assets/21ffafcc-1751-442c-abd4-7658cdb2dcef

### Texts the patient to select a new pharmacy
IMG_6486

### Patient changes pharmacy without needing a reason

https://github.com/user-attachments/assets/af9659f4-2240-4d31-9172-200a4e8789ba

### Routing History Shows Provider for initial route, Patient for the STP Reroute
Screenshot 2026-05-18 at 2 21 29 PM

### DIFF
--- a/apps/app/src/views/components/RerouteOrderButton.tsx
+++ b/apps/app/src/views/components/RerouteOrderButton.tsx
@@ -176,9 +176,10 @@ export function RerouteOrderButton({ order, organizationId }: RerouteOrderButton
             enable-send-to-patient={true}
             enable-delivery-pharmacies={true}
             infer-send-to-patient-pharmacy={false}
+            sticky-tab-header={true}
           />
         </Dialog.Body>
-        <Dialog.Footer>
+        <Dialog.Footer position="sticky" bottom="0">
           <Button variant="outline" size="sm" onClick={handleRerouteCancel} disabled={updating}>
             Cancel
           </Button>
--- a/apps/app/src/views/components/RerouteOrderButton.tsx
+++ b/apps/app/src/views/components/RerouteOrderButton.tsx
@@ -28,8 +28,6 @@ declare global {
 export function RerouteOrderButton({ order, organizationId }: RerouteOrderButtonProps) {
   const { track } = useProviderAnalytics();
   const toast = useToast();
-  const apollo = useApolloClient();
-  const { clinicalClient } = usePhoton();
 
   const [updating, setUpdating] = useState(false);
   const [rerouteModalOpen, setRerouteModalOpen] = useState(false);
@@ -43,6 +41,8 @@ export function RerouteOrderButton({ order, organizationId }: RerouteOrderButton
     () => getOrgMailOrderPharms(organizationId)?.provider?.join(',') ?? '',
     [organizationId]
   );
+
+  const [rerouteOrder] = useRerouteOrderMutation({ order, rerouteTo: pharmacyId });
 
   useEffect(() => {
     // effect to set up event listener that detects pharmacy selection from the web component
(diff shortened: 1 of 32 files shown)


apps/app/src/views/components/RerouteOrderButton.tsx
```diff
@@ -0,0 +1,318 @@
...
+    [organizationId]
+  );
+
+  const [updating, setUpdating] = useState(false);
+  const [rerouteModalOpen, setRerouteModalOpen] = useState(false);
+  const [pharmacyId, setPharmacyId] = useState('');
+  const [fulfillmentType, setFulfillmentType] = useState<string | undefined>(undefined);
+  const confirmButtonDisabled = !!fulfillmentType && !pharmacyId;
+
+  const uniqueTreatments = useUniqueMeds(order);
```
billoran8, line 47: nit: some of these hooks should be in separate files for easier reuse? also should we encourage react to use more hooks?

packages/components/src/systems/PharmacySelect/SendToPatient.tsx
```diff
@@ -47,17 +47,20 @@ const stpStates: {
   }
 };
 
-export function SendToPatient(props: { patientId: string }) {
+export function SendToPatient(props: { patientId: string; inferSendToPatientPharmacy?: boolean }) {
```
jakefromphoton: maybe a good spot for a comment explaining the 3 types of inferred STP pharmacies:
1. Patient's preferred pharmacy
2. Recent order's pharmacy
3. RTPB Coverage Option's pharmacy

jakefromphoton (approved):

apps/app/src/views/components/RerouteOrderButton.tsx
mrrsnn: I'll move `useUniqueMeds` to a common hook folder. But the mutation and query wrappers are kinda specific to the component. I could be convinced into extracting common data hooks that are wrappers around `useQuery` and `useMutation`. but maybe after we stop inlining gql expressions. One step at a time

packages/components/src/systems/PharmacySelect/SendToPatient.tsx
mrrsnn: added a comment

billoran8 (approved):
looks great; maybe lead a discussion on the test harness approach you've taken a next arch guild? and lets codify the hooks-heavy approach into being less nit and more defect if not hfollowed?

apps/app/src/views/components/RerouteOrderButton.tsx
billoran8: yep agreed

PR #2685, opened 2026-05-19
add supervisor prop to prescribe element

myang5:
Part of KLU-271

### DIFF
--- a/packages/components/src/systems/SupervisorProvider/SupervisorProvider.tsx
+++ b/packages/components/src/systems/SupervisorProvider/SupervisorProvider.tsx
@@ -19,9 +19,13 @@ export interface SupervisorContextType {
 
 const SupervisorContext = createContext<SupervisorContextType>();
 
+// At runtime `solid-element`'s customElement parses the `supervisor` HTML
+// attribute as JSON. Valid JSON → object; invalid JSON → the raw string.
+export type SupervisorPrefill = Partial<SupervisorInput> | string;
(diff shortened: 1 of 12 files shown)


packages/components/src/systems/SupervisorProvider/SupervisorProvider.tsx
```diff
@@ -0,0 +1,115 @@
...
+  return <SupervisorContext.Provider value={value}>{props.children}</SupervisorContext.Provider>;
+};
+
+export const createSupervisorFetch = async (
+  client: PhotonClient,
+  supervisorPrefill: SupervisorPrefill
+): Promise<SupervisorResult> => {
+  if (typeof supervisorPrefill === 'string') {
+    return { errors: ['Invalid supervisor json passed in'] };
+  }
```
myang5, line 78: I did notice that the prescribe element seems to automatically attempt to JSON parse props, is that a solid feature or web component feature or?

myang5: Either way I think worth leaving a comment about how supervisorPrefill shouldn't be a string anymore if it's valid JSON

jakefromphoton: the comment is on the type def for `SupervisorPrefill`

jakefromphoton (approved):

PR #2687, opened 2026-05-19
move identify call to after order is loaded and we have patient id

scottdesilva-photon:
Screenshot 2026-05-19 at 4 15 42 PM
Screenshot 2026-05-19 at 4 15 59 PM

we were grabbing the feature flag BEFORE identifying the user, which meant mixpanel was using a default distinct_id to associate the ff (default is to use device id). This resulted in us having two FF fetches - one with the device id and one with the patient id. We wait until after the order is loaded to identify the user so we can pass along relevant context. I just moved this call to happen BEFORE we attempt to grab the flag.

### DIFF
--- a/apps/patient/src/views/Main.tsx
+++ b/apps/patient/src/views/Main.tsx
@@ -106,27 +106,6 @@ export const Main = () => {
   const settings = order?.organization.settings;
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


billoran8 (approved):
good debugging bruh

PR #2706, opened 2026-06-02
Update VITest

zac-photon:
VI Test needs a version bump.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3152,6 +3152,74 @@
         "node": ">=18.0.0"
       }
     },
+    "node_modules/@esbuild/aix-ppc64": {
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -133,7 +132,7 @@
         "vite-plugin-dts": "^4.5.4",
         "vite-plugin-solid": "2.11.6",
         "vite-tsconfig-paths": "^5.1.4",
-        "vitest": "^3.2.4",
```
zac-photon, line 136: Not sure why it was in here twice but fixing that too.

jakefromphoton (approved):

PR #2718, opened 2026-06-10
hide setting pharmacy behind feature flag

scottdesilva-photon:
don't let patient set a preferred pharmacy. Other preferred pharmacy guarded will come from api changes

### DIFF
--- a/apps/patient/src/components/PharmacyCard.tsx
+++ b/apps/patient/src/components/PharmacyCard.tsx
@@ -6,8 +6,18 @@ import customParseFormat from 'dayjs/plugin/customParseFormat';
 import { Pharmacy as EnrichedPharmacy } from '../utils/models';
 import { text as t } from '../utils/text';
 import { PharmacyInfo } from './PharmacyInfo';
+import { useOrderContext } from '../views/Main';
+import { usePatientAnalytics } from '../hooks/usePatientAnalytics';
 
 dayjs.extend(customParseFormat);
+
+// feature flagging preferred pharmacy skip for certain orgs
+// should eventually be replaced with passing context in runtime properties when evaluating flag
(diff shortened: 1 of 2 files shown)


myang5 (approved):
good to know

apps/patient/src/components/PharmacyCard.tsx
```diff
@@ -34,6 +44,15 @@ export const PharmacyCard = memo(function PharmacyCard({
   showPrice = false,
   isCurrentPharmacy = false
 }: PharmacyCardProps) {
+  const { order } = useOrderContext();
+  const patientAnalytics = usePatientAnalytics();
+
+  const isSkipPreferredPharmacyOrg =
+    order?.organization.id === ignorePreferredPharmacyOrgs[import.meta.env.VITE_ENV_NAME ?? ''];
+  const skipPreferredPharmacy = isSkipPreferredPharmacyOrg
+    ? patientAnalytics.getFlagValueSync('skip_preferred_pharmacy').isActive
```
mrrsnn: need to get the flag value from remote, not local

mrrsnn: well, unless we fetch all feature flags with an identifier earlier on in the app... can't remember.

scottdesilva-photon: yeah all flags get fetched right after mixpanel init so by the time patient clicks on pharmacy card, it'll be there

zac-photon: Only drawback is real time changes and a lack of "stickyness" are we running this as an experiment or a config? If it's the former imo we should do remote evals.

scottdesilva-photon: swapped to remote flag!

zac-photon (approved):

PR #2722, opened 2026-06-12
Add claim assisted login when no Org is passed to the photon-client

zac-photon:
This is an extension of the work originally done in https://github.com/Photon-Health/client/pull/2704

The initial approach had some issues with the auto-assigning not properly re-logging in to reset permissions, this change makes use of a new custom claim on our Access Token set on login to an assigned Org. For the best results this should only be used in single-organization contexts. Multi-Org users should still pass the Organization ID in to the Photon Client explicitly.

### DIFF
--- a/packages/components/src/store.ts
+++ b/packages/components/src/store.ts
@@ -299,7 +299,25 @@ export class PhotonClientStore {
         // getAccessToken is triggering the SSO screen to appear, so for auto-login=false, we don't want to call it
         try {
           const token = await this.sdk.authentication.getAccessToken();
-          const decoded: { permissions: Permission[] } = jwtDecode(token);
+          const decoded: {
+            permissions: Permission[];
+            'https://photon.health/assigned_org_id'?: string;
+          } = jwtDecode(token);
+          const assignedOrgId = decoded['https://photon.health/assigned_org_id'];
+          if (
+            !isOrganizationIdSelectedInPhotonClient &&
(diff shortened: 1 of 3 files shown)


packages/components/src/store.ts
```diff
@@ -299,18 +308,44 @@ export class PhotonClientStore {
...
+              appState: {
+                // Preserve the original login's returnTo (recovered by handleRedirect)
+                // so the destination survives the second, org-scoped login.
+                returnTo:
+                  this.redirectAppState?.returnTo ??
+                  `${window.location.pathname}${window.location.search}`
+              }
+            });
+            return;
+          }
```
jakefromphoton, line 335: why early return?

packages/components/src/store.ts
jakefromphoton, line 304: should we remove this? was our previous attempt

zac-photon: Yes it was our original try, but it still does set the org_id in the SDK at this point (since it's not passed from the top level), so we still need it here it's just now actually able to pul the Org ID off the user now that we're logged in.

packages/components/src/store.ts
zac-photon: At this point we really shouldn't continue checking the session if we're going to start a new one. Even if we only decode permissions past this point this is to prevent issues if something more substantial gets added below.

jakefromphoton (approved):

PR #2741, opened 2026-06-22
fix ws vuln (CVE-2026-48779)

jakefromphoton:
Force ws to ^8.21.0 via overrides to collapse the nested ws@8.18.0 copies pinned by @module-federation/dts-plugin and @module-federation/node.

TripleKey found this, not dependabot

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "vite-plugin-solid": "2.11.6",
     "vite-tsconfig-paths": "^5.1.4",
     "vitest": "^4.1.8",
-    "ws": "^8.19.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -82,7 +82,7 @@
     "vite-plugin-solid": "2.11.6",
     "vite-tsconfig-paths": "^5.1.4",
     "vitest": "^4.1.8",
-    "ws": "^8.19.0"
+    "ws": "^8.21.0"
```
mrrsnn, line 85: lol, didn't know we were using websockets on the frontend

mrrsnn (approved):

PR #2743, opened 2026-06-22
add ICD-10 code prescription screening

myang5:
Part of KLU-356

* Updating some typing to use codegen generated types instead of locally-defined types (`ScreeningAlertType` > `PrescriptionScreeningAlert`)
* Moved the prescription screening code out of PrescribeWorkflow into its own provider


packages/components/src/systems/PrescribeEventDispatchProvider.tsx
myang5, line 1: Made this file DRYery by unifying the types used to type the context value and each `dispatch` function

packages/components/src/systems/PrescriptionScreeningProvider.tsx
```diff
@@ -0,0 +1,174 @@
...
+    if (!props.diagnosisCodes) {
+      return;
+    }
+    if (typeof props.diagnosisCodes === 'string') {
+      const error = 'Invalid diagnosis codes json passed in';
+      // TODO: Thoughts on logging to console as well?
+      console.error(error);
+      dispatchDiagnosisCodeError([error]);
+      return;
+    }
```
myang5, line 87: Calling out this TODO
As an external eng it might be more useful to check the console when debugging in testing and in prod

packages/components/src/systems/PrescriptionScreeningProvider.tsx
```diff
@@ -0,0 +1,174 @@
...
+      }
+    }
+
+    if (invalidCodes.length) {
+      const error = `Invalid diagnosis codes detected: ${JSON.stringify(invalidCodes)}`;
+      console.error(error);
+      dispatchDiagnosisCodeError([error]);
+      // TODO: should we abort entirely or continue screening with valid codes?
+      return;
+    }
```
myang5, line 110: Calling out this TODO

packages/components/src/systems/PrescriptionScreeningProvider.tsx
myang5, line 1: Moved the prescription screening code out of PrescribeWorkflow into its own provider, added the new diagnosis code related paths here

mrrsnn (approved):

mrrsnn:
I like the prescription screening provider.

packages/components/src/systems/PrescriptionScreeningProvider.tsx
myang5: Seems like no one has opinions on this so I'll keep the console.error in the code

packages/components/src/systems/PrescriptionScreeningProvider.tsx
myang5: Seems like no one has opinions on this so I'll keep the `return`. The backend pattern is also to throw an error if any input is invalid

PR #2764, opened 2026-06-30
Expand marketplace views to non-mail in prescriber selected Pharmacies

zac-photon:
We want to send Patients, for whom the provider chose a local pickup pharmacy, to the Marketplace to confirm that Pharmacy. We also want to ensure that compound medications are not re-directed either.


jakefromphoton (approved):

apps/patient/src/utils/getOrderFetchRedirectPath.ts
```diff
@@ -7,6 +8,25 @@ export const hasSingleAutoRouteWithNoReroutes = (order: Order): boolean => {
   return routingHistory.length === 1 && routingHistory[0]?.selector === 'AUTO';
 };
 
+const hasSingleProviderRouteWithNoReroutes = (order: Order): boolean => {
```
billoran8: lets' get some tests bruh

zac-photon: ur so right

PR #2767, opened 2026-07-06
Fix routing behavior around patient confirming their pharmacy selection

myang5:
Part of X-213

Core changes:
- We now have the ability for patients to "confirm" their pickup pharmacy on the Pharmacy.tsx page, but the page code doesn't account for that when submitting their selection. This was causing a patient confirmation to use up their reroute - fixing this bug
- Bill confirmed we no longer want to route to the Ready By page at all, so fixed that codepath and updated tests accordingly

Other changes:
- Adding analytics properties that Sonia requested


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -152,6 +157,11 @@ export const Pharmacy = () => {
 
   // selection state
   const [selectedId, setSelectedId] = useState<string>('');
+  // Patient is able to select the existing order pharmacy to "confirm" it
+  const isConfirmation = selectedId === order.pharmacy?.id;
+  const isReroute =
+    (!!order?.isReroutable && !!order?.pharmacy?.id && !isConfirmation) ||
+    !!searchParams.get('reroute');
```
myang5: The main bug fix

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -689,11 +705,24 @@ export const Pharmacy = () => {
...
+    // setSelectedId state update will not have registered yet
+    const isConfirmation = pharmacyId === order.pharmacy?.id;
     patientAnalytics.track('Pharmacy Selected', order, {
       pharmacyId: pharmacyId,
       pharmacyName: selectedPharmacy?.name,
       pharmacyRank: allPharmacies.findIndex((p) => p.id === pharmacyId) + 1,
-      isReroute: !!isReroute,
+      isReroute: isReroute && !isConfirmation,
+      isPreferred: pharmacyId === effectivePreferredPharmacyId,
+      isConfirmation,
```
scottdesilva-photon: nitpicky but isn't having all three of these fields redundant? Couldn't we just have isConfirmation and fulfillmentType?

scottdesilva-photon (approved):

apps/patient/src/Reroute.test.tsx
myang5, line 1: The tests for "Autorouted pharmacy confirmation" are duplicates of tests in Pharmacy.test.tsx, and they probably make more sense there than in a test file called "Reroute". Confirming an autorouted pharmacy involves no reroute.

apps/patient/src/views/Pharmacy.tsx
```diff
@@ -97,6 +103,36 @@ function FulfillmentTypeTabBar() {
...
 
+/** Which action is the patient taking when they submit their pharmacy selection? */
+enum RoutingOperation {
+  /** Patient selects a new pharmacy for an order that already has one */
+  Reroute = 'reroute',
+  /** Patient selects a new pharmacy for an order that doesn't have one */
+  Route = 'route',
+  /** Patient selects existing pharmacy on order, "confirming" it */
+  Confirmation = 'confirmation'
+}
```
myang5, line 107: Was having a lot of trouble holding all the boolean logic in my head, so decided to refactor the code using this enum. This makes it a lot easier to read and write branching paths based on which action the patient is taking.

apps/patient/src/views/Pharmacy.tsx
myang5: isPreferred is its own property, will talk to Sonia about consolidating isReroute and isConfirmation into something like routingOperation

scottdesilva-photon (approved):

apps/patient/src/views/Pharmacy.tsx
billoran8: smart move!

billoran8 (approved):

PR #2772, opened 2026-07-13
untangle nested timeout in Pharmacy page, fix Sent here badge rendering over footer

myang5:
Everyone I've talked to is scared of touching this nested timeout but I'm 100% sure it's only in place to show the green Thank you! button state before navigating way from the page. Refactoring this to make it easier to read and we won't be afraid to refactor this UI in the future

https://github.com/user-attachments/assets/b6776127-29a5-42aa-b014-49cd36518743

### DIFF
--- a/apps/patient/src/components/pharmacy-card/sent-here/SentHereBadge.tsx
+++ b/apps/patient/src/components/pharmacy-card/sent-here/SentHereBadge.tsx
@@ -18,7 +18,6 @@ export const SentHereBadge = ({ top = 0, selected = false }: SentHereBadgeProps)
       top={top}
       left={4}
       transform="translateY(-50%)"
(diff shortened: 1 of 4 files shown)


apps/patient/src/components/pharmacy-card/sent-here/SentHereBadge.tsx
```diff
@@ -18,7 +18,6 @@ export const SentHereBadge = ({ top = 0, selected = false }: SentHereBadgeProps)
       top={top}
       left={4}
       transform="translateY(-50%)"
-      zIndex={1}
```
myang5, line 21: Removing this zIndex fixes:

Image

zac-photon (approved):

PR #2773, opened 2026-07-13
Add draft-prescriptions to photon-prescribe-workflow (prefill + externalId, with or without a source rx)

theoomoregbee:
## What this does

Adds a `draft-prescriptions` attribute to `photon-prescribe-workflow`: a JSON map keyed by a Photon id, whose value is a `PrescriptionOverride` (`daysSupply`, `dispenseAsWritten`, `dispenseQuantity`, `dispenseUnit`, `fillsAllowed`, `instructions`, `notes`, `externalId`). The key's id prefix selects the behavior:

```html
<photon-prescribe-workflow
  draft-prescriptions='{
    "rx_01HQBE...":  { "externalId": "my-id-1", "daysSupply": 90 },
    "med_01GQ7X...": { "externalId": "my-id-2" }
  }'
/>
```

- **`rx_…`** → clone that existing prescription (fetched via `GetPrescription`), then apply the override. Overridden fields win; the rest carries over from the source.
- **`med_…`** → create a fresh draft for that medication (`treatmentId`) from the override. There is no source prescription to inherit from, so the override must carry the prescription's fields; the Photon API validates completeness.
- An entry with an empty override is skipped (for `rx_…` a no-op clone is what `prescription-ids` is for; for a medication there is nothing to create).

Both key types are collected into the same `CreatePrescriptions` bulk call already used for template prefills, so `externalId` is set per-entry at creation time. Like `template-overrides`, this runs at prefill time — the resulting drafts land in Pending Order for the provider to review and edit before sending.

## Tests

`prescribe-workflow-prescription-prefill.test.tsx` (msw):
- `rx_` entry → draft cloned from the source prescription and the override wins (non-overridden fields inherited from source; `GetPrescription` is called).
- `med_` entry → fresh draft created for the medication with the override applied (`externalId` stamped, `treatmentId` seeded); `GetPrescription` is **not** called.
- entries with an empty override (both a `med_` and an `rx_`) are skipped, and the rest of the batch still creates.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jakefromphoton:
Hey Theo! Do you think you could draft a version of this that doesnt require a pre-existing prescription? We'd like to be able to prefill prescriptions in the embed form for fresh accounts as well, not only ones that happen to have a pre-existing prescriptionID to pass in.

something like this:

```
<photon-prescribe-workflow
  draft-prescriptions='{"med_1": {"externalId": "my-id-1", "daysSupply": 90},","med_2":{"externalId": "my-id-2"}}'
/>
```

Looks like a good PR overall but this would align better with the direction we're trying to go in.

theoomoregbee:
> Hey Theo! Do you think you could draft a version of this that doesnt require a pre-existing prescription? We'd like to be able to prefill prescriptions in the embed form for fresh accounts as well, not only ones that happen to have a pre-existing prescriptionID to pass in.
> 
> something like this:
> 
> ```
> <photon-prescribe-workflow
>   draft-prescriptions='{"med_1": {"externalId": "my-id-1", "daysSupply": 90},","med_2":{"externalId": "my-id-2"}}'
> />
> ```
> 
> Looks like a good PR overall but this would align better with the direction we're trying to go in.

Thanks @jakefromphoton +1 on the direction. I think we can fold both cases into a single `draft-prescriptions` so there's one way to prefill a draft regardless of where it's sourced from.

```
<photon-prescribe-workflow
  draft-prescriptions='{
    "rx_01HQBE...": { "externalId": "my-id-1", "daysSupply": 90 },
    "med_01GQ7X...": { "externalId": "my-id-2",... }
  }'
/>
```

when the key is 
* rx_xyz we clone and override 
* for med_00 we create a fresh draft. 

I'll make the changes soon.

theoomoregbee:
closing in favour of https://github.com/Photon-Health/client/pull/2800 & https://github.com/Photon-Health/client/pull/2803

PR #2777, opened 2026-07-20
send photon-analytics-track-events to the analytics api

myang5:
Part of TECH-708

Core changes:
- Created AnalyticsClient which is initialized in PhotonClient. It contains a single method `track` for sending a POST req to analytics-api
- Add `AnalyticsEventListener` to `<photon-client>`, which attaches an event listener to the DOM that calls `AnalyticsClient.track`
- Remove event listeners from the clinical app code since this is now handled by `<photon-client>`
- Add/update test coverage


apps/app/src/instrumentation/analyticsTrackEventListenerUtils.ts
myang5, line 1: These utils were only used when handling `photon-analytics-track-event` HTML events, moved logic into the new `<photon-client>` event listener

apps/app/src/views/routes/NewPatient/PatientForm.tsx
```diff
@@ -44,14 +46,6 @@ export const PatientForm = () => {
...
     const listenerOptions = { signal: abortControllerSignal };
 
-    ref.current.addEventListener(
-      'photon-analytics-track-event',
-      (e: CustomEvent<PhotonEmbedAnalyticsEventInput>) => {
-        trackAnalyticsEvent(e.detail, providerAnalytics.track);
-      },
-      listenerOptions
-    );
-
```
myang5, line 54: Removed these event listeners from the clinical app code. The clinical app also renders `<photon-client>`, so these events will bubble up to the event listener there.

apps/app/src/views/routes/NewPatient/PatientForm.tsx
```diff
@@ -30,8 +28,12 @@ const orgSettingsQuery = graphql(/* GraphQL */ `
 export const PatientForm = () => {
   const ref: MutableRefObject<any> = useRef(null);
   const navigate = useNavigate();
-  const providerAnalytics = useProviderAnalytics();
   const { clinicalClient } = usePhoton();
+  const { track } = useProviderAnalytics();
+
+  useEffect(() => {
+    track('New Patient Page Viewed');
+  }, []);
```
myang5, line 36: This event was originally emitted by the `<photon-patient-dialog>` element, but it made more sense to me that the clinical app would emit a "page viewed" event.

packages/components/src/analytics/AnalyticsEventListener.tsx
myang5, line 1: Core changes here: new component responsible for attaching a `photon-analytics-track-event` listener to the DOM when `<photon-client>` is rendered. Copied over the context-building logic from `client/apps/app/src/hooks/useProviderAnalytics.tsx`. Accepts an optional `appAnalyticsProperties` for when <photon-client> is rendered inside the clinical app

packages/elements/src/photon-patient-dialog/photon-patient-dialog-component.tsx
```diff
@@ -163,18 +163,6 @@ const Component = (props: PatientDialogProps) => {
...
   };
 
-  createEffect(() => {
-    if (props.open) {
-      dispatchAnalyticsTrackEvent(
-        'pageViewed',
-        {
-          name: props.patientId ? 'Update Patient Page Viewed' : 'New Patient Page Viewed'
-        },
-        ref
```
myang5, line 173: Moved this event tracking into the clinical app layer - made more sense to me that a "page viewed" event would be fired by the clinical app and not an element

packages/sdk/src/analytics/AnalyticsClient.ts
myang5, line 1: Core changes here: new class that wraps a POST request to analytics-api. Kept it as context-less as possible in case we use it for the patient app one day

packages/sdk/src/analytics/clinicalAnalyticsTypes.ts
```diff
@@ -10,20 +10,10 @@ export type DraftPrescriptionSource = 'form' | 'med_history_refill' | 'prefill';
...
-      hasPrefillTemplateIds: boolean;
-      hasPrefillPrescriptionIds: boolean;
-      hasPrefillWeight: boolean;
-      weightUnit: string;
-    }
-  | { name: 'Signature Attestation Page Viewed'; attestationVersion: string };
+export type PageViewEvent = {
+  name: 'Signature Attestation Page Viewed';
+  attestationVersion: string;
+};
```
myang5, line 16: Left this "page viewed" event for now since it wasn't easy to extract into the clinical app - quirk of our React app + web component architecture

packages/sdk/src/lib.ts
myang5, line 1: Core changes here: initialize AnalyticsClient

apps/app/e2e/create_prescription.spec.ts
myang5, line 1: Copied some of the assertions out of `apps/app/src/order-workflow-analytics.test.tsx` into this file, now that analytics tracking is split across the clinical app layer and embed layer.

apps/app/src/order-workflow-analytics.test.tsx
myang5, line 1: Update this file to use some of the global test-utils instead of local utils, removed tests for functionality that are no longer owned by the clinical app layer

packages/sdk/src/analytics/AnalyticsClient.ts
```diff
@@ -0,0 +1,71 @@
...
+      'Content-Type': 'application/json',
+      ...getAuthHeaders({ token: this.token, isServices: true })
+    };
+
+    if (this.isNonProduction) {
+      console.log(`📊 [Analytics: To Analytics API] ${body.event}`, enrichedBody.properties);
+    }
+
+    try {
+      await fetch(`${this.apiUrl}/event`, {
```
zac-photon, line 61: Do we need to await this here? I'm just a bit worried about slowing down the elements since we can get error logs from the API itself.

packages/components/src/analytics/AnalyticsEventListener.tsx
```diff
@@ -0,0 +1,111 @@
...
+  /* Pass extra properties when photon-client is rendered in our web app */
+  appAnalyticsProperties?: Record<string, unknown>;
+  children: JSXElement;
+}) => {
+  const client = usePhotonClient();
+  const [contextData, setContextData] = createSignal<ProviderContextData | null>(null);
+
+  onMount(async () => {
+    // Fetch me + organization data via GraphQL
+    const { data } = await client.apolloClinical.query({ query: AnalyticsContextQuery });
```
zac-photon: Similar worry RE: latency here, if we're only using this for inside the elements and they're always going to be forwarded to the Analytics API why not make this request there?

packages/sdk/src/analytics/AnalyticsClient.ts
myang5: The `await` here should be fine as long as the actual `track` method isn't awaited. We need the `await` here to actually catch errors and console.error them

packages/components/src/analytics/AnalyticsEventListener.tsx
myang5: I wanted to keep the analytics API as agnostic of client context as possible since we do want to eventually forward all analytics through that api. This request should only happen once anyway, after that contextData is saved in state

mrrsnn (approved):
great job

PR #2780, opened 2026-07-22
stop calling pick up pharmacies twice

scottdesilva-photon:
this useEffect fires the getPharmaciesByLocation call and sets initialLoad to false but also has a dependency on initial load causing it to fire twice


apps/patient/src/views/Pharmacy.tsx
```diff
@@ -666,8 +666,7 @@ export const Pharmacy = () => {
     longitude,
     loadPharmacies,
     order?.readyBy,
-    toast,
-    initialLoad
+    toast
```
mrrsnn, line 669: are you hitting a new react hooks linting error here? If so, this `initialLoad` state could probably just be a ref

scottdesilva-photon: lint looks gucci

scottdesilva-photon: what are the chances we change the following from 'warn'? If we did, linter would complain and someone could accidentally reintroduce this. But if unlikely, then I'm happy with keep the local state 
`react-hooks/exhaustive-deps': 'warn',`

mrrsnn: probably wouldn't change that from warn to error until we rebuild this. As long as this still works as expected when there are no pharmacies in the location search results due to distance, then i'm good

myang5 (commented):
Not sure I understand how removing the initialLoad dep would affect the values inside the effect body, should the useEffect callback body just be wrapped in a conditional that checks if `initialLoad` is false?

mrrsnn (approved):

scottdesilva-photon:
> Not sure I understand how removing the initialLoad dep would affect the values inside the effect body, should the useEffect callback body just be wrapped in a conditional that checks if `initialLoad` is false?

we're setting initialLoad in the effect body so it automatically re-triggers the effect when initialLoad is a dep. Removing it makes it so it only calls once on the page load. 

If we wrapped the callback in that conditional, I'm pretty sure we just wouldn't trigger this useEffect on page load (because initialLoad is defaulted to true) which is not want we want. Ultimately this useEffect stinks because there are so many nested conditionals/pathways but I think this is a low risk change

PR #2788, opened 2026-07-23
add ability to retry NPM Publish GHA

myang5:
If we add `workflow_dispatch`, this shows a button on the action page:
actions-workflow-dispatch

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - photon
-  # Let's us retry the action if it fails on push to photon
+  # Lets us retry the action if it fails on push to photon
   workflow_dispatch:
 
 permissions:
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -43,47 +43,39 @@ jobs:
       - name: Check for version updates
         id: check
         run: |
-          PACKAGES=""
-          for file in $(git diff --name-only HEAD~1...HEAD | grep '/package.json' | grep 'packages/'); do
-            if git diff HEAD~1...HEAD -- $file | grep '\"version\"'; then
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -41,47 +43,39 @@ jobs:
       - name: Check for version updates
         id: check
         run: |
-          PACKAGES=""
-          for file in $(git diff --name-only HEAD~1...HEAD | grep '/package.json' | grep 'packages/'); do
-            if git diff HEAD~1...HEAD -- $file | grep '\"version\"'; then
```
myang5, line 46: `git diff --name-only HEAD~1...HEAD` only looks for version changes in the most recent commit on the `photon` branch. If we allow engineers to retry the publish at any time, the version change might be missed.

The new script gets `PACKAGE_VERSION` from elements/package.json on the HEAD of `photon` branch and `PUBLISHED_VERSION` from npm and compares them to see if we need to publish `PACKAGE_VERSION`.

.github/workflows/npm-publish.yml
```diff
@@ -41,47 +43,39 @@ jobs:
...
+            echo "Version bump found: $PUBLISHED_VERSION -> $PACKAGE_VERSION"
+            echo "version=$PACKAGE_VERSION" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Publish changed package
         run: |
-          PACKAGES="${{ steps.check.outputs.packages }}"
-          if [ -z "$PACKAGES" ]
+          PACKAGE_VERSION="${{ steps.check.outputs.version }}"
+          if [ -z "$PACKAGE_VERSION" ]
```
myang5, line 59: It seems like we only want to publish elements at this point and not sdk?

mrrsnn (approved):

PR #2802, opened 2026-07-29
Fix bad redirects

zac-photon:


packages/sdk/src/auth.ts
```diff
@@ -11,6 +11,37 @@ const CODE_RE = /[?&]code=[^&]+/;
...
+ */
+const AUTH_REDIRECT_PARAMS = [
+  'code',
+  'state',
+  'error',
+  'error_description',
+  'error_uri',
+  'iss',
+  'photon'
+];
```
jakefromphoton, line 31: Is auth0 SDK supposed to manage/remove these itself, and it fails for some reason?

packages/elements/src/photon-client/photon-client-component.tsx
```diff
@@ -68,8 +68,9 @@ export const PhotonClientComponent = (props: PhotonClientProps) => {
   const [store] = createSignal<PhotonClientStore>(client);
 
   const handleRedirect = async () => {
-    await store()?.authentication.handleRedirect();
-    if (props.redirectPath) window.location.replace(props.redirectPath);
+    const exchanged = await store()?.authentication.handleRedirect();
```
jakefromphoton: is exchange a synonym for redirect in this context? why not `redirected` as the name?

packages/sdk/src/auth.ts
zac-photon: No from what I can see from the docs for Auth0 they delegate cleanup to the app

packages/elements/src/photon-client/photon-client-component.tsx
zac-photon: Sure! I was thinking successful token exchange but happy to update

jakefromphoton:
Could you add a list of scenarios worth manually testing to this PR (for both web app and embed)? Probably the invite flow at least. I just get nervous when we change auth, even if these changes look like a big improvement overall.

zac-photon:
> Could you add a list of scenarios worth manually testing to this PR (for both web app and embed)? Probably the invite flow at least. I just get nervous when we change auth, even if these changes look like a big improvement overall.

Chatted live, and comfortable with the system tests and local reproduction.

jakefromphoton (approved):

PR #2803, opened 2026-07-30
initial-prescriptions embed prop, analytics on how embed is configured

myang5:
Part of TECH-756

Core changes:
- Add an initial-prescriptions attribute to the embed, which takes in a JSON array of objects to create draft prescriptions without requiring existing entities in our DB (how template-ids and prescription-ids work)
- Emit analytics event to understand what attributes consumers pass in to the prescribe element
- Refactor AnalyticsEventListener in photon-client to queue events dispatched while context data is being fetched instead of dropping the events

Other changes
- While writing up the attributes event, I reviewed how some of the other prefill attributes are consumed and updated them to fit our existing definition
- Documented the existing definition in a Prefill TS type


packages/components/src/systems/DraftPrescriptions/utils/initialPrescriptionInputSchema.ts
myang5, line 1: Using zod to parse and validate the new `initial-prescriptions` prop value since that's the direction we want to move in. This schema mirrors the basic validation in place on the frontend DraftPrescriptionForm. Further validation is owned by the backend.

Chose not to refactor the validation in DraftPrescriptionForm to share this zod schema since that would require a larger effort to trash `createFormStore`, but it would be nice to do that eventually

packages/components/src/systems/DraftPrescriptions/utils/usePrefillPrescriptions.ts
myang5, line 1: Core change: handling `initial-prescriptions` value to create draft prescriptions

Code related to auto-creating draft prescriptions was getting beefy so I extracted it into its own hook. This covers creating draft prescriptions from templates (existing), existing prescriptions (existing), and prescription inputs JSON (new)

packages/components/src/systems/DraftPrescriptions/DraftPrescriptionsProvider.tsx
myang5, line 1: Large diff here from extracting usePrefillPrescriptions

packages/components/src/utils/Prefill.ts
myang5, line 1: Since the number of HTML attributes that take in JSON is growing, I wanted to add a type that enforced and documented proper validation of those attributes

packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
myang5, line 1: Noticed that the existing `disableList` attribute wasn't properly typed and validated - another motivation for adding the Prefill type

packages/elements/src/photon-multirx-form/photon-prescribe-workflow-component.tsx
myang5, line 1: Sending a new analytics event that covers what attributes are passed to the prescribe flow element. Will be useful for answering multiple questions including what attributes are most common and how prefill prescription attributes are used

packages/sdk/src/analytics/clinicalAnalyticsTypes.ts
```diff
@@ -1,19 +1,73 @@
...
+  | 'form'
+  | 'med_history_refill'
+  | 'template_prefill'
+  | 'prescription_id_prefill'
+  | 'initial_prescriptions_prefill';
 
 // ---------------------------------------------------------------------------
-// Page View Events — each page gets a unique event name
+// Element View Events — each element gets a unique event name
 // ---------------------------------------------------------------------------
```
myang5, line 16: Got confused by this naming initially since to me "Page" means a route in a web app, which doesn't cover when elements are dropped individually into customers's apps

packages/sdk/src/analytics/clinicalAnalyticsTypes.ts
```diff
@@ -1,19 +1,73 @@
...
 
-export type PageViewEvent = {
-  name: 'Signature Attestation Page Viewed';
-  attestationVersion: string;
-};
+export type ElementViewEvent =
+  | {
+      name: 'Prescribe Workflow Viewed';
+      patientId?: string;
```
myang5, line 21: Def open to feedback on what properties are included here. I tried to avoid passing whole JSON objects since Mixpanel docs generally recommend primitive values or arrays of primitive values

packages/components/src/analytics/AnalyticsEventListener.tsx
myang5, line 1: Instead of ignoring analytics events while we haven't fetched the analytics context, queue the events in memory until the fetch is complete, then flush them to the API.

The exact implementation is a bit arbitrary because I really don't expect the queue to contain more than 5 events. This might change as we add more events so we can reevalute if we're always hitting MAX_QUEUED_EVENTS

apps/patient/src/SendToPatientDemo.test.tsx
myang5, line 1: This test suite kept failing randomly - we weren't setting the timeout that waits for the button animation correctly

packages/components/src/systems/DraftPrescriptions/utils/usePrefillPrescriptions.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    if (hasPrefills && !!props.patientId && !hasCreatedPrefillPrescriptions()) {
+      setHasCreatedPrefillPrescriptions(true);
+      setIsLoadingPrefills(true);
+
+      try {
+        const rxNotesPrefill = options.rxNotesPrefill();
+        const newRxs: Prescription[] = [];
+
+        const rxPromises = [
```
mrrsnn: I think we gotta break up some of these promises for readability. Split them out into different function calls and invert the selection statement at the top so there's lest nesting. Should improve readability greatly

packages/components/src/analytics/AnalyticsEventListener.tsx
```diff
@@ -33,6 +33,10 @@ function flattenSnapshot(fields: FieldCompletionSnapshot): Record<string, boolea
   );
 }
 
+// The context query usually resolves in less than a second, so we'll probably never
+// reach this max. Can adjust as needed.e
```
zac-photon:
```suggestion
// reach this max. Can adjust as needed.
```

packages/components/src/analytics/AnalyticsEventListener.tsx
```diff
@@ -95,29 +132,19 @@ export const AnalyticsEventListener = (props: {
...
-        }
-
-        const properties: ApiObject = {
-          ...context,
-          ...payload,
-          ...props.appAnalyticsProperties
-        };
-        // Don't await this method, tracking analytics shouldn't block the JS thread
-        client.analytics.track({ event: name, userId: context.providerId, properties });
+        trackEvent(e.detail, context);
```
zac-photon, line 147: love the consolidation here

packages/components/src/analytics/AnalyticsEventListener.tsx
zac-photon: Only a nit about PRs themselves and not really a blocker here, but would we want to just put this change in a different PR?

zac-photon (approved):

packages/components/src/analytics/AnalyticsEventListener.tsx
myang5: I probably should have, my bad. I thought it would be quick work to emit an analytics event when the embed is first rendered but like most things in the codebase that assumption is never true lol

packages/components/src/utils/Prefill.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * The prescribe flow element has several "prefill" attributes
+ * that accept complex JSON strings.
+ * At runtime, `solid-element`'s customElement attempts to parse these values.
+ * Valid JSON → object; invalid JSON → the raw string.
```
jakefromphoton: A bit confused about `Valid JSON → object; invalid JSON` and the `Unverified<T>` naming. What is the difference between valid and verified

packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
```diff
@@ -13,17 +13,17 @@ import { PhotonDropdown } from '../photon-dropdown';
 // Types
 import { Medication, PrescriptionTemplate, TreatmentOption } from '@photonhealth/sdk/dist/types';
 import { CatalogStore } from '../stores/catalog';
-import { DisableList } from '../photon-multirx-form/components/PrescribeWorkflow';
+import { DisableListPrefill } from '../photon-multirx-form/components/PrescribeWorkflow';
```
jakefromphoton: If renaming, something even more specific could be nice. Is this disabling medication IDs? So `DisableMedicationPrefills` maybe?

packages/elements/src/photon-multirx-form/components/PrescribeWorkflow.tsx
```diff
@@ -68,11 +69,11 @@ export type Address = {
 };
 
 export interface DisabledItem {
-  treatmentIds?: string[];
+  treatmentIds: string[];
```
jakefromphoton, line 72: whats the intent of making this required when it wasnt previously? makes sense to me that if you're disabling something, ID shuold be required - is this finally enforcing that?

packages/elements/src/photon-multirx-form/components/PrescribeWorkflow.tsx
```diff
@@ -95,8 +96,8 @@ export type PrescribeProps = {
   formActions?: any;
   externalOrderId?: string;
   catalogId?: string;
-  allowOffCatalogSearch?: boolean;
-  disableList?: DisableList;
+  allowOffCatalogSearch: boolean;
```
jakefromphoton, line 99: Intent of making this required?

packages/components/src/analytics/AnalyticsEventListener.tsx
```diff
@@ -77,11 +104,21 @@ export const AnalyticsEventListener = (props: {
...
 
+  // Send anything that was dispatched before the context query resolved
+  createEffect(() => {
+    const context = contextData();
+    if (!context || queuedEvents.length === 0) return;
+
+    const pending = queuedEvents;
+    queuedEvents = [];
+    pending.forEach((detail) => trackEvent(detail, context));
+  });
```
jakefromphoton, line 120: would be helpful to identify where this logic is tested

packages/elements/src/photon-multirx-form/prescribe-workflow-analytics.test.tsx
```diff
@@ -52,6 +52,144 @@ test('page views', async () => {
   );
 });
 
+test('Prescribe Workflow Viewed reports the element configuration', async () => {
```
jakefromphoton, line 55: does this test or another one exercise the flow where the analytics takes ~1sec to load, and we want to queue the analytics event? want to confirm that flow is tested somewhere.

packages/components/src/utils/Prefill.ts
myang5: "Valid" means it was literally just parseable as JSON.

"Verified" - I think a better name would be "validated". Any string might be JSON parseable but engs need to take further steps to validate that the result of the parsing matches the Typescript type, for ex. `{"firstName": "John"}` is valid JSON but it does not match the type enforced by `SupervisorInput`, where many other properties are required.

Will update the comment with examples

packages/elements/src/photon-multirx-form/components/PrescribeWorkflow.tsx
myang5: Yes, enforcing.

I think this was previously typed optional to document the kind of JSON prefill behavior we've been encountering more frequently, so this is updating types so that the code is forced to narrow `DisableListPrefill` to `DisabledItem[]`

packages/elements/src/photon-multirx-form/components/PrescribeWorkflow.tsx
myang5: allowOffCatalogSearch always defaults to true in photon-multirx-form/photon-prescribe-workflow-component.tsx, so just updating the type to reflect that it should always be defined

packages/elements/src/photon-medication-search/photon-medication-search-component.tsx
myang5: done

packages/components/src/analytics/AnalyticsEventListener.tsx
myang5: It's tested in packages/elements/src/photon-client/photon-client-analytics.test.tsx

packages/elements/src/photon-multirx-form/prescribe-workflow-analytics.test.tsx
myang5: Yes, tested in packages/elements/src/photon-client/photon-client-analytics.test.tsx

.github/workflows/vitest.yml
myang5, line 1: Was trying to investigate why some tests became flaky and noticed these unnecessary steps in the file

.github/workflows/vitest.yml
```diff
@@ -20,14 +20,8 @@ jobs:
         with:
           node-version: 22
           cache: 'npm'
-      # there was an issue with the rollup binary not being available in the runner https://github.com/npm/cli/issues/4828
-      - name: Install Rollup
-        run: npm install @rollup/rollup-linux-x64-gnu
```
myang5, line 25: This install step was taking anywhere from 30-45 seconds lol, doesn't seem like it's needed

packages/elements/src/photon-multirx-form/components/PrescribeWorkflow.tsx
jakefromphoton: got it - so this is an internal required type, not external facing (external would be a breaking change for some customers)

.github/workflows/vitest.yml
jakefromphoton: removing these could be pre-mature. might some tests depend on the GQL types being generated via codegen? are you sure you didnt just have these generated GQL types cached? would recommend clearing your caches and running all the tests across the repo, if you didnt! particualrly the patient app tests

.github/workflows/vitest.yml
jakefromphoton: sounds good. this patched an old issue in codebuild/githubActions that is hopefully resolved now due to other version/image updates. if you see an issue in codebuild/githubActions, could be related to this

.github/workflows/vitest.yml
myang5: oh sorry this isn't clear - i update the the patient project.json to depend on codegen, so patient:codegen will now run as part of patient:test

packages/elements/src/photon-multirx-form/components/PrescribeWorkflow.tsx
myang5: ya, internal typing

jakefromphoton (approved):

PR #2804, opened 2026-07-30
group alerts by med name

scottdesilva-photon:
Screenshot 2026-07-30 at 11 54 06 AM

We're doing this to reduce the duplicated alert names. Before this change, we would show two Zepbounds (one for auto-injector and one for solution). We show the brand name so we would still show "Zepbound Kwikpen" AND "Zepbound" but at least this is a step in the right direction.

### DIFF
--- a/packages/components/src/systems/ScreeningAlerts/ScreeningAlertsByPrescription.tsx
+++ b/packages/components/src/systems/ScreeningAlerts/ScreeningAlertsByPrescription.tsx
@@ -23,11 +23,12 @@ function groupAlertsByPrescription(
   const prescriptionMap = new Map<string, AlertsForPrescription>();
 
   for (const alert of screeningAlerts) {
-    // loop through alerts for the drafted prescriptions in the order and organize by prescription id
+    // loop through alerts for the drafted prescriptions and organize by prescription name
+    // so multiple alerts for the same medication share a single group
     for (const prescription of alert.involvedEntities.filter(isDraftedPrescription)) {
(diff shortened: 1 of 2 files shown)


billoran8 (approved):
could dedupe by tc10 too if you wanted to condense the zepbound kwikpen and zepbound

packages/components/src/systems/ScreeningAlerts/ScreeningAlertsByPrescription.tsx
```diff
@@ -23,11 +23,12 @@ function groupAlertsByPrescription(
...
   for (const alert of screeningAlerts) {
-    // loop through alerts for the drafted prescriptions in the order and organize by prescription id
+    // loop through alerts for the drafted prescriptions and organize by prescription name
+    // so multiple alerts for the same medication share a single group
     for (const prescription of alert.involvedEntities.filter(isDraftedPrescription)) {
-      const group = prescriptionMap.get(prescription.id) ?? { prescription, alerts: [] };
+      const group = prescriptionMap.get(prescription.name) ?? { prescription, alerts: [] };
       if (!group.alerts.includes(alert)) group.alerts.push(alert);
-      prescriptionMap.set(prescription.id, group);
+      prescriptionMap.set(prescription.name, group);
```
billoran8, line 31: maybe a test?

scottdesilva-photon:
> could dedupe by tc10 too if you wanted to condense the zepbound kwikpen and zepbound

good point. I think that's ideal but I'm worried about what kind of side effects/edge cases there might be if we group by tc10...do you think it could it end up being too generalized in some cases?

PR #2806, opened 2026-07-31
small improvements to optimize marketplace load

scottdesilva-photon:
- didn't bother looking into how often top ranked pharmacies are set but we should be making these calls in parallel with the general pharmacies call
- removed loading guard that prevents us showing anything on pharmacy page until the GetPharmaciesByLocation resolves. We can start displaying the page with just order details and a loading spinner on the pharmacies list 

https://github.com/user-attachments/assets/27ddbe67-96e2-47db-bd8a-dcc61290970c

https://github.com/user-attachments/assets/a5cd0002-1130-4757-afb8-b59ddead54ce


apps/patient/src/views/Review.tsx
```diff
@@ -48,7 +48,7 @@ export const Review = () => {
 
       // Address saved successfully
       setAddressSubmitted(true);
-      await fetchOrder();
+      await fetchOrder(undefined, { triggerNavigationAfterFetch: false });
```
mrrsnn, line 51: what's this for?

mrrsnn (approved):
Big improvement, great job. Just curious about one line.

apps/patient/src/views/Pharmacy.test.tsx
```diff
@@ -534,6 +534,49 @@ describe('Pharmacy page', () => {
...
+    await navigateToPharmacyScreen();
+
+    // Pharmacy results haven't resolved yet so don't render preferred pharmacy alone
+    expect(screen.queryByTestId('pharmacy-info')).not.toBeInTheDocument();
+
+    resolvePharmacies!({
+      pharmaciesByLocation: [generatePharmacy({ id: 'phr_otherId999', name: 'Other Pharmacy' })]
+    });
+
+    expect(await screen.findByText('Preferred Pharmacy')).toBeInTheDocument();
```
billoran8, line 576: good test!

billoran8 (approved):

apps/patient/src/SendToPatientDemo.test.tsx
```diff
@@ -64,7 +65,7 @@ describe('Send To Patient Demo', () => {
     await userEvent.click(screen.getByText('Capsule Pharmacy'));
     await userEvent.click(screen.getByText('Select pharmacy'));
 
-    await waitFor(() => screen.findByText('Preparing order...'), { timeout: 2500 });
+    expect(await screen.findByText('Order placed', {}, { timeout: 2500 })).toBeInTheDocument();
```
billoran8, line 68: just curious why are we now looking for `order placed` instead of `preparing order...` ?

apps/patient/src/views/Review.tsx
scottdesilva-photon: fetchOrder will trigger the navigateForOrder callback in Main.tsx but we're also attempting to navigate to /pharmacy at the end of this funciton in Review.tsx. After entering their address, a patient should always go to /pharmacy so i'm letting this function handle that navigation. `{ triggerNavigationAfterFetch: false }` skips the navigating that tries to happen in Main.tsx.

(fought with tests a lot in this PR but one of them was actually the reason I found this)

apps/patient/src/SendToPatientDemo.test.tsx
scottdesilva-photon: when I tested locally, I wasn't able to see "Preparing order..." for this scenario. Not exactly sure how it wasn't failing before but maybe it's flaky?

PR #2812, opened 2026-08-04
fix combine orders sending duplicate rx

jakefromphoton:
Prevent duplicate order submissions from the combine-orders dialog

  - Close the combine dialog on terminal failure (both updateOrder and the createOrder fallback failed) instead of leaving the button re-enabled. Previously, each re-click re-sent the same failing mutations, which generated orphaned fulfillments server-side
  - Clear draft prescriptions after a successful combine (both the updateOrder path and the createOrder fallback). Embed hosts that don't handle `photon-order-combined` never unmount the workflow, so consumed drafts stayed resubmittable; the clinical app is unaffected since it navigates away regardless.
  - Close the dialog on fallback-order success, matching the primary success path.

Part of TECH-532


zac-photon (approved):

packages/elements/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@photonhealth/elements",
-  "version": "0.23.11",
+  "version": "0.23.12",
```
zac-photon: Do we have specific stakeholders we need to communicate this update to?

PR #2818, opened 2026-08-05
remove datadog RUM session wiring

myang5:
Part of TECH-807


apps/app/src/globals.d.ts
myang5, line 1: This file was just TS config for the `__COMMIT_HASH__` env var

apps/app/buildspec.yml
myang5, line 1: ⚠️ I'm pretty sure these steps are safe to remove from the build since afaik the sourcemap upload is purely for Datadog RUM, but lmk if otherwise

apps/app/vite.config.ts
myang5, line 1: ⚠️ I'm pretty sure the sourcemaps were only used for Datadog RUM, but lmk if otherwise

mrrsnn (approved):

